PR #7, opened 2018-06-01
Updating helm chart and improving style

louis-murray:
Hi, I was interested in using the argo helm chart but noticed that it wasn't one to one with what happens in `argo install`. I know this PR will add some breaking changes but I strongly feel that it makes the chart more usable and understandable. Happy to make any changes you guys feel need to happen for this PR to go in. Looking forward to hearing your feedback!

The things I added/tweaked are:
* Moved the CRD define into the `_helpers.tpl` because the `json` file was flagged as invalid by the helm linter
* added the info related to setting up argo's service accounts + cluster role bindings
* added a field to the `values.yaml` to handle giving the controller's service account the same cluster role binding in namespaces where workflows will run
* updated minio to the latest version
* refactored the `values.yaml` to group similar values together
* setup some default minio subchart value overrides
* added a README to give an overview of the chart


charts/argo/templates/_helpers.tpl
```diff
@@ -14,3 +14,23 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+
+{{- define "workflow-crd-json" }}
```
DSchmidtDev: I personally don't like the workflow-crd.json to be placed in the helpers.tpl because it's no real helper but an important resource definition.
Maybe we could place it it in an separate templates/etc folder as _workflow-crd.json and create the resource from there?

louis-murray: I agreed it's better to have it separated out, I'll make it it's own `tpl` file

louis-murray: done

alexmt:
published new chart version

louis-murray:
sweet! ty for merging 👍

PR #13, opened 2018-08-09
set minio to not be installed by default

louis-murray:
I've found that helm has issues with conditional sub charts of a sub chart so I think it'd be better to have minio not be installed as part of the chart by default so if you want to leverage this chart as part of another chart you don't have to worry about unintended resources getting created on cluster.

### DIFF
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -29,7 +29,7 @@ useStaticCredentials: true
 # If set to true then chart installs minio and generate according artifactRepository section in workflow controller config map
 minio:
   # This is not used by the subchart, but makes sense to group here.
-  install: true
+  install: false
   defaultBucket:
     enabled: true
     name: argo-artifacts


louis-murray:
Yes exactly. Like it was always defaulting to to whatever `minio.install` is in the `values.yaml` even when it's being overridden in the parent chart that has a dependency on Argo

jrnt30:
One thing that gets a bit strange here is that transitive dependencies settings are still defined at the `root` level of your `values.yaml`.  If your chart depends on `argo-helm/argo` and you want to configure the `minio` chart.  Those settings still occur at the `minio.` level.  

That being said, I don't see disabling it by default an issue.

louis-murray:
I wasn't aware of that, good to know. Thanks for the insight @jrnt30!

louis-murray:
With all that being said if we want to leave it on by default and just note this behavior in the readme I think that's good also

jessesuen:
OK, I think *not* installing minio as the default behavior, is be the behavior I would expect out of the box. Thanks all.

PR #16, opened 2018-08-23
Move to helm's crd-install hook

ishitatsuyuki:


ishitatsuyuki:
This is having RBAC errors for unknown reasons. Closing.

Issue #20, opened 2018-09-18
Simpler CRD creation


sebastien-prudhomme:
Hi,

Just need an explanation why a job is used to created the CRD in Argo chart instead of using a simple crd.yaml file with all the declarations.

Thanks 

BTW: there is a duplicate definition of workflow-crd-json, in _helpers.tpl and in _workflow-crd.tpl

PR #21, opened 2018-09-23
[WIP] Update argo-events chart for the upcoming v0.5 release (argoproj/argo-events#92)

ilyasotkov:
### Update argo-events Helm chart for the upcoming v0.5 (argoproj/argo-events#92)

- [ ] Apply principle of least privilege for `ClusterRole` resources
- [ ] Change gateway-controller docker image back to `argoproj`
- [x] Resolve https://github.com/argoproj/argo-helm/issues/20
- [ ] Be consistent with label use
- [ ] Something else?


VaibhavPage:
@ilyasotkov v0.5 is released.

charts/argo-events/templates/gateway-controller-clusterrole.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
andrewgdavis: is this applying the principle of least privilege?  it looks like the role has access to everything

ilyasotkov: @andrewgdavis obviously not :( This was just copied from the upstream.

I'd prefer to get it fixed in https://github.com/argoproj/argo-events/blob/master/hack/k8s/manifests/argo-events-cluster-roles.yaml (not sure if an issue is already filed or not), then copy it here.

VaibhavPage: The chart has been updated. Can we close this pr?

PR #31, opened 2019-02-05
Allow disabling ui and deploying ui related resources

kbial:
Allow disabling the UI and deploying UI related resources.

For security reasons we do not want to allow the UI in production deployments and disabling ingress is not enough since it’s preferable to not have unused resources running.


bclermont:
as long as you don't turn on ingress, it don't really make a big difference, no?

wookasz:
> as long as you don't turn on ingress, it don't really make a big difference, no?

That's not true. There's an unnecessary container sitting around in the cluster, which violates security best practices and uses up some, albeit small, amount of resources unnecessarily. Our vulnerability scanning tools are flagging the image as well so the choice is to either continually patch and maintain something that isn't used or just get rid of it.

bclermont:
You're right.

BTW, which "vulnerability scanning tools" you are using?

wookasz:
We're using twistlock.

alexec:
@kbial this looks good to me, but can you fix the merge conflict please?

PR #32, opened 2019-02-08
Add archiveLogs option to Argo workflow controller config map

kornesh:
Adds the option to persist workflow logs to default `artifactRepository`

### DIFF
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -72,7 +72,7 @@ useDefaultArtifactRepo: false
 useStaticCredentials: true
 artifactRepository:
   # archiveLogs will archive the main container logs as an artifact
-  archiveLogs: true
+  archiveLogs: false
   s3:
     # Note the `key` attribute is not the actual secret, it's the PATH to
     # the contents in the associated secret, as defined by the `name` attribute.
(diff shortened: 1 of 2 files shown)


charts/argo/values.yaml
```diff
@@ -71,6 +71,8 @@ ui:
 useDefaultArtifactRepo: false
 useStaticCredentials: true
 artifactRepository:
+  # archiveLogs will archive the main container logs as an artifact
+  archiveLogs: true
```
alexec: Should this be false by default?

kornesh: When some enables `useDefaultArtifactRepo` wouldn't make sense to automatically start archiving the logs ?

kornesh: For instance, if `minio.install` is enabled, `minio.defaultBucket.enabled` is automatically enabled by default

alexec: Not sure. Does this change behaviour?

kornesh: I'm not sure. Probably only for those who have already have enabled `useDefaultArtifactRepo`, now it will automatically upload logs as well. Should we just disable it by default?

alexec: I think so. Better not to have surprises when you upgrade.

kornesh: Agreed. I've just updated the PR and set `archiveLogs` to `false` by default.

PR #33, opened 2019-02-09
Helm chart for ArgoCD

lcostea:
This will create a Helm chart for ArgoCD.
Compared to the existing manifests it adds:

- support for ingress with ssl passthrough (default is disabled)
- parameter to install on cluster level(default) or namespace level
- crds are created with helm crd-install hooks
- new labels recommended by https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
- parameterised the most important values: image repo, tags, pullPolicy and container and service ports

After installation with helm on minikube I made some tests the apps defined here: https://github.com/lcostea/argocd-helm-deployment

I didn't try to setup argocd with argocd (to update itself), it seems there will be some additional work here, at least on how the crd are installed (can't use the crd-install hook probably), but I think that should be a separate issue/PR. So this version is probably good enough to start playing with argocd on a cluster, but for production the kustomize version is still recommended, at least it can update itself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 output
 .vscode
+.DS_Store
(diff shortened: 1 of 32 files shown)


lcostea:
Closing so I can fix the missing email on the repo.

PR #36, opened 2019-02-16
Make containerRuntimeExecutor configurable

dtaniwaki:
I'd like to configure `containerRuntimeExecutor`. Would you consider merging this change?


alexec:
@dtaniwaki do you still need this change please?

PR #40, opened 2019-02-28
respect .Release.Namespace

nstott:
removes the hardcoded argo-events namespace, and uses `{{ .Release.Namespace }}` instead where it's needed in the rolebinding


VaibhavPage:
I guess we can close this pr?

PR #47, opened 2019-04-02
Allow extraargs for argocd server

viggeh:
Allowing extra arguments to be specified for argocd-server.

This is useful, for instance, if you need to change the basehref or use insecure when using ingress
https://github.com/argoproj/argo-cd/blob/master/docs/ingress.md


alexmt:
Looks great! Can you please update chart version to 0.2.1 ?

arnarg:
I have write permissions to the repository so I took the liberty of updating the chart version.

Issue #27, opened 2018-12-14
Installation fails with "CRD already exists" error


mikeraimondi:
This is on version 0.3.1

`helm install` fails with the following error: 
`Error: release argo failed: customresourcedefinitions.apiextensions.k8s.io "workflows.argoproj.io" already exists`

I have verified that the CRD is not present with `kubectl get crd` prior to running `helm install`.

alexmt:
Tried to run `helm install` on fresh minikube and could not reproduce the issue. I'm pretty sure chart don't include CRD definition twice. 

Can you double-check if your cluster has workflow CRD installed? You also might run `helm install argo/argo --no-crd-hook` to skip any CRD installation.

mikeraimondi:
I double and triple-checked that my cluster didn't have the workflow CRD installed. In fact, I manually deleted the workflow CRD, then ran `helm install` and got the same error about the CRD already existing.

Yep, we're using `--no-crd-hook` as a workaround. Very puzzled by this.

If it helps to correlate with other bug reports, the environment is GKE, version 1.10.6-gke.11.

bbhoss:
Just ran into this myself using Docker for Mac/Kubernetes for Mac, stable channel (2.0.0.0-mac81 (29211)). Kubernetes  v1.10.3

harshal-shah:
This happened to me with helm version 2.12.0 but does not happen in 2.11. I ran into this with jaeger-operator.

bclermont:
still happens on helm 2.13

chriskolenko:
Work around:

Delete custom definitions 
```
kubectl delete customresourcedefinition/applications.argoproj.io -n argocd --force=true --grace-period=0 
kubectl delete customresourcedefinition/appprojects.argoproj.io -n argocd --force=true --grace-period=0
```

Then reinstall. :'(

rehevkor5:
Deleting the CRDs worked for one deploy, but now the problem is back again :( with:

```
    * module.argo.helm_release.argo_workflows: 1 error occurred:
    * helm_release.argo_workflows: rpc error: code = Unknown desc = object is being deleted: customresourcedefinitions.apiextensions.k8s.io "workflows.argoproj.io" already exists
```

bclermont:
I wanted to use argo for ci/cd but as it's not ready to deploy on k8s I abandon that thought

I'll take a look again in 6 months

mattmelgard:
[...]

PR #50, opened 2019-04-11
workflow-crd.yaml: delete before hook creation

yurrriq:
Without this change, you end up with the following error on rolling upgrades:
```
Error: customresourcedefinitions.apiextensions.k8s.io "workflows.argoproj.io" already exists
ERROR: Job failed: exit code 1
```


yurrriq:
Fixes #27

PR #63, opened 2019-05-07
Add configManagementPlugins option to argo-cd

cw-ozaki:
https://argoproj.github.io/argo-cd/user-guide/application_sources/#config-management-plugins

Added `configManagementPlugins` option because it was not supported.


mumoshu:
@alexec Would you mind publishing this to the `gh-pages`-based charts repo? Thanks!

apicht:
Did this get removed/reverted? Why?

Issue #67, opened 2019-05-09
Expose Type in Server Ingress


Sherwoodwt:
To set up ArgoCD Server behind a Load Balancer, the service must be type LoadBalancer. The service template (argocd-server-service.yaml) doesn't include a way to set the type based on a values file.

I'd like to be able to set the type using a values file. The only workarounds I've found for this have been to either fork the repo, modify the one template file and reference the forked repo, or to call `kubectl patch` to set the type after creation as suggested here: https://argoproj.github.io/argo-cd/getting_started/#service-type-load-balancer.

Here's an example of what the fix might look like:
```
# argocd-server-service.yaml

apiVersion: v1
kind: Service
metadata:
  ...
spec:
  type: {{ .Values.server.serviceType }}
  ports:
     ...
  selector:
    app.kubernetes.io/name: {{ include "argo-cd.name" . }}-server
```

alexec:
Open for contribution.

PR #68, opened 2019-05-10
Add type value to argocd server service

Sherwoodwt:
Issue link: https://github.com/argoproj/argo-helm/issues/67

### DIFF
--- a/charts/argo-cd/templates/argocd-server-service.yaml
+++ b/charts/argo-cd/templates/argocd-server-service.yaml
@@ -13,6 +13,7 @@ metadata:
   annotations:
 {{ toYaml .Values.server.serviceAnnotations | indent 4}}{{- end }}
 spec:
+  type: {{ .Values.server.serviceType }}
   ports:
   - name: http
     protocol: TCP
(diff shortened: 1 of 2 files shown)


charts/argo-cd/templates/argocd-server-service.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
   annotations:
 {{ toYaml .Values.server.serviceAnnotations | indent 4}}{{- end }}
 spec:
+  type: {{ .Values.server.serviceType }}
```
alexec: I think you need if condition around this new field? Or is `ClusterIP` default value?

alexec:
Please sign CLA!

charts/argo-cd/templates/argocd-server-service.yaml
Sherwoodwt: ClusterIP is the default value: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types.

However, if its preferable I could wrap this in a conditional and leave ClusterIP out of the values file. That way it's still modifiable, but if not added it will have the exact same functionality it had before the change.

Sherwoodwt:
I'm working on getting approval checked through work to ensure I can sign the CLA 👍

brunowego:
@Sherwoodwt ping

amaya382:
Any progress?
I want to use this feature

damascenorakuten:
Any update on this?

Thanks

PR #69, opened 2019-05-11
[argo] Create aggregate roles as pre-install hook

naseemkullah:
cc @jessesuen

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "v2.2.1"
 description: A Helm chart for Kubernetes
 name: argo
-version: 0.3.2
+version: 0.4.0
--- a/charts/argo/templates/workflow-aggregate-roles.yaml
+++ b/charts/argo/templates/workflow-aggregate-roles.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.createAggregateRoles }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -70,3 +71,4 @@ rules:
   - patch
   - update
   - watch
+{{- end }}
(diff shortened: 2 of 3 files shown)


charts/argo/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "v2.2.1"
 description: A Helm chart for Kubernetes
 name: argo
-version: 0.3.2
+version: 0.4.0
```
alexec: Should this be `0.3.3`?

charts/argo/templates/workflow-aggregate-roles.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
alexec: Should there be a feature flag around these manifests?

naseemkullah: Sure, what should it look like in the values file? Here are a few ideas, can also replace `create` with `enable` or `install` if you'd prefer. 

```
aggregateRoles:
  view:
    create: true
  edit:
    create: true
  admin:
    create: true
```

```
createAggregateRoles: true
```

```
aggregateRoles:
  create: true
```

charts/argo/Chart.yaml
naseemkullah: I think minor bump is ok since minor typically means "add functionality in a backwards-compatible manner",and I believe this is the case. Thoughts?

charts/argo/templates/workflow-aggregate-roles.yaml
alexec: I'd suggest a single big flag. KISS

naseemkullah: Added it as such, please review and resolve conversation if you see fit.

charts/argo/Chart.yaml
alexec: semver - you're correct.

PR #73, opened 2019-05-23
Update Argo to v2.3.0 and add support for Prometheus operator

sebastien-prudhomme:
This PR does several things:

- update to the latest version of Argo (v2.3.0)

- allow to enable telemetry endpoint, the same way the metrics endpoint is enabled (personal comment: i suggest to merge this two endpoints in the Argo product to simplify things)

- allow to activate Prometheus ServiceMonitor for both metrics and telemetry endpoints

- add a Service resource for both metrics and telemetry endpoints (needed for Prometheus ServiceMonitor)

I've tried to match the current way of declaring resources in the Argo chart.


dberuben:
Any update about this PR ?

PR #80, opened 2019-06-25
loadBalancerIP support in argo-ui

l15k4:
Guys I think that Argo-ui is used a lot internally in companies through VPN, therefore it would make sense to allow for setting static IP address for internal loadBalancer. WDYT ?

### DIFF
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -51,6 +51,9 @@ ui:
   serviceAccount: argo-ui
   # Annotations to be applied to the UI Service
   serviceAnnotations: {}
+  # Static IP address to assign to loadBalancer
+  # service type `LoadBalancer`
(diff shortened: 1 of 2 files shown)


charts/argo/values.yaml
```diff
@@ -51,6 +51,9 @@ ui:
   serviceAccount: argo-ui
   # Annotations to be applied to the UI Service
   serviceAnnotations: {}
+  # Static IP address to assign to loadBalancer
+  # service type `LoadBalancer`
+  loadBalancerIP: "123"
```
alexec: I think this should be removed, otherwise it'll default to "123".

l15k4: Ah, yeah I tested if it works and forgot it there :+1:

alexec:
Merging!

PR #81, opened 2019-07-01
argo: add support for individual image tags.

tompowell-dd:
This allows the use of digests for the controller/ui/executor.


tompowell-dd:
Sorry for the delay - needed to wait on a CCLA to be signed/sent.

PR #83, opened 2019-07-05
[Argo] Fix appVersion in argo chart

paguos:
- Update from v2.2.1 to v2.3.0


charts/argo/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "v2.2.1"
+appVersion: "v2.3.0"
 description: A Helm chart for Kubernetes
 name: argo
 version: 0.5.0
```
paguos: Not sure if the version number has to be updated... 😉

PR #89, opened 2019-07-24
Improve chart description

se-jaeger:


charts/argo/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 appVersion: "v2.3.0"
-description: A Helm chart for Kubernetes
+description: A Helm chart for Argo
```
alexec: minor - should say "Argo Workflows"

se-jaeger: You're right that's better.

alexec:
LGTM!

PR #91, opened 2019-08-09
Updating the version of argo-cd

wxncjonorri:
Method of comparing:
See differences between the k8s resources here: https://github.com/argoproj/argo-cd/tree/master/manifests
Between the 0.12.3 tag and the 1.1.2 tag.


alexec:
I approve - but I think we should get more community feedback.

spencergilbert:
I was looking into this as well, based on the kustomize project that's more maintained we can also upgrade the dex image to v2.14.0

I wasn't sure if the chart should be bumped to v1.0.0 since argocd also hit that milestone.

PR #97, opened 2019-08-19
add admin password and tls values to argocd admin-secret.yaml

trawler:
This PR proposes to add the following values to argo-secret.yaml for ArgoCD:
* `.Values.config.admin.password`
* `.Values.config.admin.passwordMtime`
* `.Values.config.server.secretkey`
* `.Values.config.tls.crt`
* `.Values.config.tls.key`

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.1.2"
 description: A Helm chart for Argo-CD
 name: argo-cd
-version: 0.3.0
+version: 0.3.1
--- a/charts/argo-cd/templates/argocd-secret.yaml
+++ b/charts/argo-cd/templates/argocd-secret.yaml
@@ -2,14 +2,29 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: argocd-secret
-  labels: 
+  labels:
     app.kubernetes.io/name: {{ include "argo-cd.name" . }}
     helm.sh/chart: {{ include "argo-cd.chart" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/part-of: {{ include "argo-cd.name" . }}
 type: Opaque
-data:  
+data:
+{{- if .Values.config.admin.password }}
+  admin.password: {{ .Values.config.admin.password }}
+{{- end }}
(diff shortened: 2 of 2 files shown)


alexec:
I approve. Please update chart version.

trawler:
@lcostea 
> When dealing with secrets we usually want them to be handled outside, like another application to create and populate it initially (at least that's what I am used to).

You are right, that it's a decision worth discussing, if we want this helm chart to even manage argocd secrets.. Currently, however, they are managed by this helm chart, but with a missing configuration that requires patching after installation (since argocd controller complains about the missing keys in the secret).
A possible solution perhaps is to define also a `manage_secrets` flag that will enable/disable the creation of the argocd secret, in something similar to this:
```{{- if .Values.config.manage_secrets }}
apiVersion: v1
kind: Secret
metadata:
  name: argocd-secret
  labels:
     etc. etc....
type: Opaque
data:
{{- if .Values.config.admin.password }}
  admin.password: {{ .Values.config.admin.password }}
{{- end }}
{{- if .Values.config.admin.passwordMtime }}
  admin.passwordMtime: {{ .Values.config.admin.passwordMtime }}
{{- end }}
{{- if .Values.config.server.secretkey }}
  server.secretkey: {{ .Values.config.server.secretkey }}
{{- end }}

etc.. etc..

{{- end }}
```

> While this seems like a good addition, I can't find a scenario when it is actually useful. Could you provide one so we can better understand the reason?

The reason behind this PR is that I'm working on automating the ArgoCD server using a helm operator, by creating a custom resource with the relevant values.yaml variables. In this way, I don't have to manage the secrets separately... especially since they are partially managed in this chart.

lcostea:
Thanks for the explanation, it starts to make sense now.

ArgoCD can be installed without those secrets because they are generated on install. They are also missing in the kustomize version and simple manifests. Sometimes the application controller might start before the actual generation takes place, but it shouldn't be a problem, you can follow this for more details: https://github.com/argoproj/argo-cd/issues/1936

So actually the chart should not manage those secrets, it is not a good idea to have them passed as values and be stored some where in source control. I believe it is better to have the secret created by another app, if you want to control them. For example we use something like [aws-ssm](https://github.com/cmattoon/aws-ssm) to populate an existing secret. Or use the cli after installation to change the password. Will something like that work for you?

trawler:
Thanks @lcostea for clarifying the expected behaviour.
From first look, seems like I may indeed be running into the bug you've linked above, so I'll try to reproduce it, and if possible, to provide a fix. For now, this PR can be closed.

PR #100, opened 2019-09-02
Be able to define extra configuration in argo

samwyma:
Let me know if I need to do/add anything else


alexec:
I've requested a review of this from Argo Workflow devs on Slack. I can probably tell you that you need to increment the chart version.

samwyma:
@alexec Sorry about the delay on this. I've fixed the conflicts and have also bumped the chart patch version

PR #106, opened 2019-09-13
Use Argo CD URL as first ingress host

yann-soubeyrand:
This avoids configuration duplication.

### DIFF
--- a/charts/argo-cd/templates/argocd-server-ingress.yaml
+++ b/charts/argo-cd/templates/argocd-server-ingress.yaml
@@ -17,7 +17,16 @@ metadata:
 {{- end }}
 spec:
   rules:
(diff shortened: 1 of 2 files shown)


yann-soubeyrand:
Yes, I think it can be considered a breaking change even though it won't break anything in what I think is the most common case (setting ingress host to the external domain of Argo CD).

alexec:
I trust you.

yann-soubeyrand:
Hi @alexec,

I'm sorry, I've clearly overlooked that Argo CD URL was (obviously) a URL whereas ingress host must be a domain. I'll fix it using a regex to extract the domain from the config URL (and maybe the path?).

Once again, sorry for the mistake.

PR #120, opened 2019-09-23
argo-cd/crds: add 'hook-delete-policy: before-hook-creation'

gpaul:
The pattern is standard to support 'helm upgrade' for charts that define CRDs.


gpaul:
I don't know what the effect is when the CRD is upgraded though - whether all the existing Application resources will get nuked when `helm upgrade` deletes and recreates the Application CRD. If that is the case, we should probably not merge this PR.

alexec:
Question to community - do we really want to delete CRDs?

PR #129, opened 2019-10-11
init v1.0.0

jdeprin:
Checklist:

* [x] Any new values are backwards compatible and/**or have sensible default.**
* [x] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md). (guide updated with new service name)

proposed v1.0.0 chart with several additions including
* resource request and limit settings
* full featured ingress for the argocd server
* metrics service enabling and prometheus servicemonitor creation
* Expanded and definable command line flags
* Global argocd image repo and tag setting
* nodeSelectors, Affinity and tolerations.
* Pod annotations and labels.
* more supportable and expandable configuration settings
* better organization.
* More detailed comments and readme

v1.0.0, not backwards compatible.


alexec:
This is a large change and really needs a few sets of eyes on it.

charts/argo-cd/templates/NOTES.txt
```diff
@@ -1,13 +0,0 @@
-In order to access the server UI you have the following options:
```
lcostea: I think the NOTES.txt do provide valuable information after you install the chart, for example showing you the generated admin password for the ui.

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
```diff
@@ -0,0 +1,108 @@
+{{- if .Values.controller.enabled }}
```
lcostea: I don't think there is ever a case when the controller can be disabled, the server, repo-server and controller they all work together. So I don't see any point in configuring enable/disable of it.

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "argo-cd.controller.fullname" . }}
+  labels:
+    release: {{ .Release.Name }}
```
lcostea: I don't see why we need a new label, especially that it is not in the recommended list: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
We could use `app.kubernetes.io/instance` for selectors

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
```diff
@@ -0,0 +1,108 @@
+{{- if .Values.controller.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "argo-cd.controller.fullname" . }}
```
lcostea: What if we don't template this and use it like in the labels below: `{{ include "argo-cd.fullname" . }}-{{ .Values.controller.name }}` We get rid of one extra level of indirection, have less code (we should remove the functions from `_helpers.tpl`) and the name becomes more clear.
This applies to all the naming functions used.

charts/argo-cd/templates/argocd-configs/argocd-cm.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.server.enabled }}
```
lcostea: I might be missing something, but I don't see any case where the server will ever be disabled. I think we should not have such a configuration at all.

charts/argo-cd/templates/argocd-configs/argocd-ssh-known-hosts-cm.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if and .Values.server.enabled }}
```
lcostea: There is something extra or missing here with the `and`. Anyway I think we should remove the `server.enabled` configuration from all the places.

charts/argo-cd/values.yaml
```diff
@@ -1,184 +1,528 @@
...
+    serviceMonitor:
+      enabled: false
+    #   selector:
+    #     prometheus: kube-prometheus
+    #   namespace: monitoring
+    #   additionalLabels: {}
+
+## Dex
+dex:
+  enabled: true
```
lcostea: This is a really good thing, dex is not critical to argo-cd and many times you don't actually need it. I wonder even if the default should be false, what do you think @alexec ?

charts/argo-cd/templates/argocd-application-controller/clusterrole.yaml
```diff
@@ -1,15 +1,16 @@
-{{- if .Values.clusterAdminAccess.enabled }}
+{{- if .Values.controller.enabled }}
```
lcostea: I have doubts that this is correct, there are 2 ways to configure argo-cd, if cluster admin is enabled it means it will be able to install application also in the cluster where it runs, if it is disabled then it most likely install the applications in external clusters (so it doesn't need cluster wide rights). So I don't see any connection with enabling/disabling the controller.

charts/argo-cd/templates/argocd-application-controller/clusterrolebinding.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.controller.enabled }}
```
lcostea: I think clusterAdmin should be used here, not controller

charts/argo-cd/templates/argocd-repo-server/deployment.yaml
```diff
@@ -0,0 +1,123 @@
+{{- if .Values.repoServer.enabled }}
```
lcostea: This should not be configurable, you can't disable repo-server

charts/argo-cd/templates/redis/deployment.yaml
```diff
@@ -0,0 +1,61 @@
+{{- if .Values.redis.enabled }}
```
lcostea: @alexec do you think redis can be disabled? I know it is used like a throw away cache, but I also think that argo-cd expects it to be there and I am not sure it will function correctly if it doesn't exist.

lcostea:
I did go once through all of the files, there are pretty many changes to be discussed, so I have finished my review for now.
Thanks @jdeprin that was a lot of work, I really like that you added ServiceMonitors, I also had that in mind for some time.

charts/argo-cd/values.yaml
```diff
@@ -1,184 +1,528 @@
...
+    #   selector:
+    #     prometheus: kube-prometheus
+    #   namespace: monitoring
+    #   additionalLabels: {}
+
+## Argo Configs
+configs:
+  knownHosts:
+    data:
+      ssh_known_hosts: |
```
jannfis: Looks legit from my point of view -- no diff to `hacks/ssh_known_hosts` or to the data in the manifest's ConfigMap

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
```diff
@@ -0,0 +1,108 @@
+{{- if .Values.controller.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
dcherman: This (and other namespaced resources) should specify

```yaml
namespace: {{ Release.Namespace }}
```

in order to support installing with `helm template`.

charts/argo-cd/templates/argocd-configs/argocd-secret.yaml
```diff
@@ -0,0 +1,28 @@
...
+    release: {{ .Release.Name }}
+    app.kubernetes.io/name: {{ include "argo-cd.name" . }}-secret
+    helm.sh/chart: {{ include "argo-cd.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/part-of: {{ include "argo-cd.name" . }}
+    app.kubernetes.io/component: {{ .Values.server.name }}
+type: Opaque
+{{- if or .Values.configs.secret.githubSecret (or .Values.configs.secret.gitlabSecret .Values.configs.secret.bitbucketSecret) }}
+# Setting a blank data again will wipe admin password/key/cert
```
dcherman: Should we support passing in the admin password here?

dcherman:
As a general question, do we want to support HA installations in this chart?

charts/argo-cd/templates/redis/deployment.yaml
jdeprin: Eventually I think bot Dex and Redis should be moved to sub charts.  I don't see a huge value in managing the charts for these two apps as they both have very well documented and supported charts of their own.

charts/argo-cd/templates/NOTES.txt
jdeprin: Yes, thanks for catching that.  added back and updated: https://github.com/argoproj/argo-helm/pull/129/commits/79a683f8c61210340727332e2ec773557395fa95

charts/argo-cd/templates/redis/deployment.yaml
alexec: Good question - we're looking into writing an operator that install Argo CD in a "cluster only" mode. That would not include Redis. I think it is therefore fine - going forward - to disable it.

charts/argo-cd/values.yaml
```diff
@@ -1,184 +1,528 @@
...
+        ssh.dev.azure.com ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7Hr1oTWqNqOlzGJOfGJ4NakVyIzf1rXYd4d7wo6jBlkLvCA4odBlL0mDUyZ0/QUfTTqeu+tm22gOsv+VrVTMk6vwRU75gY/y9ut5Mb3bR5BV58dKXyq9A9UeB5Cakehn5Zgm6x1mKoVyf+FFn26iYqXJRgzIZZcZ5V6hrE0Qg39kZm4az48o0AUbf6Sp4SLdvnuMa2sVNwHBboS7EJkm57XQPVU3/QpyNLHbWDdzwtrlS+ez30S3AdYhLKEOxAG8weOnyrtLJAUen9mTkol8oII1edf7mWWbWVf0nBmly21+nZcmCTISQBtdcyPaEno7fFQMDD26/s0lfKob4Kw8H
+        vs-ssh.visualstudio.com ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7Hr1oTWqNqOlzGJOfGJ4NakVyIzf1rXYd4d7wo6jBlkLvCA4odBlL0mDUyZ0/QUfTTqeu+tm22gOsv+VrVTMk6vwRU75gY/y9ut5Mb3bR5BV58dKXyq9A9UeB5Cakehn5Zgm6x1mKoVyf+FFn26iYqXJRgzIZZcZ5V6hrE0Qg39kZm4az48o0AUbf6Sp4SLdvnuMa2sVNwHBboS7EJkm57XQPVU3/QpyNLHbWDdzwtrlS+ez30S3AdYhLKEOxAG8weOnyrtLJAUen9mTkol8oII1edf7mWWbWVf0nBmly21+nZcmCTISQBtdcyPaEno7fFQMDD26/s0lfKob4Kw8H
+  tlsCerts:
+    data:
+      argocd.example.com: |
+        -----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
alexec: this needs deleting for sure

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
jdeprin: Fixed in https://github.com/argoproj/argo-helm/pull/129/commits/89716e2ebc369b56fbf8749df0c304e647156529.  removed `release` label and replaced selectors with `app.kubernetes.io/instance` label.

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
jdeprin: Some charts, such as the prometheus operator allow you to enable or disable central parts of the system.  The use-case I can imagine is deploying different parts of ArgoCD in different sub-carts.  I don't think that is a very strong argument and so for simplicity  sake I've removed these items.
controller enabled: https://github.com/argoproj/argo-helm/pull/129/commits/4fe65a4e0932e66155e21c485ad04df41098249a
repo server enabled: https://github.com/argoproj/argo-helm/pull/129/commits/41df9ed6725dd7ece14944bc8daff335dd365b15
server enabled: https://github.com/argoproj/argo-helm/pull/129/commits/ca53ad463a24ee62e496e50cd5346744ac31fdd8

charts/argo-cd/templates/argocd-configs/argocd-cm.yaml
jdeprin: Resolved, see comment above.

charts/argo-cd/templates/argocd-configs/argocd-ssh-known-hosts-cm.yaml
jdeprin: .Values.server.enabled removed completely per above comment.

charts/argo-cd/templates/argocd-repo-server/deployment.yaml
jdeprin: This item is fixed on commit https://github.com/argoproj/argo-helm/pull/129/commits/41df9ed6725dd7ece14944bc8daff335dd365b15

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
jdeprin: That is an option but I would argue that it becomes harder to make global changes to names in the future.  having the name of objects defined in the helpers.tpl is very standard and if in the future these names need to change, enabling that change to occur in a single place is a large benefit.

if changed you would need to accurately update names:
argo controller: 11 places / 9 files
argo server: 14 places / 11 files
reposerver: 9 places / 7 files

the way it is now, you simply need to update 3 lines in 1 file.

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
jdeprin: I would disagree for a couple reasons.
First, resources that require a namespace in their spec (roles / cluster roles) have it defined as the {{ Release.Namespace }} however it is not standard for other resources to have this defined.  To define standard I am using the common practices of the helm/stable repository specify charts for graduated or incubating cncf.io projects.  
Second, ArgoCD itself doesn't require items to have namespaces defined during deployment.  It will template items and put them in the namespace defined by the application spec.

dcherman: The usage of `{{ Release.Namespace}}` in templates is not consistent, I'll give you that, but I'd argue that it should be a standard for most charts, especially those that use it elsewhere in templates as in the case of roles, etc.

I can't think of a valid use-case where one would do something like `helm template --namespace foo | kubectl apply -f -` and not expect the templates to be installed in the provided namespace, but that's how it is today.  In order to install things correctly, you actually need to do something along the lines of `helm template --namespace foo | kubectl apply -n foo -f -` which seems redundant.

You're completely correct that ArgoCD itself will place resources in the correct namespace, however this chart is for installing ArgoCD, so you have to have some way to bootstrap the installation; there has to be a first install of ArgoCD somewhere.  Adding the namespace explicitly reduces the difference between `helm install argocd --namespace argocd` and `helm template argocd --namespace argocd | kubectl apply -f -`.

This admittedly is far less of an issue in Helm 3 where the lack of Tiller makes it more reasonable to skip using `helm template`, but something to consider.

charts/argo-cd/templates/argocd-application-controller/clusterrole.yaml
jdeprin: This is moved to the controller block and updated readme/comments https://github.com/argoproj/argo-helm/pull/129/commits/e2215f864ab833364e2c294f63d86755d0e05186

charts/argo-cd/values.yaml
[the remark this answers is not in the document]
jdeprin: removed: https://github.com/argoproj/argo-helm/pull/129/commits/3d929938278edc7355adf769b8f5c6ab7c195a25

charts/argo-cd/templates/redis/deployment.yaml
jdeprin: I defaulted redis.enabled to false: https://github.com/argoproj/argo-helm/pull/129/commits/37e0797a1d59476121f686db81f42d91d18e7265

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
jdeprin: I agree that it can be re-evaluated in helm 3, though for the sake of keeping things consistent with other major projects with helm charts there isn't a compelling reason to stray from the standard. 
I would be of a different opinion if ArgoCD needed to be installed in multiple namespaces but I don't think that is the case at the moment.
When/if that does happen I would strongly suggest not using `{{ Release.Namspace }}` but using a templated `namespaceOverride` value as in the Jenkins chart https://github.com/helm/charts/pull/15202 because it adds a good amount of flexibility for those who only use help as a templating tool.

alexec:
@lcostea @dcherman @jannfis do you all approve of this to be merged?
@jdeprin are you happy to fix any bugs anyone find please? We don't provide support for this charts due to lack of bandwidth.

jdeprin:
@alexec Yup I'll provide support.  bugs can be assigned to me.  If I miss them in my inbox, you can ping me on slack.

alexec:
@lcostea @dcherman @jannfis are you happy with these changes please?

jannfis:
@alexec Sure, I thought my approval from 4 days ago would be sufficient :) But yes, I'm more than fine with the awesome work of @jdeprin and hereby approve.

dcherman:
👍 This is a huge step in the right direction, thanks for doing this!

alexec:
OK. So only change is to sync with master.

charts/argo-cd/templates/argocd-configs/argocd-secret.yaml
seanson: I'd say yes, it's a common enough pattern that folks want to set it and the entire secret can be done out of band if more security is needed.

jdeprin:
I think the linting error is a result of using the helm 3 linter.  Could it be downgraded to helm 2?
according to the docs my charts.yaml is fine. https://helm.sh/docs/charts/#the-chart-yaml-file

charts/argo-cd/values.yaml
seanson: I'd consider good SSO support one of the defining features of ArgoCD and I'd consider Dex enabled by default a newcomer friendly choice, with the option to move it externally later once your ArgoCD installation needs to scale.

alexec:
Brace yourself... merging....

alexec:
Master build: https://circleci.com/gh/argoproj/argo-helm/107

alexec:
You should be able to get this from https://argoproj.github.io/argo-helm now.

PR #130, opened 2019-10-17
Update argo chart for 2.4.2

sudermanjr:
Notable changes:

* Default argo image is 2.4.2
* WorkflowTemplate CRD is created
* RBAC is updated to include workflowtemplate permissions
* CRD install is disabled with a Helm flag in order to be more compatible with tools such as [Reckoner](https://github.com/fairwindsops/reckoner)
* README suggests where to find CRDs for manual install

I have installed this chart using the values that I had for my previous install of Argo on 2.3.0 and it installs cleanly. I tested running both standard workflows and submitting templates and workflows that use templates.

Checklist:

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "v2.4.1"
+appVersion: "v2.4.2"
 description: A Helm chart for Argo Workflows
 name: argo
 version: 0.6.0
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "v2.3.0"
+appVersion: "v2.4.1"
 description: A Helm chart for Argo Workflows
 name: argo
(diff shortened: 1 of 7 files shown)


charts/argo/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "v2.3.0"
+appVersion: "v2.4.1"
```
sarabala1979: can you change  the version to v2.4.2

charts/argo/values.yaml
```diff
@@ -4,9 +4,10 @@ images:
   ui: argoui
   executor: argoexec
   pullPolicy: Always
-  tag: v2.3.0
+  tag: v2.4.1
```
sarabala1979: v2.4.2

charts/argo/Chart.yaml
sudermanjr: done.

charts/argo/values.yaml
sudermanjr: done

PR #134, opened 2019-10-23
[argo-events] Rename argo-events-sa to argo-events

jbehling:
The suffix -sa is redundant, remove it to be consistent with
ServiceAccount naming in argo, and argo-cd charts.

Checklist:

### DIFF
--- a/charts/argo-events/Chart.yaml
+++ b/charts/argo-events/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to install Argo-Events in k8s Cluster
 name: argo-events
-version: 0.5.0
+version: 0.5.1
 keywords:
 - argo-events
 - sensor-controller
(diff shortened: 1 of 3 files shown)


alexec:
Is this a breaking change?

jbehling:
Any gateway or sensor resources that specified `argo-events-sa` for `serviceAccountName:` will need to be updated to `argo-events`.

alexec:
https://argoproj.slack.com/archives/CAM37QYMS/p1571934011069900

jdeprin:
I agree with the redundancy of the -sa extension but this change would break other argo-event objects created outside of the chart.
A better solution may be to use the `{{ .Values.serviceAccount }}` variable in the service account, cluster role and deployment specs and keep the default value as it is now.

jbehling:
Ok, PR is up to date. This is no longer a breaking change. If you keep the default value of `argo-events-sa` you get the previous behavior. If you want to make the naming consistent ( and less redundant) you can provide `serviceAccount` as a variable

jbehling:
Any thoughts @VaibhavPage ?

jbehling:
> Can we have a default `serviceAccount` specified in `values.yaml` as `argo-events-sa`?

Yes, `serviceAccount` is already set to `argo-events-sa` in values.yaml

Issue #147, opened 2019-11-05
Deploying to custom namespace no longer works with v1.0.0 of argo-helm chart


TomHellier:
**Describe the bug**
Installing the v1.0.0 argo-cd helm chart causes the `argocd-server`, `argocd-dex-server`, and `argocd-application-controller` to start crash looping because it cannot find the argocd-cm configuration map.

I think this is because this [line of code in util/settings/settings.go](https://github.com/argoproj/argo-cd/blame/37641cf2d08138140ca957106d5fdac6942a77c8/util/settings/settings.go#L489) expects that the `app.kubernetes.io/part-of` is set to `argocd`, I propose we need to update the helm chart to not change the part-of section of any of the manifests.
```
    app.kubernetes.io/part-of: argocd-ref
```

**To Reproduce**
Configure the argocd chart to install to a namespace other than `argocd`, and configure the `.Values.nameOverride` to be `argocd-ref`.

**Expected behavior**
The `argocd-dex-server`, `argocd-server`, and `argocd-application-controller` all start up successfully.

TomHellier:
I think because my change has been merged, I can close this pull request.

PR #149, opened 2019-11-05
Resolves issues relating to setting the .Values.nameOverride to a values other than argocd.

TomHellier:
Relates to github issue: https://github.com/argoproj/argo-helm/issues/147

The argocd application expects that the label 'app.kubernetes.io/part-of' is set to 'argocd' otherwise the configmap or secret is rejected, and the argocd-dex-server, argocd-server, and argocd-application-controller fail to start.

Checklist:


jdeprin:
Would it be better to update all instances of this?  Only changing specific files will lead to drift.  The alternative is to just update the _helpers.tpl file to be a static `argocd` value.

TomHellier:
> Would it be better to update all instances of this? Only changing specific files will lead to drift. The alternative is to just update the _helpers.tpl file to be a static `argocd` value.

I'm happy to update the _helpers.tpl file to contain a special case for the `app.kubernetes.io/part-of` and update every part-of to use that value. The changes made here are very useful for deploying multiple argocds into a single k8s cluster because the ClusterRole and ClusterRoleBindings clash otherwise. 

I have no idea why that line I linked in the issue I raised has this check on the `part-of` label, do you think it is worth raising an issue on the main argocd repository?

TomHellier:
Closing because I don't know how to use github yet apparently.

PR #151, opened 2019-11-05
Issue #147, k8s label part-of expected to be argocd

TomHellier:
The argocd application expects that the label 'app.kubernetes.io/part-of' is set to 'argocd' otherwise the configmap or secret is rejected, and the argocd-dex-server, argocd-server, and argocd-application-controller fail to start.

[Issue 147](https://github.com/argoproj/argo-helm/issues/147)
Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.2.4"
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 1.0.1
+version: 1.0.2
 home: https://github.com/argoproj/argo-helm
 icon: https://raw.githubusercontent.com/argoproj/argo/master/argo.png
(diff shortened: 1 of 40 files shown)


TomHellier:
I've updated all labels in the application to `app.kubernetes.io/part-of: argocd` as per [this comment on a prior pull request](https://github.com/argoproj/argo-helm/pull/149#issuecomment-550041100). I've tested this on my own installation of argocd, which we had deploying another argocd, (argocd-ref) in another namespace.

This change is only really neccesary in the configmaps and secrets because of [this commit on the main argocd application](https://github.com/argoproj/argo-cd/commit/9cf744f4357230fe119628f4ca8976b6fe146d42#diff-1e8041f22bda5d97c39f9ff6d190ab88R399), where the `part-of` label has to be `argocd`

seanson:
I would call this a good change for every resource as the approach for `app.kubernetes.io/part-of` appears to be the overall software package as per https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Issue #154, opened 2019-11-06
Access app resource from repo-server -> add RBAC to the repo-server


fiftoine:
**Is your feature request related to a problem? Please describe.**
I'm writing a plugin for argocd config management. That plugin needs to read the argo app details (like destination namespace, annotations, labels,...) and I don't want to put those values as env for each of the apps that will use that plugin. 

**Describe the solution you'd like**
Adding configurable RBAC for the repo-server component would allow to execute `kubectl` commands and read the app manifest.

**Describe alternatives you've considered**
I forked the chart to add the SA and associated RBAC and put it into a private helm chart repo.

**Additional context**

seanson:
Is this a common enough use case that we should be providing a service account template in the Helm chart itself, or would it be enough to allow a single Helm value pointing to an externally created serviceAccount as is a common pattern in other charts?

seanson:
Tested and merged, thanks for raising the issue!

PR #160, opened 2019-11-07
Repo server rbac

fiftoine:
Checklist:

* [V] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [V] Any new values are backwards compatible and/or have sensible default.
* [V] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md).
* [V] I have signed the CLA and the build is green.
* [V] I will test my changes again once merged to master and published.

### DIFF
--- a/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
@@ -105,6 +105,7 @@ spec:
       affinity:
 {{- toYaml .Values.repoServer.affinity | nindent 8 }}
     {{- end }}
+      serviceAccountName: {{ template "argo-cd.repoServerServiceAccountName" . }}
       volumes:
       {{- if .Values.repoServer.volumes }}
 {{- toYaml .Values.repoServer.volumes | nindent 8}}
--- a/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
@@ -105,9 +105,7 @@ spec:
       affinity:
 {{- toYaml .Values.repoServer.affinity | nindent 8 }}
     {{- end }}
-    {{- if .Values.repoServer.serviceAccount.create }}
       serviceAccountName: {{ template "argo-cd.repoServerServiceAccountName" . }}
-    {{- end }}
       volumes:
       {{- if .Values.repoServer.volumes }}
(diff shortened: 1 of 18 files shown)


fiftoine:
Solving issue https://github.com/argoproj/argo-helm/issues/154

seanson:
Looks good, just needs a chart version bump!

fiftoine:
> Looks good, just needs a chart version bump!

Bumped ;-)

charts/argo-cd/templates/argocd-repo-server/deployment.yaml
```diff
@@ -104,6 +104,9 @@ spec:
     {{- if .Values.repoServer.affinity }}
       affinity:
 {{- toYaml .Values.repoServer.affinity | nindent 8 }}
+    {{- end }}
+    {{- if .Values.repoServer.serviceAccount.create }}
```
seanson: Could you please remove this conditional so we can support the three use cases:
- create: false, service account name unset: template defaults to serviceAccountName: "default"
- create: false, service account name set: template uses specific service account name (common use case for externally created service accounts)
- create: true: service account name set: new service account created

charts/argo-cd/templates/argocd-repo-server/serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.repoServer.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "argo-cd.repoServerServiceAccountName" . }}
+  labels:
```
seanson: Could you please add the option to set annotations on the service account. This is required for things like workload identity on Google Cloud and AWS and would be very helpful for any reposerver that needs to pull info from S3 or google storage.

charts/argo-cd/values.yaml
```diff
@@ -480,6 +480,23 @@ repoServer:
     #   namespace: monitoring
     #   additionalLabels: {}
 
+  ## Repo server service account
+  ## If create is set to true, make sure to uncomment the name and update the rbac section below
+  serviceAccount:
+    create: false
+  #  name: argocd-repo-server
```
seanson: As per above, needs a new field for service account annotations.

fiftoine:
> Awesome work @fiftoine! I did some testing on my side and there are a few small changes we'd need to make in order to cover some common reposerver service account use cases.

All done ;-)

PR #171, opened 2019-11-21
[charts/argo-cd] Fix reference to cert config

masondaven:
Fix for: #171 

Checklist:

* [X] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [X] Any new values are backwards compatible and/or have sensible default.
* [x] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md).
* [x] I have signed the CLA and the build is green.
* [x] I will test my changes again once merged to master and published.

Changes are automatically published when merged to `master`. They are not published on branches.

PR #173, opened 2019-11-22
Take certificate, route and ingress host from Argo CD server URL config

yann-soubeyrand:
Checklist:


yann-soubeyrand:
This PR is not ready yet, I opened it for discussion before finalizing it.

seanson:
What's the primary driver for this change?

yann-soubeyrand:
Since ingress host and path must be identical to the ones in Argo CD external URL for the latter to be reachable, we should not have separate settings for it. The same reasoning holds true for certificate domain and route host. This way we eliminate potential configuration errors (in addition to not having to repeat ourselves).

yann-soubeyrand:
@seanson it also fixes the certificate which is broken at the time.

seanson:
Certificate path fix is pending in #171 
I appreciate the effort but I'm going to reject this PR in preference to keeping things simple and explicit even if redundant. It doesn't look like any of the official Helm repo charts use `regexReplaceAll`, other than some dashboard name mangling in the Grafana chart.

PR #174, opened 2019-11-25
Correct the values.yaml comment for extra arg

rytswd:
Extra arg handling changed in v1 release #129 , and expects map rather than list.
Just updating the comment to align with that expectation.

With the previous doc, I got the following in deployment:

```
      containers:
      - command:
        - argocd-server
        ...
        - --0=map[insecure:true]
        image: argoproj/argocd:v1.3.0
```

And got a straight error when a pod started:

```
time="2019-11-25T11:24:29Z" level=fatal msg="unknown flag: --0"
```

Checklist:


rytswd:
I wasn't sure if it would make sense to bump the chart version when only updating the comments - as CI was failing, I did it anyhow as a separate commit. Please let me know if it's better to revert.

PR #178, opened 2019-12-01
[argo-cd] Added simple PrometheusRule

mnaser:
This patch adds a few simple rules which you can leverage to make
sure that your infrastructure is up to date.  They are optional
and opt-in only.

There is not a lot of customization initially but this is a start
of a simple framework to deliver this.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.3.0"
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 1.2.2
+version: 1.4.1
 home: https://github.com/argoproj/argo-helm
 icon: https://raw.githubusercontent.com/argoproj/argo/master/argo.png
 keywords:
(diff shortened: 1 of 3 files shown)


mnaser:
Hi there,

I was hoping that we can iterate onto this a step down the line if that’s cool :)

If we can land this then I totally can do the extra work to slowly make it more and more configurable. 

Thanks,
Mohammed

seanson:
I would prefer to have an update with a standard configurable change than intermediate versions later. It should be a fairly trivial change:

- Move spec.groups list to values.yaml
- Replacing the spec.groups list with `{{- toYaml .Values.controller.metrics.rules | nindent <x> }}`

mnaser:
voila :)

seanson:
Awesome, thanks for the change @mnaser! One last request, the sample rules in `values.yaml` should be commented out and the default `spec` value should be `[]` as Prometheus alerts are likely to vary wildly by environment in requirements. These rules are awesome but should be up to the user to add if they'd like.
Should be good to merge after that!

mnaser:
donezo :+1:

seanson:
Looks good, merged an upstream selector fix. Thanks for the work @mnaser!

mnaser:
yay, thanks!

PR #199, opened 2019-12-31
feat: allow custom secrets to be added to argocd-secret

DaytonG:
Checklist:


DaytonG:
My particular use case: I need to configure LDAP with dex, and our bind password contains "$". Because of that, we need to feed in the value via environmental variables. According to https://argoproj.github.io/argo-cd/operator-manual/sso/, argocd-secret is mounted to the container and available in the dex.config.

gageorsburn:
I have the same use-case and would love for this to get merged. 😸

Issue #172, opened 2019-11-22
[argo-helm/argo-cd] add HA functionality


jdeprin:
**Is your feature request related to a problem? Please describe.**
Currently there exists good documentation for how to run ArgoCD in HA mode but the existing chart does not support it.

**Describe the solution you'd like**
Add in HA functionality.  I believe this can be done while keeping backwards compatibility with the v1.*.* chart version. 

**Describe alternatives you've considered**
N/A

**Additional context**
I'll probably work on this over the next couple weeks.

nouseforaname:
Hello,

did you do anything on that story already? Our team currently has the same thing on the roadmap and we could take over if you did not have time to work on this yet.
Let me know if you'd like some support

papanito:
The docu helm chart of Argo-CD currently states
> This chart currently installs the non-HA version of ArgoCD.

So is this statement still true? If yes, what is missing? Glad to help though

spencergilbert:
I suspect this is just an issue with the documentation - the chart can currently deploy HA redis and HA for the rest of the components can just be controlled via the replica values

PR #232, opened 2020-02-06
feat: ArgoCD HA chart reusing existing charts

carlosjgp:
Would say that this is far from ideal but it renders a set of resources that are equivalent to the [manifests provided on the argo-cd repository](https://github.com/argoproj/argo-cd/tree/v1.4.2/manifests/ha)

Fixes (ish) #172 

I would like to have some feedback about this approach that is not as neat and nice as the [manifest rendered with Kustomize](https://github.com/argoproj/argo-cd/tree/v1.4.2/manifests/ha)

The other approach could be to create the Helm chart from scratch or maybe using Kustomize to implement the shared bits for HA and non HA...

Thoughts?

Signed-off-by: Carlos Juan Gómez Peñalver <carlosjuangp@gmail.com>

Checklist:


cabrinha:
I'm not sure if it'll make much sense seeing as how we can reference the argo-cd chart, but adding a README in here would be nice: https://github.com/norwoodj/helm-docs

seanson:
I think I'd rather have this configured in the ArgoCD via flags as maintaining two charts would be difficult.

carlosjgp:
> I think I'd rather have this configured in the ArgoCD chart via flags as maintaining two charts would be difficult.

Sounds good.

I could use the Redis HA stable on the ArgoCD chart. I'll update this PR to do so.

Will this change be ok for the Argo team?

charts/argo-cd-ha/values.yaml
```diff
@@ -0,0 +1,53 @@
+argo-cd:
+  installCRDs: false
+  controller:
+    extraArgs:
+      sentinel=argocd-redis-ha-announce-0:
```
cabrinha: Maybe we should handle the key/value mapping differently if the redis-ha chart is added to the argo-cd chart... Something like if redis-ha.enabled -- We could be more clear about adding these args into the deployments.

cabrinha:
I tried out this setup and ArgoCD seems to be having a problem managing itself and rendering certain parts of this chart:

1. Sentinel configuration via `extraArgs`
Screen Shot 2020-02-19 at 8 11 21 AM

When I render the chart locally via `helm template . -f values.yaml` these sentinel flags render fine. But, when this chart is applied via ArgoCD, these flags fail to render.

This remains true for all three deployment manifests that use redis sentinel. The application-controller, the repo-server and the server.

2. HPA fails to reach a "Synced" state

```
level=warning msg="Failed to convert resource: no kind \"APIService\" is registered for version \"apiregistration.k8s.io/v1\" in scheme \"k8s.io/kubernetes/pkg/api/legacyscheme/scheme.go:30\""
```

Screen Shot 2020-02-19 at 9 30 53 AM

For some reason, ArgoCD feels the need to re-order the HPA manifest to have the memory come before the cpu spec.

carlosjgp:
> I tried out this setup and ArgoCD seems to be having a problem managing itself and rendering certain parts of this chart:

The hacky expression for `extraArgs` is due to this piece of code
https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-application-controller/deployment.yaml#L59-L65
and the fact that this field is using a `map[string]string` instead of just `[]string` which means that you can't use repeated keys (`sentinel`) will be resolved into only one, the last one.
I might raise a ticket for this _breaking change_ separately and then correct this PR.
I haven't checked ArgoCD code but depending on how is marshalling and unmarshalling YAML what is valid for `helm template ...` might not be for ArgoCD
(I'll check this and come back with a more concrete answer for these :point_up: )

Could you provide your K8s cluster details, please?

```
$ kubectl version 
```

> level=warning msg="Failed to convert resource: no kind \"APIService\" is registered for version \"apiregistration.k8s.io/v1\" in scheme \"k8s.io/kubernetes/pkg/api/legacyscheme/scheme.go:30\""

Are you running `metrics-server` ? This message and problems with Horizontal Pod Autoscaler are normally related to old `metrics-server` versions or even using `heaspter` (deprecated now)

cabrinha:
> > I tried out this setup and ArgoCD seems to be having a problem managing itself and rendering certain parts of this chart:
> 
> The hacky expression for `extraArgs` is due to this piece of code
> https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-application-controller/deployment.yaml#L59-L65
> and the fact that this field is using a `map[string]string` instead of just `[]string` which means that you can't use repeated keys (`sentinel`) will be resolved into only one, the last one.
> I might raise a ticket for this _breaking change_ separately and then correct this PR.
> I haven't checked ArgoCD code but depending on how is marshalling and unmarshalling YAML what is valid for `helm template ...` might not be for ArgoCD
> (I'll check this and come back with a more concrete answer for these ☝️ )
> 
> Could you provide your K8s cluster details, please?
> 
> ```
> $ kubectl version 
> ```
> 
> > level=warning msg="Failed to convert resource: no kind "APIService" is registered for version "apiregistration.k8s.io/v1" in scheme "k8s.io/kubernetes/pkg/api/legacyscheme/scheme.go:30""
> 
> Are you running `metrics-server` ? This message and problems with Horizontal Pod Autoscaler are normally related to old `metrics-server` versions or even using `heaspter` (deprecated now)

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-13T18:08:14Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-eks-c0eccc", GitCommit:"c0eccca51d7500bb03b2f163dd8d534ffeb2f7a2", GitTreeState:"clean", BuildDate:"2019-12-22T23:14:11Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

charts/argo-cd-ha/values.yaml
carlosjgp: The "problem" with that is to find the balance between how much you want to hide behind that flag and how much flexibility you give to the user to deploy the chart...
But yes, it's a good idea. I'll try some GoLang templates (aka `_helpers.tpl`) to see if I can get that flexibility and default values

carlosjgp: I raised a PR to improve this weird syntax

https://github.com/argoproj/argo-helm/pull/244

cabrinha:
> > I think I'd rather have this configured in the ArgoCD chart via flags as maintaining two charts would be difficult.
> 
> Sounds good.
> 
> I could use the Redis HA stable on the ArgoCD chart. I'll update this PR to do so.
> 
> Will this change be ok for the Argo team?

On the releases page for ArgoCD, there are two options for installation: HA and Non-HA.

I think this helm chart should leave these two options open to the user, with the [redis-ha](https://github.com/helm/charts/tree/master/stable/redis-ha) chart being enabled for the HA option and the [redis](https://github.com/helm/charts/tree/master/stable/redis) chart being enabled for the Non-HA option.

cabrinha:
This PR brings up some important points: https://github.com/argoproj/argo-cd/pull/3147

cabrinha:
@carlosjgp -- Now that https://github.com/argoproj/argo-cd/pull/3147 has been merged, it's probably time to get a major update to the argo-cd chart that mirrors that setup.

carlosjgp:
Superseded by #270

PR #263, opened 2020-03-11
feat: add support for namespaced install

kristofarkas:
Checklist:


kristofarkas:
Hey @alexec @benjaminws 👋 What do you guys think of this MR?

kristofarkas:
Updated the chart to 2.6.3 too!

nicorikken:
Great. I was still using the `forceNamespaceIsolation: true`, but it seems that this methodology is deprecated. I haven't been able to find confirmation for it, but if so the old way of isolation can be removed as well.

kristofarkas:
Thanks @nicorikken. Removed it. Have you tried this new chart yourself?

nicorikken:
@kristofarkas Yeah, not this specific PR, but the recent master. It works great. The namespacing didn't, but I reintroduced it afterwards by manually adding the `--namespaced` flag, and that works. 
In that regard, https://github.com/argoproj/argo-helm/pull/271 is also helpful, as it exposes all args from the helm chart.

shaikatz:
Hi guys, any way to push that PR forward? will be happy to help.

shaikatz:
@spencergilbert can you approve that please?

stefansedich:
+1 we are in the same boat and this would be useful for us.

charts/argo/templates/server-cluster-role.yaml
```diff
@@ -1,6 +1,10 @@
 {{- if .Values.server.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
+{{- if .Values.singleNamespace }}
+kind: Role
+{{ else }}
 kind: ClusterRole
+{{- end }}
 metadata:
   name: {{ .Release.Name }}-{{ .Values.server.name}}-cluster-role
```
stefansedich, line 9: The name of this and other roles are now slightly confusing, we should either use `-role` or `-cluster-role` based on what it is or drop the suffix altogether.

However this could be a breaking change if others are re-using the existing roles for other purposes...

charts/argo/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "v2.6.1"
+appVersion: v2.7.2
```
stefansedich, line 2: Bump to latest v2.7.6

charts/argo/templates/server-deployment.yaml
```diff
@@ -34,33 +33,28 @@ spec:
...
+          - "--namespaced"
+          {{- end }}
           image: "{{ .Values.images.namespace }}/{{ .Values.images.server }}:{{ default .Values.images.tag .Values.server.image.tag }}"
           imagePullPolicy: {{ .Values.images.pullPolicy }}
           {{- if .Values.server.podPortName }}
           ports:
           - name: {{ .Values.server.podPortName }}
-            ports:
             containerPort: 2746
-            readinessProbe:
```
stefansedich, line 44: I take all this was broken? I am still unclear of the purpose of `podPortName` to switch this on and off and why not just always expose the named port? but might pay not to change things now.

charts/argo/templates/server-deployment.yaml
```diff
@@ -34,33 +33,28 @@ spec:
...
-            value: "true"
-          {{- end }}
           - name: IN_CLUSTER
             value: "true"
           - name: ARGO_NAMESPACE
-            valueFrom:
-              fieldRef:
-                apiVersion: v1
-                fieldPath: metadata.namespace
+            value: {{ .Release.Namespace | quote }}
```
stefansedich, line 57: Any reason for this change?

charts/argo/templates/server-sa.yaml
```diff
@@ -3,4 +3,6 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ .Values.server.serviceAccount }}
+  annotations:
+{{ toYaml .Values.server.serviceAccountAnnotation | indent 4 }}
```
stefansedich, line 7: What about using `servrer.serviceAccount.annotations`?

charts/argo/templates/workflow-controller-sa.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ .Values.controller.serviceAccount }}
+  annotations:
+{{ toYaml .Values.controller.serviceAccountAnnotation | indent 4 }}
```
stefansedich, line 6: What about using `controller.serviceAccount.annotations`?

charts/argo/Chart.yaml
```diff
@@ -1,5 +1,5 @@
...
 description: A Helm chart for Argo Workflows
 name: argo
 version: 0.7.6
```
stefansedich, line 5: Will want to bump the chart version too.

charts/argo/values.yaml
```diff
@@ -4,7 +4,7 @@ images:
   server: argocli
   executor: argoexec
   pullPolicy: Always
-  tag: v2.6.1
+  tag: v2.7.2
```
stefansedich, line 7: Latest is now v2.7.6

charts/argo/templates/workflow-aggregate-roles.yaml
```diff
@@ -1,6 +1,10 @@
 {{- if .Values.createAggregateRoles }}
 apiVersion: rbac.authorization.k8s.io/v1
+{{- if .Values.singleNamespace }}
+kind: Role
+{{ else }}
 kind: ClusterRole
+{{- end }}
```
moensch, line 7: Aggregated roles are always `ClusterRole`: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

Instead, maybe provide a boolean option to `includeAggregateClusterRoles` to avoid this failing when installing onto the same cluster twice?

kristofarkas: Looks like there is already an option for it: `.createAggregateRoles`. Removing the non ClusterRole changes.

charts/argo/templates/server-deployment.yaml
kristofarkas: Given we are using Helm, I thought it would make more sense to use this than `valueFrom`. Less code and easier to understand (imo). I don't know if the two values could ever be different, I suspect not?

charts/argo/templates/server-deployment.yaml
kristofarkas: I just fixed the indents here, it was misaligned. Somebody probably had a use case for it as some point so it was added.

charts/argo/templates/server-cluster-role.yaml
kristofarkas: Agree 100%. Changed all the names to the correct one depending on the `singleNamespace` flag.

charts/argo/templates/workflow-controller-sa.yaml
kristofarkas: Agree that that would make more sense. Currently in the values.yaml `controller.serviceAccount` is reserved for the name `string`, so changing that wouldn't be backwards compatible. Is this what you had in mind:
```
controller:
    serviceAccount:
      name: argo 
      annotations: ~ 
```

charts/argo/templates/server-sa.yaml
kristofarkas: Same comment as below for `controller`

charts/argo/templates/workflow-controller-sa.yaml
stefansedich: Makes sense we don't want to do anything breaking, perhaps we just use `serviceAccountAnnotations` and call it a day.

charts/argo/templates/server-deployment.yaml
stefansedich: This is true, however a scenario I think of is that someone could use `helm template` to generate manifests then apply them into a namespace.

In that case `.Release.Namespace` may end up as `default` as they did not specify the namespace while templating but did when applying.

charts/argo/templates/server-deployment.yaml
stefansedich: Yup makes sense!

charts/argo/values.yaml
```diff
@@ -59,6 +63,8 @@ controller:
     enabled: false
     additionalLabels: {}
   serviceAccount: argo
+  # Service account annotations
+  serviceAccountAnnotation: {}
```
stefansedich, line 67: Lets use `serviceAccountAnnotations` as it is more clear.

charts/argo/values.yaml
```diff
@@ -124,6 +130,8 @@ server:
   servicePort: 2746
   # servicePortName: http
   serviceAccount: argo-server
+  # Service account annotations
+  serviceAccountAnnotation: {}
```
stefansedich, line 134: Lets use `serviceAccountAnnotations` as it is more clear.

PR #264, opened 2020-03-12
fix: More realistic limits/requests for repoServer

jessebye:
Checklist:


spencergilbert:
I'm not sure - I think it's fairly standard to set up the value file as it is today, leaving the responsibility on the user to provision resources based on their particular needs. @seanson what are your thoughts?

jessebye:
@spencergilbert to be clear, this just updates the comment since we naively tried to use those values and encountered [an issue](https://github.com/argoproj/argo-cd/issues/3229) due to lack of sufficient resources. Just figured if anyone else made the same mistake, they would have a better chance of success with these new values.

spencergilbert:
Yep, I've definitely seen fewer charts set "reasonable" defaults in the value file - think it's probably fine to have a lower limit (commented out still) in the default value file.

PR #268, opened 2020-03-16
feat: add imagePullSecrets, labels, annotations and update CRDs

cabrinha:
Checklist:

### DIFF
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.7"
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 0.1.2
+version: 0.2.0
 icon: https://raw.githubusercontent.com/argoproj/argo/master/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.7"
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 0.1.1
+version: 0.1.2
 icon: https://raw.githubusercontent.com/argoproj/argo/master/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
--- a/charts/argo-rollouts/README.md
+++ b/charts/argo-rollouts/README.md
@@ -2,7 +2,7 @@ Argo Rollouts Chart
 =============
 A Helm chart for Argo Rollouts, progressive delivery for Kubernetes.
 
-Current chart version is `0.1.2`
+Current chart version is `0.2.0`
 
 Source code can be found [here](https://github.com/argoproj/argo-rollouts)
 
--- a/charts/argo-rollouts/README.md
+++ b/charts/argo-rollouts/README.md
@@ -2,7 +2,7 @@ Argo Rollouts Chart
 =============
 A Helm chart for Argo Rollouts, progressive delivery for Kubernetes.
 
-Current chart version is `0.1.0`
+Current chart version is `0.1.2`
 
 Source code can be found [here](https://github.com/argoproj/argo-rollouts)
 
@@ -35,5 +35,8 @@ $ helm install --name my-release argo/argo-rollouts
 | controller.image.repository | string | `"argoproj/argo-rollouts"` |  |
 | controller.image.tag | string | `"v0.7.0"` |  |
 | controller.name | string | `"argo-rollouts"` |  |
+| imagePullSecrets | list | `[]` |  |
 | installCRDs | bool | `true` |  |
+| podAnnotations | object | `{}` |  |
+| podLabels | object | `{}` |  |
 | serviceAccount.name | string | `"argo-rollouts"` |  |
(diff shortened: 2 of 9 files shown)


charts/argo-rollouts/templates/argo-rollouts-clusterrole.yaml
```diff
@@ -2,11 +2,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ .Release.Name }}-clusterrole
+  name: argo-rollouts-clusterrole
   labels:
-    app.kubernetes.io/component: {{ .Values.controller.component }}
-    app.kubernetes.io/name: {{ .Release.Name }}-clusterrole
-    app.kubernetes.io/part-of: {{ .Release.Name }}
+    app.kubernetes.io/component: rollouts-controller
```
igaskin: Why are you hardcoding these labels?  That seems undesireable.

charts/argo-rollouts/templates/argo-rollouts-clusterrole.yaml
```diff
@@ -29,6 +29,14 @@ rules:
   - list
   - watch
   - patch
+- apiGroups:
+    - ""
+  resources:
+    - secrets
```
igaskin: Is the intent here to give the `argo-rollouts-clusterrole` full read access to all cluster secrets?  If the intent of this PR is to add the ability to run the controller using images from a private registry, I think we can do without this?  I'm reading the docs, and it looks like we can add imagepull secrets to a service account if need be: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

charts/argo-rollouts/templates/argo-rollouts-clusterrole.yaml
cabrinha: Ah, good catch. I'll revert that change and keep the templated values.

charts/argo-rollouts/templates/argo-rollouts-clusterrole.yaml
cabrinha: This change comes from the AnalysisTemplates being able to reference secrets: https://github.com/argoproj/argo-rollouts/pull/420

This originally came up for use with the WebMetric provider, and I've recently been using it with the datadog metric provider I'm testing.

charts/argo-rollouts/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.7"
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 0.1.1
+version: 0.1.2
```
eytanhanig:
```suggestion
version: 0.2.0
```

This is a feature add. How about incrementing the minor version instead of the patch?

charts/argo-rollouts/templates/argo-rollouts-clusterrole.yaml
```diff
@@ -29,6 +29,14 @@ rules:
   - list
   - watch
   - patch
+- apiGroups:
+    - ""
```
eytanhanig: Should this be an empty list instead? ie
```yaml
 - apiGroups: {}
```

charts/argo-rollouts/templates/argo-rollouts-deployment.yaml
```diff
@@ -2,10 +2,19 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ .Release.Name }}
+  {{- if .Values.podAnnotations }}
+  annotations:
+  {{- range $key, $value := .Values.podAnnotations }}
+    {{ $key }}: {{ $value | quote }}
```
eytanhanig: I like that you used `| quote`. Is there any reason that you decided not to use it elsewhere in the PR?

cabrinha: I think it's just to make the quotation a bit more explicit in this one section.

charts/argo-rollouts/templates/argo-rollouts-clusterrole.yaml
cabrinha: from: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole

```yaml
kind: Role
metadata:
  namespace: default
  name: pod-reader
rules:
- apiGroups: [""] # "" indicates the core API group
  resources: ["pods"]
  verbs: ["get", "watch", "list"]
```

charts/argo-rollouts/README.md
```diff
@@ -2,7 +2,7 @@ Argo Rollouts Chart
 =============
 A Helm chart for Argo Rollouts, progressive delivery for Kubernetes.
 
-Current chart version is `0.1.0`
+Current chart version is `0.1.2`
```
eytanhanig:
```suggestion
Current chart version is `0.2.0`
```

spencergilbert:
Not sure why I got added to review here, sorry. Haven't used rollouts before!

charts/argo-rollouts/templates/argo-rollouts-deployment.yaml
```diff
@@ -2,10 +2,19 @@ apiVersion: apps/v1
...
+  {{- end }}
+  {{- end }}
   labels:
     app.kubernetes.io/component: {{ .Values.controller.component }}
     app.kubernetes.io/name: {{ .Release.Name }}
     app.kubernetes.io/part-of: {{ .Release.Name }}
+    {{- if .Values.podLabels }}
+{{- toYaml .Values.podLabels | nindent 4 }}
```
spencergilbert: Labels also have to be strings too right? Do we want to ensure this with the `quote` function again?

cabrinha: We certainly can, I just don't see it as often.

PR #271, opened 2020-03-19
feat: Add support for extra arguments to Argo Server

vladlosev:
This PR provides a way to specify additional command line arguments for `argo-server`. This gives chart users the flexibility in invoking `argo-server` without having to modify the chart.

One of the ways this can be used is to specify a custom config map. Currently, the chart creates a config map for the workflow controller and configures its command line to use it. It does not configure `argo-server` to use that map, however. That means `argo-server` installed via the chart cannot pick up persistence database configuration  and is unable to show archived workflows. Being able to specify a config map via extraArgs helps mitigate issues like this.

Checklist:

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "v2.6.1"
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.7.2
+version: 0.7.3
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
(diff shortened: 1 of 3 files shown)


nicorikken:
Nice, and very flexible!

nicorikken:
Is there a list of flags somehwere? I'd like for the server to use a default serviceaccount when I submit a job through the YAML UI. Is that a commandline option? In that case we could also expose it through Helm.

vladlosev:
The command line options for the server are specified [here](https://github.com/argoproj/argo/blob/96c80e3e2c6eb6867e360dde3dea97047b963c2f/cmd/argo/commands/server.go#L106). Unfortunately, there is no way to specify a default service account. One workaround you could use is to employ WorkflowTemplates - I believe they have the service account as part of the spec.

Issue #277, opened 2020-03-24
Argo server role should allow Pod delete for retry


kristofarkas:
The argo server role needs to be able to delete pods, as this is how the Retry button works.

https://github.com/argoproj/argo-helm/blob/805371544d5ff408e0da4b10a97b61002729959c/charts/argo/templates/server-cluster-role.yaml#L23-L30

stefansedich:
With #278 merged this should be fixed! thanks @kristofarkas

PR #278, opened 2020-03-24
fix: update server-cluster-role.yaml

kristofarkas:
Checklist:


kristofarkas:
Fixes https://github.com/argoproj/argo-helm/issues/277

charts/argo/templates/server-cluster-role.yaml
```diff
@@ -28,6 +28,9 @@ rules:
   - get
   - list
   - watch
+  - update
+  - patch
+  - delete
```
stefansedich, line 33: Should we only be adding `delete` here? looking at the manifests over in the Argo repo https://github.com/argoproj/argo/blob/3f8be0cd48963958c493e7669a1d03bb719b375a/manifests/namespace-install/argo-server-rbac/argo-server-role.yaml#L30 looks like it only has `get,list,watch,delete`.

stefansedich:
@kristofarkas mind rebasing, resolving the conflicts and bumping the chart version.

charts/argo/templates/server-cluster-role.yaml
stefansedich: If you could also rebase, fix the conflicts and bump the chart version in `Chart.yaml` that would be great :smile:

kristofarkas: Done!

charts/argo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "v2.7.6"
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.8.3
+version: 0.8.5
```
stefansedich, line 5: Looks like we skipped 0.8.4, want to bump down and I will get this in!

stefansedich: Nevermind I was able take care of it!

PR #286, opened 2020-03-30
feat: add support for setting s3 region for artifact repository

stefansedich:
Adding support for setting the region for the artifact repository when using s3 as we use buckets not in us-east-1 and there is currently no way to set the region on the workflow controller configmap.

Checklist:


stefansedich:
@benjaminws just bumping this one, any objections to the change? we run in a non us-east-1 region and this was needed so I could set a default artifcact repo.

PR #291, opened 2020-04-02
feat(argo-cd): Add shareProcessNamespace option to repoServer

k:
Checklist:


spencergilbert:
@k can you tell me a little more about the use case here? I'm not super familiar with that spec, but it seems like it would only be useful if you're running multiple containers in the repo-server pod. Are you using something like an istio-sidecar and running into problems?

charts/argo-cd/templates/argocd-repo-server/deployment.yaml
```diff
@@ -39,6 +39,9 @@ spec:
 {{- toYaml .Values.repoServer.podLabels | nindent 8 }}
         {{- end }}
     spec:
+      {{- if .Values.repoServer.shareProcessNamespace }}
+      shareProcessNamespace: true
+      {{- end }}
```
seanson, line 44: Please change the invocation of the template to `with` so it's easier to template and future proof.
```suggestion
      {{- with .Values.repoServer.shareProcessNamespace }}
      shareProcessNamespace: {{ . }}
     {{- end }}
```

alexec:
Stale.

PR #296, opened 2020-04-05
feat: upgrade Argo to v2.7.6

### LINKED ISSUE #287: argo: argo-server is not watching the right configmap
**Expected behavior**
argo-server should watch the right configmap and not the one with the default name ```workflow-controller-configmap``` 

A quick fix should be to add something like this in argo-server container args:

```
--configmap={{ .Release.Name }}-{{ .Values.controller.name }}-configmap
```

sebastien-prudhomme:
Checklist:

Also fix #287 because v2.7.0 is not working if the default configmap does not exists.


clementlecorre:
👍

sebastien-prudhomme:
@alexec @alexmt @jessesuen @benjaminws can a maintainer take a few minutes to approve this PR?

alexec:
v2.7.5 is the latest stable version?

sebastien-prudhomme:
@alexec updated to 2.7.5

meurig:
I'd also welcome this being reviewed/approved, pretty please? @alexec @alexmt @jessesuen @benjaminws

charts/argo/templates/server-deployment.yaml
```diff
@@ -31,6 +31,7 @@ spec:
         - name: argo-server
           args:
           - server
+          - --configmap={{ .Release.Name }}-{{ .Values.controller.name }}-configmap
```
stefansedich, line 34: Is this intended to be part of this PR? also https://github.com/argoproj/argo-helm/pull/271 should allow you to set this using `extraArgs` now if desired.

charts/argo/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "v2.6.1"
+appVersion: "v2.7.5"
```
stefansedich, line 2: Latest is v2.7.6 now, want to bump?

charts/argo/templates/server-deployment.yaml
sebastien-prudhomme: Yes, there is a configmap in the chart that is not using the default configmap name. First versions of 2.7.x didn't even start without this option because Argo couldn't find the configmap.

stefansedich: Yup makes sense, best be sensible by default.

PR #329, opened 2020-04-30
feat: add support for setting roleARN and useSDKCreds properties for S3 artifact repository

ygapon-mio:
Checklist:


charts/argo/templates/workflow-controller-config-map.yaml
```diff
@@ -51,4 +56,4 @@ data:
 {{ toYaml .Values.controller.persistence | indent 6 }}{{- end }}
     {{- if .Values.controller.workflowDefaults }}
     workflowDefaults:
-{{ toYaml .Values.controller.workflowDefaults | indent 6 }}{{- end }}
+{{ toYaml .Values.controller.workflowDefaults | indent 6 }}{{- end }}
```
stefansedich, line 59: The newline should stay, I think the problem is that above you are missing an {{- end }}

charts/argo/templates/workflow-controller-config-map.yaml
```diff
@@ -38,6 +38,11 @@ data:
         {{- end }}
         {{- if .Values.artifactRepository.s3.region }}
         region: {{ .Values.artifactRepository.s3.region }}
+        {{- if .Values.artifactRepository.s3.roleARN }}
+        roleARN: {{ .Values.artifactRepository.s3.roleARN }}
```
stefansedich, line 42: Missing {{- end }} here which is why you are seeing the lint error.

stefansedich:
@ygapon-mio if you can just add back the removed EOF newline will get this one in!

charts/argo/templates/workflow-controller-config-map.yaml
ygapon-mio: Oops. Fixed. Added {{-end}} and reverted commit to remove end line

ygapon-mio:
Thank you for quick review and approval :)

PR #346, opened 2020-05-13
fix: update CRDs

yann-soubeyrand:
Argo CD has been updated without updating the CRDs. This PR fixes it.

Checklist:


johnmarcou:
> Thanks for the catch, we should find a better way to track when the CRDs are changed 🤔

What about ArgoCD to install the CRD by itself on start?

PR #353, opened 2020-05-18
Added initcontainers to server deployment.yaml

smark88:
Checklist:

Changes: Added initcontainers to server deployment.yaml

This gives the ability to have an init container on deploy, so if you are going to use custom plugin you can make your `deployment.yaml` similar to the one in the [documentation](https://argoproj.github.io/argo-cd/operator-manual/custom_tools/) 

```
    spec:
      # 1. Define an emptyDir volume which will hold the custom binaries
      volumes:
      - name: custom-tools
        emptyDir: {}
      # 2. Use an init container to download/copy custom binaries into the emptyDir
      initContainers:
      - name: download-tools
        image: alpine:3.8
        command: [sh, -c]
        args:
        - wget -qO- https://storage.googleapis.com/kubernetes-helm/helm-v2.12.3-linux-amd64.tar.gz | tar -xvzf - &&
          mv linux-amd64/helm /custom-tools/
        volumeMounts:
        - mountPath: /custom-tools
          name: custom-tools
      # 3. Volume mount the custom binary to the bin directory (overriding the existing version)
      containers:
      - name: argocd-repo-server
        volumeMounts:
        - mountPath: /usr/local/bin/helm
          name: custom-tools
          subPath: helm
```

### DIFF
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -342,6 +342,15 @@ server:
   ##
   podLabels: {}
 
+  ## initContainer
+  initContainers: {}
+    ## - name: download-tools
+    ##   image: alpine:3.8
+    ##   command: [sh, -c]
+    ##   args:
+    ##   - wget -qO- https://storage.googleapis.com/kubernetes-helm/helm-v2.12.3-linux-amd64.tar.gz | tar -xvzf - &&
+    ##     mv linux-amd64/helm /custom-tools/
+
   ## Configures the server port
   containerPort: 8080
 
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -343,7 +343,7 @@ server:
   podLabels: {}
 
   ## initContainer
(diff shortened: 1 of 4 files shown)


spencergilbert:
Based on the docs it looks like the tooling just needs to be loaded to the repoServer, which should be possible here? https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml#L717

charts/argo-cd/values.yaml
```diff
@@ -342,6 +342,15 @@ server:
   ##
   podLabels: {}
 
+  ## initContainer
+  initContainers: {}
```
seanson, line 346:
```suggestion
  initContainers: []
```
This should be a list instead of a map.

jtcarnes:
What is the status of this? I know for a number of applications for secret management additions (KSOPS, Helm secrets), the init containers would be a huge help.

alexec:
Stale.

PR #358, opened 2020-05-22
fix: take argocd serviceaccount flags into account

### LINKED ISSUE #354: [argo-cd] `serviceAcccount.create` flags are ignored
**Describe the bug**

```console
$ helm install argocd argo/argocd --set=server.serviceAccount.create=false`
```

yields

```
Unable to continue with install: existing resource conflict: namespace: argocd, name: argocd-server, existing_kind: /v1, Kind=ServiceAccount
```

due to a missing check here: https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-server/serviceaccount.yaml#L1

**Expected behavior**

No conflicts when a service account already exists and `*.serviceAccount.create` is set to `false`.

hiddeco:
Fixes #354


alexmt:
Thank you @hiddeco !

PR #362, opened 2020-05-27
fix: remove Argo CD server route status field

yann-soubeyrand:
Status fields prevent resources from converging.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.5.4"
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 2.3.5
+version: 2.3.6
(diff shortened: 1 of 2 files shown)


spencergilbert:
I'm not super familiar with openshift resources, @yann-soubeyrand I'm guessing the status spec gets updated live in-cluster and shouldn't be defined ahead of time?

yann-soubeyrand:
@spencergilbert you're guessing right, route's status has the same role as in other Kubernetes resources: carry information between miscellaneous controllers acting on the resource. Therefore, when you include it in your manifests, it prevents Argo CD from converging as it'll always say there are diffs on this field. I'm wondering if Argo CD shouldn't include a global ignore rule for the status field (and I'm wondering if I've not seen discussions about it yet) :thinking:

yann-soubeyrand:
> Sorry for the long delay, life's been hitting me hard lately. Thanks for the PR

No problem. Sorry to hear that, I hope things will get better for you.

Issue #323, opened 2020-04-27
Argo-server does not show any archived workflows


olevski:
**Describe the bug**
Argo-server does not show archived workflows in the web UI - even though they appear in the database and the workflow controller says that workflow archiving has been enabled.

**To Reproduce**
1. Create a new GKE cluster on GCP by using web UI (k8s version 1.14.10-gke27, single zone: us-west-2a, 1 node, no autoscaling, n1-standard-4, defaults for everything else)
2. Connect kubectl context
`gcloud container clusters get-credentials <cluster-name> --zone <zone-name> --project <gcp-project-name>`
3. Install postgres with default values from helm 
`helm install argo-workflows-db bitnami/postgresql`
4. Create secret with database username and password as required by argo for workflow persistence
```bash
export POSTGRES_PASSWORD=$(kubectl get secret --namespace default argo-workflows-db-postgresql -o jsonpath="{.data.postgresql-password}" )
# see additional context below for secrets k8s spec file
k apply -f db-secret.yaml
```
5. Install argo `helm install argo-test argo/argo -f values.yaml`. See [gist here](https://gist.github.com/olevski/306a3eb05abf6cb98b905b8105d6cca5) for values.yaml file. The only difference from default is the persistence section which I uncommented and properly aligned. The whole persistence section is not properly aligned btw - I will submit a PR for this.
6. Edit clusterrole so that controller can get the postgres secret. Will submit an issue for this as the helm chart currently allows only the server to get access to secrets but not the workflow controller, and the workflow controller needs access to at least the secret with the postgres username and password for persistence to work. 
    - `kubectl edit clusterrole argo-test-workflow-controller-cluster-role`
    - paste the following at the end and save
```yaml
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
```
7. Run [hello world workflow](https://github.com/argoproj/argo/blob/master/examples/README.md#hello-world) `argo submit --serviceaccount argo hello-world.yaml`

**Expected behavior**
Once the workflow completes the postgres database is populated with information of the workflow that completed and the workflow should appear in the 'Archived workflows' section of the website from argo-server. However this does not happen and the workflow information appears only in the database but not in the archived workflows section.
```
➜  ~ export POSTGRES_PASSWORD=$(kubectl get secret --namespace default argo-workflows-db-postgresql -o jsonpath="{.data.postgresql-password}" | base64 --decode)
[...]

PR #363, opened 2020-05-27
Fix persistence archive option

wxmiked:
This resolves issue #363. It's a pretty minimal fix. Details of this fix can be found in [this comment](https://github.com/argoproj/argo-helm/issues/323#issuecomment-631158261) for that same issue.

Checklist:

### DIFF
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -34,7 +34,6 @@ controller:
     path: /metrics
     port: 8080
   persistence:
-    archive: false
   # connectionPool:
   #   maxIdleConns: 100
   #   maxOpenConns: 0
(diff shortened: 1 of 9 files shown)


charts/argo/values.yaml
```diff
@@ -33,7 +33,8 @@ controller:
     enabled: false
     path: /metrics
     port: 8080
-  persistence: {}
+  persistence:
+    archive: false
```
stefansedich, line 37: Is the point here just to document this is an option? if so default is already false? perhaps we can just add it as a comment to document the option.

wxmiked: @stefansedich you're right. A commented out version of this would be just fine. I will make this change.

wxmiked: @stefansedich moved to a comment in the latest commit.

stefansedich: Thanks! mind just rebasing and bumping the chart version to the next available?

wxmiked: @stefansedich turns out the exact same issue was raised in this [accepted PR](https://github.com/argoproj/argo-helm/pull/363). So there's no reason to merge this. Thanks for reviewing this tho. I'm closing this MR.

stefansedich: Oops my bad I should have remembered this considering I merged that one :laughing: thanks again!

PR #364, opened 2020-06-01
fix: add pod and workflow workers to argo helm values

AntoineDao:
Checklist:

### DIFF
--- a/charts/argo/templates/workflow-controller-deployment.yaml
+++ b/charts/argo/templates/workflow-controller-deployment.yaml
@@ -43,13 +43,13 @@ spec:
           {{- if .Values.singleNamespace }}
           - "--namespaced"
           {{- end }}
-          {{- if .Values.controller.scaling.workflowWorkers }}
+          {{- with .Values.controller.workflowWorkers }}
           - "--workflow-workers"
-          - "{{ .Values.controller.scaling.workflowWorkers }}"
+          - {{ . | quote }}
           {{- end }}
-          {{- if .Values.controller.scaling.podWorkers }}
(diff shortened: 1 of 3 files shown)


charts/argo/values.yaml
```diff
@@ -57,6 +57,9 @@ controller:
   #  spec:
   #    ttlStrategy:
   #      secondsAfterCompletion: 84600
+  scaling: {}
```
stefansedich, line 60: I am not sure the `scaling` secion is 100% nescassary and keeping it directly under `controller` would suffice.

charts/argo/templates/workflow-controller-deployment.yaml
```diff
@@ -43,6 +43,14 @@ spec:
           {{- if .Values.singleNamespace }}
           - "--namespaced"
           {{- end }}
+          {{- if .Values.controller.scaling.workflowWorkers }}
```
stefansedich, line 46: You can use {{ with }} here to simplify things:

```
{{- with .Values.controller.podWorkers }}
- "--pod-workers"
- {{ . | quote }}
{{- end }}
```

charts/argo/values.yaml
AntoineDao: Yep makes sense. Was thinking of making it "thematic" but it probably makes for unnecessary nesting.

charts/argo/templates/workflow-controller-deployment.yaml
AntoineDao: Yep definitely cleaner. Will implement.

PR #381, opened 2020-06-18
feat(argo): Adds SSO configuration for Argo Server.

vladlosev:
This change will allow to configure the Argo Server SSO [support](https://github.com/argoproj/argo/pull/2745) (new feature in Argo 2.9).

This should only be merged after Argo v2.9.0 is released.

Checklist:

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: v2.8.0
+appVersion: v2.9.0
 description: A Helm chart for Argo Workflows
 name: argo
 version: 0.9.7
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: v2.9.0
+appVersion: v2.8.0
 description: A Helm chart for Argo Workflows
 name: argo
 version: 0.9.7
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v2.8.0
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.9.7
+version: 0.9.8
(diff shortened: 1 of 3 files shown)


charts/argo/templates/workflow-controller-config-map.yaml
```diff
@@ -67,3 +67,7 @@ data:
     {{- if .Values.controller.workflowDefaults }}
     workflowDefaults:
 {{ toYaml .Values.controller.workflowDefaults | indent 6 }}{{- end }}
+    {{- with .Values.server.sso }}
+    server:
```
stefansedich, line 71: Looking at the docs, `sso` does not sit under `server` and sits directly under `data` https://github.com/argoproj/argo/blob/2dae724496a96ce2e0993daea0a3b6a473f784da/docs/workflow-controller-configmap.yaml

charts/argo/values.yaml
```diff
@@ -4,7 +4,7 @@ images:
   server: argocli
   executor: argoexec
   pullPolicy: Always
-  tag: v2.7.6
+  tag: v2.9.0-rc2
```
stefansedich, line 7: I would leave this out, we don't want the chart to deploy RC2 by default, if someone wants to use RC2 they should just set the tag themselves.

charts/argo/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v2.8.0
+appVersion: v2.9.0
```
stefansedich, line 2: Should be left as 2.8.0.

charts/argo/templates/workflow-controller-config-map.yaml
vladlosev: Ah, good good point - thanks for the catch! Updated.

charts/argo/values.yaml
vladlosev: Done.

charts/argo/Chart.yaml
vladlosev: Done.

PR #383, opened 2020-06-18
fix: Add support for SSO in Argo

wnagele:
Checklist:

### DIFF
--- a/charts/argo/templates/server-deployment.yaml
+++ b/charts/argo/templates/server-deployment.yaml
@@ -63,6 +63,14 @@ spec:
             value: {{ .Values.server.baseHref | quote }}
           resources:
             {{- toYaml .Values.server.resources | nindent 12 }}
+          {{- with .Values.server.volumeMounts }}
+          volumeMounts:
+            {{- toYaml . | nindent 12}}
+          {{- end }}
+      {{- with .Values.server.volumes }}
+      volumes:
+        {{- toYaml . | nindent 8}}
+      {{- end }}
       {{- with .Values.server.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/argo/templates/workflow-controller-config-map.yaml
+++ b/charts/argo/templates/workflow-controller-config-map.yaml
@@ -67,3 +67,6 @@ data:
     {{- if .Values.controller.workflowDefaults }}
     workflowDefaults:
 {{ toYaml .Values.controller.workflowDefaults | indent 6 }}{{- end }}
+    {{- with .Values.controller.sso }}
+    sso:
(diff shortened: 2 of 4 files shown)


charts/argo/templates/server-deployment.yaml
```diff
@@ -63,6 +63,14 @@ spec:
             value: {{ .Values.server.baseHref | quote }}
           resources:
             {{- toYaml .Values.server.resources | nindent 12 }}
+          volumeMounts:
+          {{- if .Values.server.volumeMounts }}
+  {{- toYaml .Values.server.volumeMounts | nindent 12}}
```
stefansedich, line 68: We normally indent in the `toYaml` to where it would sit so things are easier to ready, mind doing that?

stefansedich, line 68: Mind indenting these to where they would normally sit? potentionaly using a `with` block like the done for `tolerations`?

charts/argo/templates/workflow-controller-config-map.yaml
```diff
@@ -67,3 +67,6 @@ data:
     {{- if .Values.controller.workflowDefaults }}
     workflowDefaults:
 {{ toYaml .Values.controller.workflowDefaults | indent 6 }}{{- end }}
+    {{- if .Values.controller.sso }}
```
stefansedich, line 70: I suggest a `with` block too for consistency and to clean things up and indenting the toYaml to the correct place.

charts/argo/templates/server-deployment.yaml
wnagele: done

charts/argo/templates/workflow-controller-config-map.yaml
wnagele: done

wnagele:
Found indent problems while testing the `sso` part. Helm templates and indents are a mess. Anyway, now fixed and tested.

charts/argo/templates/server-deployment.yaml
```diff
@@ -63,6 +63,14 @@ spec:
             value: {{ .Values.server.baseHref | quote }}
           resources:
             {{- toYaml .Values.server.resources | nindent 12 }}
+          volumeMounts:
```
stefansedich, line 66: I would put the `volumeMounts:` inside the with to so it is not output at all if there are no mounts present, same goes for `volumes` below.

wnagele: done

stefansedich:
Thanks @wnagele I am also about to merged https://github.com/argoproj/argo-helm/pull/381 which changes a few things, it moves the `sso` section under from `controller.sso` to `server.sso` which makes more sense as that is what SSO is configuring, but apart from that we should be good to sso!

PR #396, opened 2020-07-02
fix: Use GKEBackendConfig when enabled

redbaron:
Checklist:

### DIFF
--- a/charts/argo-cd/templates/argocd-server/service.yaml
+++ b/charts/argo-cd/templates/argocd-server/service.yaml
@@ -1,10 +1,13 @@
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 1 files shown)


spencergilbert:
@redbaron I haven't used GKE's backendconfig before, looking at the docs they're setting some port information as well. Is that not necessary?

redbaron:
> @redbaron I haven't used GKE's backendconfig before, looking at the docs they're setting some port information as well. Is that not necessary?

It is possible to specify backend config for each port of the service. If they don't require different configurations, `default` works for all ports.

seanson:
@redbaron  - Is this PR ready to go? If so, could you move it from "WIP/Draft" status?

alexec:
Stale.

PR #402, opened 2020-07-10
feat(argo): Allow setting up ServiceAccount and RBAC resources for running workflows

lippertmarkus:
Before a workflow can be run, one would need to set up a service account as well as according roles and role bindings so that the argoexec sidecar container can get the logs and the status of the main container.

This is included in the [quickstart YAML](https://raw.githubusercontent.com/argoproj/argo/stable/manifests/quick-start-postgres.yaml) of Argo for the `default` service account (see `workflow-role` Role and RoleBinding) but not in this Helm chart.

For new users it would be helpful if the chart would provide an option to create those resources instead of letting the user manually create those.

This option is added by this PR.

Checklist:


lippertmarkus:
@benjaminws @stefansedich could you please look into this?

charts/argo/values.yaml
```diff
@@ -20,6 +20,15 @@ createAggregateRoles: true
 # and start argo-cli with the --namespaced flag. Use it in clusters with strict access policy.
 singleNamespace: false
 
+workflow:
+  namespace: ""  # Speficy namespace if workflows run in another namespace than argo. This controls where the service account and RBAC resources will be created.
+  serviceAccount:
+    create: false  # Specifies whether a service account should be created
+    annotations: {}
+    name: "default"  # Service account which is used to run workflows
```
stefansedich, line 28: IMO maybe something like `argo-workflow` might be a better default name?

charts/argo/templates/workflow-rb.yaml
```diff
@@ -0,0 +1,19 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ .Release.Name }}-workflow
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.workflow.serviceAccount.name }}
+  {{- if .Values.workflow.namespace }}
+  namespace: {{ .Values.workflow.namespace }}
+  {{- end }}
+{{- end }}
```
stefansedich, line 19: Could we get an \n at the end of this file.

charts/argo/templates/workflow-role.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - watch
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - pods/log
+  verbs:
+  - get
+  - watch
+{{- end }}
```
stefansedich, line 25: Could we get an \n at the end of this file.

stefansedich (changes requested):
Few small comments!

charts/argo/values.yaml
lippertmarkus: I chose `default` because it is used per default for running workflows when not specifying `serviceAccountName` inside the workflow definition. All the example workflows provided by argo are using the `default` service account and I think those workflows will be the first new user will try. If they would need to edit the example workflows provided by argo to include a service account like `argo-workflow` first before trying them out it won't be that easy for them to get started. The official quickstart YAML is also creation a RoleBinding for `default`.

What do you think?

stefansedich: Main issue here is if I try the following:

`helm upgrade --install argo charts/argo --set workflow.serviceAccount.create=true`

It results in an error:

`Unable to continue with install: ServiceAccount "default" in namespace "default" exists and cannot be imported into the current release`

So will break in that scenario, the name needs to be something other than default for that to work, IMO the examples using the default SA isn't giving the best advice and we should be pushing users to doing the right thing which is leaving the default SA alone and creating one specific to this purpose.

lippertmarkus: Thanks for the feedback. I changed it.

stefansedich (approved):
LGTM thanks!

Issue #404, opened 2020-07-14
Use a non-root redis image


FrediWeber:
**Is your feature request related to a problem? Please describe.**
The current redis image runs the container as root user.

**Describe the solution you'd like**
We should use a image that doesn't run as root.

**Describe alternatives you've considered**
Allow the container to run as root via a PSP.

**Additional context**
The "main" issue is: https://github.com/argoproj/argo-cd/issues/3933

ErikLundJensen:
This change breaks deployment at OpenShift where a specific user should not be used. As in the remaining of the Argo CD chart the securityContext must be empty set {} and it is the responsibility of the users for the chart to setup the securityContext as it is platform/environment specific.

spencergilbert:
> This change breaks deployment at OpenShift where a specific user should not be used. As in the remaining of the Argo CD chart the securityContext must be empty set {} and it is the responsibility of the users for the chart to setup the securityContext as it is platform/environment specific.

Can you open a new issue for this?

ErikLundJensen:
Opened new issue #441 as this fix does not work at OpenShift.

PR #405, opened 2020-07-14
fix: #404 - Change Redis image to run as non-root user

FrediWeber:
Checklist:

* [x ] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [x ] Any new values are backwards compatible and/or have sensible default.
* [x ] I have signed the CLA and the build is green.

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.6.1"
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 2.5.3
+version: 2.5.4
 home: https://github.com/argoproj/argo-helm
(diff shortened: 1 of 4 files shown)


danmx:
You should use the same image tag in both HA and non-HA versions.

https://github.com/argoproj/argo-helm/pull/405/files#diff-d8a32ce00c5f1ee2edbf400b509af243R314)

[`bitnami/redis`](https://hub.docker.com/r/bitnami/redis) doesn't support `5.0.8-alpine` tag

FrediWeber:
> You should use the same image tag in both HA and non-HA versions.
> 
> https://github.com/argoproj/argo-helm/pull/405/files#diff-d8a32ce00c5f1ee2edbf400b509af243R314)
> 
> [`bitnami/redis`](https://hub.docker.com/r/bitnami/redis) doesn't support `5.0.8-alpine` tag

Yes, that's why I left the HA unchanged. Why do we need the same image?

FrediWeber:
Okay I changed the approach. We can keep the image and just set the securityContext. The securityContext is already set in the HA variant.

PR #317, opened 2020-04-22
feat: Allow custom paths for server ingress

mhaley-miovision:
We use AWS EKS to host our cluster.  When we setup services using https we normally use ACM for the certificate, route53 for the domain and an ALB for handling the traffic to the service itself.  As part of this we like to put an http -> https redirect on port 80.  To do this we need to add another backend, as well as add an additional annotation on the load balancer.
```
alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
```

This PR just gives us the ability to add the additional listener rule on the ALB to perform that redirect.

Checklist:

* [x] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md).
* [x] I have signed the CLA and the build is green.
* [x] I will test my changes again once merged to master and published.

Changes are automatically published when merged to `master`. They are not published on branches.

PR #424, opened 2020-08-14
feat: Argo CD allow custom paths for ingress

mbarrien:
Similar to #317, this allows us to add extra paths to Argo CD.
This adds compatibility with the ALB Ingress Controller when using EKS
to be able to add the additional listener rule to perform a HTTP->HTTPS
redirect on port 80.

Unfortunately, we can't use the same structure as used in #317 because
there is an existing .Values.server.ingress.paths that we have to maintain
backwards compatibility with. The existing one is hard-coded to use a
fixed serviceName and servicePort, which doesn't work for the EKS ALB case.

Instead, we add .Values.server.ingress.extraPaths (and a similar var for
ingressGrpc) where a user directly specifies the entire rule, including
the path and backend information. Examples are added to the values.yaml
file for the extra path for EKS, with an empty list as a backwards
compatible default.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.6.2"
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 2.6.1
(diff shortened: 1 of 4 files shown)


mbarrien:
Any word?

mbarrien:
Done. Please re-review

PR #426, opened 2020-08-17
feat(argo-rollouts): allows adding args to argo-rollout controller

aweis89:
- Adds args to values as are often required to customize the behaviour of the contoller


cabrinha:
@aweis89 lets bump the chart version to 0.3.6

cabrinha:
Thanks for the PR!

Try to make your build green. Seems like you need to bump your version.

alexec:
@cabrinha you're approval is adequate for any PR.  @aweis89 please rebase.

yann-soubeyrand (changes requested):
There is a problem whith this PR: without digging into the details, it removes way too much things, like the Argo CD chart for example ;-)

PR #435, opened 2020-08-31
fix: Helm giving linting error when no ingress or ingressGrpc extraPaths are given

ekarademir:
Checklist:


cbanek:
This looks reasonable to me, I just got bit by the same bug.  Thanks for posting a PR.  But I'm not allowed to approve here, but if this could get fixed pretty soon, that'd be nice since it is broken for me since I use paths.

lukasmrtvy:
I can not enable ingress with these values.yaml:
```
  server:
    ingress:
        enabled: true
        hosts:
        - argocd.example.com
        tls: 
        - secretName: argocd.example.com-tls
          hosts:
          - argocd.example
        annotations:
          foo:bar
```
Am I affected by this ?
Thanks

ekarademir:
@lukasmrtvy Looks like it, yes.

PR #440, opened 2020-09-04
chore: bump to latest

VaibhavPage:
Checklist:

### DIFF
--- a/charts/argo-events/values.yaml
+++ b/charts/argo-events/values.yaml
@@ -46,21 +46,21 @@ singleNamespace: true
 sensorController:
   name: sensor-controller
   image: sensor-controller
-  tag: v0.17.0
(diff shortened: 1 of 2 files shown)


charts/argo-events/values.yaml
```diff
@@ -46,21 +46,21 @@ singleNamespace: true
 sensorController:
   name: sensor-controller
   image: sensor-controller
-  tag: v0.17.0
+  tag: latest
```
whynowy, line 49: I don't think we should use `latest` here, otherwise if we update the `latest` image tag, user's deployment will automatically pick up.

VaibhavPage: ok, makes sense

whynowy (approved):
LGTM

PR #460, opened 2020-09-30
feat(Argo): Make Argo Server service account creation conditional.

vladlosev:
In some situation, one needs to provide Argo Server with a different set of permissions. For example, to prohibit editing of workflow templates or prohibit creating new ones. This PR will allow users to use externally created service account which can be bound to a role with any desired set of permissions.

Checklist:

### DIFF
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -157,7 +157,7 @@ server:
   servicePort: 2746
   # servicePortName: http
   serviceAccount: argo-server
-  # Whether to create the service account wiht the name specified in
+  # Whether to create the service account with the name specified in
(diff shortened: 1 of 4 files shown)


charts/argo/values.yaml
```diff
@@ -157,6 +157,9 @@ server:
   servicePort: 2746
   # servicePortName: http
   serviceAccount: argo-server
+  # Whether to create the service account wiht the name specified in
```
stefansedich, line 160:
```suggestion
  # Whether to create the service account with the name specified in
```

stefansedich (approved):
LGTM

PR #465, opened 2020-10-04
feat: add extraContainers argument to deployment

simonfr:
Checklist:

### DIFF
--- a/charts/argo-cd/templates/argocd-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-server/deployment.yaml
@@ -120,6 +120,9 @@ spec:
           failureThreshold: {{ .Values.server.readinessProbe.failureThreshold }}
         resources:
 {{- toYaml .Values.server.resources | nindent 10 }}
+      {{- if .Values.server.extraContainers }}
+      {{- toYaml .Values.server.extraContainers | nindent 6 }}
+      {{- end }}
     {{- if .Values.server.nodeSelector }}
       nodeSelector:
 {{- toYaml .Values.server.nodeSelector | nindent 8 }}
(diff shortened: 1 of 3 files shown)


simonfr:
@spencergilbert @seanson Can you review my PR please ? :-)

charts/argo-cd/templates/argocd-server/deployment.yaml
```diff
@@ -120,6 +120,9 @@ spec:
           failureThreshold: {{ .Values.server.readinessProbe.failureThreshold }}
         resources:
 {{- toYaml .Values.server.resources | nindent 10 }}
+      {{- if .Values.server.extraContainers }}
```
seanson, line 123: Could we align the Helm template as per the indent above.

charts/argo-cd/values.yaml
```diff
@@ -653,6 +653,8 @@ server:
   #      oauthclientCredentials:
   #        secretName: argocd-secret
 
+  extraContainers: []
```
seanson, line 656: Could we get an example of an extra container in the comments below this entry here so the purpose is clear.

seanson (changes requested):
Hi @simonfr, thanks for the PR! A few requests and it should be good to go.

charts/argo-cd/templates/argocd-server/deployment.yaml
simonfr: It's done :-)

charts/argo-cd/values.yaml
simonfr: Added :-)

seanson (approved):
Awesome, thanks @simonfr!

PR #470, opened 2020-10-12
feat(argo): add PriorityClass support

dmegyesi:
Checklist:

---

This PR adds support to optionally use PriorityClasses in the Argo Workflows chart. In case the value is omitted, the pod will run with the cluster defaults.

Priority classes allow pods to avoid being evicted if there's resource contention in a cluster. In case of a race condition, the smaller priority pods will be kicked out first; similarly, when the kube-scheduler makes scheduling decisions, it will ensure first the higher priority pods get deployed.

This can be quite an important parameter for the Argo Workflows main components, as a [random user workflow](https://github.com/argoproj/argo/blob/346292b1b0152d5bfdc0387a8b2c11b5d6d5bac1/workflow/controller/workflowpod.go#L667-L672) is potentially able to evict the controller pods.

For more details you can read the [official docs](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/).

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v2.11.3
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.13.0
+version: 0.13.1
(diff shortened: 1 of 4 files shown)


stefansedich (approved):
LGTM thanks!

PR #474, opened 2020-10-19
feat(argo-rollouts): Add service monitor

alantang888:
Checklist:

### DIFF
--- a/charts/argo-rollouts/README.md
+++ b/charts/argo-rollouts/README.md
@@ -39,6 +39,7 @@ $ helm install --name my-release argo/argo-rollouts
 | controller.tolerations | [Tolerations for use with node taints](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) | `[]` |
 | controller.affinity | [Assign custom affinity rules to the deployment](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/) | `{}` |
 | controller.nodeSelector | [Node selector](https://kubernetes.io/docs/user-guide/node-selection/) | `{}` |
+| controller.metrics.serviceMonitor.enabled | bool | `false` |  |
 | imagePullSecrets | list | `[]` |  |
 | installCRDs | bool | `true` |  |
 | podAnnotations | object | `{}` |  |
--- a/charts/argo-rollouts/templates/argo-rollouts-service-monitor.yaml
+++ b/charts/argo-rollouts/templates/argo-rollouts-service-monitor.yaml
@@ -1,0 +1,28 @@
+{{- if .Values.controller.metrics.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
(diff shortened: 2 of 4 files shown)


charts/argo-rollouts/README.md
```diff
@@ -39,6 +39,7 @@ $ helm install --name my-release argo/argo-rollouts
 | controller.tolerations | [Tolerations for use with node taints](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) | `[]` |
 | controller.affinity | [Assign custom affinity rules to the deployment](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/) | `{}` |
 | controller.nodeSelector | [Node selector](https://kubernetes.io/docs/user-guide/node-selection/) | `{}` |
+| controller.enableMonitor | bool | `true` |  |
```
hobbsh, line 42: I'm not a maintainer, but I believe it makes more sense to use the convention followed by other charts, so this would be `controller.metrics.serviceMonitor.enabled`. Maybe it also makes sense to add the full functionality like other charts with a [service monitor](https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-server/servicemonitor.yaml) and full [metrics values block](https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml#L122-L134).

alantang888: Changed value name to follow argo-cd.

charts/argo-rollouts/values.yaml
```diff
@@ -24,6 +24,10 @@ controller:
   #    cpu: 50m
   #    memory: 64Mi
 
+  metrics:
+    serviceMonitor:
+      enabled: true
```
cabrinha, line 29: I would not enable this by default. Some users may not have the `monitoring.coreos.com/v1` CRD installed on their cluster.

charts/argo-rollouts/templates/argo-rollouts-service-monitor.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.controller.metrics.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ .Release.Name }}
+  labels:
```
cabrinha, line 6: I think we may want to allow users to add their own labels and annotations to this template.

cabrinha (changes requested):

charts/argo-rollouts/values.yaml
alantang888: Done

charts/argo-rollouts/templates/argo-rollouts-service-monitor.yaml
alantang888: Added both label and annotation. But I wonder is there really have use case for add annotation...

cabrinha (approved):

Issue #476, opened 2020-10-20
Expose argo workflows parallelism to helm chart


dbseel:
**Is your feature request related to a problem? Please describe.**
We would like to restrict the number of argo workflows running at one time, but this is currently not possible with the helm chart.

**Describe the solution you'd like**
The workflow controller config map supports the parallelism parameter, which sets the maximum number of workflows. Recommendation is to set this based on a value:

workflow-controller-config-map.yaml:
data:
  config: |
    parallelism: {{ .Values.controller.parallelism }}

**Describe alternatives you've considered**
We have tried using semaphores at the workflow/template level but have encountered other issues preventing us using this method (ie, create a bunch of workflows, delete them all, the lock was not released).

**Additional context**
I have currently forked the argo helm chart and implemented this recommendation, it is working well. However, I'd prefer not to maintain the fork.

PR #482, opened 2020-10-27
feat: Add parallelism to argo-helm chart

dbseel:
https://github.com/argoproj/argo-helm/issues/476

Checklist:

### DIFF
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -37,7 +37,7 @@ controller:
     # Overrides .images.tag if defined.
     tag: ""
   # parallelism dictates how many workflows can be running at the same time
-  parallelism: 10
+  parallelism:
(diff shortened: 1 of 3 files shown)


charts/argo/values.yaml
```diff
@@ -36,6 +36,8 @@ controller:
   image:
     # Overrides .images.tag if defined.
     tag: ""
+  # parallelism dictates how many workflows can be running at the same time
+  parallelism: 10
```
stefansedich, line 40: We should not provide a restrictive default here, can we make the default null and then conditionally set the `paralellism` option in the config-map.

stefansedich (changes requested):
One comment on defaulting.

charts/argo/values.yaml
dbseel: Done

stefansedich (approved):
LGTM thanks!

PR #486, opened 2020-11-03
feat: add variable "server.readinessProbe.Scheme" to argo chart

rtang03:
Checklist:

* [ X ] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [ X ] Any new values are backwards compatible and/or have sensible default.
* [ X ] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md).
* [ X ] I have signed the CLA and the build is green.
* [ X ] I will test my changes again once merged to master and published.

### DIFF
--- a/charts/argo/templates/server-deployment.yaml
+++ b/charts/argo/templates/server-deployment.yaml
@@ -52,7 +52,7 @@ spec:
             httpGet:
               path: /
               port: 2746
-              scheme: HTTP
+              scheme: {{ .Values.server.readinessProbeScheme }}
             initialDelaySeconds: 10
             periodSeconds: 20
           env:
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -157,6 +157,7 @@ server:
   # SecurityContext to set on the server pods
   podSecurityContext: {}
   name: server
+  # either HTTP or HTTPS
   readinessProbeScheme: HTTP
   serviceType: ClusterIP
   servicePort: 2746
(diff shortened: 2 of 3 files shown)


charts/argo/templates/server-deployment.yaml
```diff
@@ -52,7 +52,7 @@ spec:
             httpGet:
               path: /
               port: 2746
-              scheme: HTTP
+              scheme: {{ .Values.server.readinessProbeScheme }}
```
stefansedich, line 55: Could we change this to `readinessProbe.scheme`.

charts/argo/values.yaml
```diff
@@ -157,6 +157,8 @@ server:
   # SecurityContext to set on the server pods
   podSecurityContext: {}
   name: server
+  # either HTTP or HTTPS
+  readinessProbeScheme: HTTP
```
stefansedich, line 161: As above can we change this `readinessProbe.scheme`.

stefansedich (changes requested):
Can we change to use `readinessProbe.scheme` that appears more common with charts for controlling the readiness probe.

charts/argo/values.yaml
rtang03: Sure.

charts/argo/templates/server-deployment.yaml
rtang03: Sure. Should I submit another PR, to make above change?

josephmcgovern-wf:
The core issue here may have already been solved by #506

rtang03:
> The core issue here may have already been solved by #506

That is exactly what I want. And, this PR no longer requires.

Issue #455, opened 2020-09-23
Error installing minio dependency for Argo chart


cgorman:
**Describe the bug**
Trying to install Argo using helm chart 0.12.1 produces the following error:

```
Error: template: argo/charts/minio/templates/deployment.yaml:198:20: executing "argo/charts/minio/templates/deployment.yaml" at <(not .Values.gcsgateway.enabled) (not .Values.azuregateway.enabled) (not .Values.s3gateway.enabled) (not .Values.b2gateway.enabled)>: can't give argument to non-function not .Values.gcsgateway.enabled
```

While looking for [other bug reports](https://github.com/helm/charts/issues/21446), I noticed that they have [moved the Minio charts](https://hub.helm.sh/charts/minio/minio) to helm.min.io. Has that been reflected here? Either way it looks like it was fixed in chart version 5.0.16 so could we get the requirements updated, please?

**To Reproduce**
```
$ helm version
version.BuildInfo{Version:"v3.3.1", GitCommit:"249e5215cde0c3fa72e27eb7a30e8d55c9696144", GitTreeState:"clean", GoVersion:"go1.14.7"}
```
1. Install Argo helm chart with minio.install=true

**Expected behavior**
Argo and minio install correctly.

SniperBean:
I got same issue with helm v2.16.12 when installing Argo Workflow helm chart with minio.install=true

thenetengineer:
Same issue.

tsunamishaun:
I noticed the requirements is pointed at the old repo and version of the minio chart. It has moved to [here](https://github.com/minio/charts/tree/master/minio) but there also appears to be a bitnami chart that is slightly more up to date. I leave that choice up to the maintainers :)

jaawasth:
any updates here ?

tsunamishaun:
@jaawasth Posted on argo [slack](https://argoproj.slack.com/archives/C8J6SGN12/p1604435599312400), hopefully this will get some 👀

jaawasth:
@seanson , fyi, this is the issue which references the minio chart update that needs to be done with argo-helm

jaawasth:
@jessesuen @benjaminws , can you please help here, helm2 will end support on 12th Nov, 2020

jiaranda:
Please merge this PR :eyes:

naveenb30:
I'm also having same issue with `minio.install=true`

mszostok:
[...]

PR #495, opened 2020-11-12
fix(argo): Update minio chart to solve install error

mszostok:
Checklist:

Fixes https://github.com/argoproj/argo-helm/issues/455

---

The `8.0.3` version fails on `ct` with:
```
[ERROR] templates/: render error in "argo/charts/minio/templates/deployment.yaml": template: argo/charts/minio/templates/_helpers.tpl:49:43: executing "minio.deployment.apiVersion" at <.Capabilities.KubeVe...>: can't evaluate field Version in type *version.Info

Error: 1 chart(s) linted, 1 chart(s) failed
Error linting charts
------------------------------------------------------------------------------------------------------------------------
 ✖︎ /workdir/charts/argo > Error waiting for process: exit status 1
```

To solve that the repo should be updated to helm3 and use the `quay.io/repository/helmpack/chart-testing` image in version v3.3.0 or higher, so for now the latest minio that can be used is the `v8.0.0`.

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v2.11.3
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.13.2
+version: 0.13.3
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -8,6 +8,7 @@ rm -rf $SRCROOT/output && git clone -b gh-pages git@github.com:argoproj/argo-hel
 
 helm repo add stable https://kubernetes-charts.storage.googleapis.com
 helm repo add argoproj https://argoproj.github.io/argo-helm
+helm repo add minio https://helm.min.io/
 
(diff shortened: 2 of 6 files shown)


scripts/publish.sh
```diff
@@ -8,6 +8,7 @@ rm -rf $SRCROOT/output && git clone -b gh-pages git@github.com:argoproj/argo-hel
 
 helm repo add stable https://kubernetes-charts.storage.googleapis.com
```
stefansedich, line 9: We can probably remove this right?

charts/argo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v2.11.3
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.13.2
+version: 0.13.3
```
stefansedich, line 5: Can you bump to the next chart version, think it is 0.14.1 at this point.

stefansedich (changes requested):
Few small changes and this should be good to get in! apologies for the delay missed this one.

cabrinha:
I could be wrong but, now that [apiVersion](https://github.com/argoproj/argo-helm/blob/master/charts/argo/Chart.yaml#L1) is `v2`: the requirements.yaml and requirements.lock files need to be deleted. The "dependencies" list needs to be moved into Chart.yaml

mszostok:
I tested the `0.16.7` Argo Helm Chart version and the MinIO bug is fixed, so I'm closing this PR.

PR #498, opened 2020-11-13
fix(argo): Map the /tmp directory into server pods.

vladlosev:
https://github.com/argoproj/argo/pull/4350 started using the `/tmp` directory for storing temporary files created for the purposes of downloading artifacts. But the standard image the project built is created from scratch and does not contain `/tmp`. The project's own manifests were updated to map the directory using an `emptyDir` volume, but this chart was not updated.  That started causing failures for users of this chart.

This change adds the `/tmp` mapping into the chart, thus fixing the artifact downloading.

Checklist:

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v2.11.3
+appVersion: v2.11.7
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.13.2
+version: 0.13.3
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
--- a/charts/argo/templates/server-deployment.yaml
+++ b/charts/argo/templates/server-deployment.yaml
@@ -67,16 +67,20 @@ spec:
             value: {{ .Values.server.baseHref | quote }}
           resources:
             {{- toYaml .Values.server.resources | nindent 12 }}
-          {{- with .Values.server.volumeMounts }}
           volumeMounts:
+          - name: tmp
+            mountPath: /tmp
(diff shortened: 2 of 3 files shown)


charts/argo/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: v2.11.3
+appVersion: v2.11.7
```
stefansedich, line 2: If we are bumping the appVersion, I think we should bump `image.tag` in values.yaml too.

charts/argo/templates/server-deployment.yaml
```diff
@@ -67,16 +67,20 @@ spec:
             value: {{ .Values.server.baseHref | quote }}
           resources:
             {{- toYaml .Values.server.resources | nindent 12 }}
-          {{- with .Values.server.volumeMounts }}
           volumeMounts:
+          - name: tmp
```
stefansedich, line 71: I assume this is always `/tmp` and the temp path cannot change? i.e. is it controllable in Argo?

stefansedich (changes requested):

charts/argo/Chart.yaml
vladlosev: Done.

charts/argo/templates/server-deployment.yaml
vladlosev: No, it's [hard-coded](https://github.com/argoproj/argo/pull/4350/files#r511150735).

stefansedich (approved):
LGTM thanks! :shipit:

willemm:
You have used the wrong indent level for these additions.  They should math the 'nindent' level used in the following bit where custom volume mounts are added.  This breaks for anyone with custom volumemounts.

vladlosev:
Ah, good point, my bad. Fix is in https://github.com/argoproj/argo-helm/pull/500.

PR #499, opened 2020-11-13
feat(Argo): Add secret access whitelist for server.

vladlosev:
Currently, the chart allows the Argo server to access all the secrets in the namespace. It's not good security practice, as an attacker taking over a server pod will have access to all of them. This change adds support for restricting the secrets the server can read to a given list so that only need-to-know secrets can be allowed.

Checklist:


stefansedich (approved):
LGTM

Issue #505, opened 2020-11-23
Support overwriting argo-server readiness probe scheme in values.yaml


josephmcgovern-wf:
**Is your feature request related to a problem? Please describe.**

When enabling [TLS](https://argoproj.github.io/argo/tls/#encrypted) in an argo testing environment using the `--secure` flag, the readiness probe fails because it only makes http requests (instead of https). The argo `values.yaml` file should support overwriting the readiness probe get scheme.

Looking at the [readiness probe definition](https://github.com/argoproj/argo-helm/blob/c99716bc2036efc74a46bdd76da5a9630ccd786b/charts/argo/templates/server-deployment.yaml#L51-L57), it's clear to see that those values cannot be overwritten out of the box by helm values.

Instead of checking out this entire file, overwriting it, and making sure it stays in sync with the latest updates to this repository, it would be great if I could just provide a value to the chart deployment that allows me to tweak this scheme.

**Describe the solution you'd like**

I would like to be able to specify in my values yaml file a new http scheme. Example:
```
server:
  enabled: true
  extraArgs: ["--secure"]
  readinessProbe:
    httpGet:
        scheme: HTTPS
```

**Describe alternatives you've considered**
I could overwrite this entire file but then I would have to keep it in sync with every release (which I don't want to do).

**Additional context**
N/A

PR #506, opened 2020-11-24
fix(argo): Use https for readiness probe in secure mode

josephmcgovern-wf:
Checklist:

This fixes the issue raised in https://github.com/argoproj/argo-helm/issues/505. It's worth noting this implementation is different than the suggestion in the original issue. Instead of requiring users to update both the `extraArgs` and readiness probe http scheme to enable tls, we can automatically change the http scheme if tls is enabled.

Tested this locally by:
* Deploying the chart with no extra args
* Submitting workflow defined in the app notes
* Update the values.yaml to have `server.secure: true` and re-deploy the chart
* Submitting workflow defined in the app notes

My test results:
✅ confirmed that the chart was deployed successfully with `server.secure` set to false
✅ confirmed that the workflow completed successfully with `server.secure` set to false
✅ confirmed that the chart was deployed successfully with the `server.secure` set to true
✅ confirmed that the workflow completed successfully with the `server.secure` set to true
✅ confirmed that the pods readiness probe used the https scheme when deployed with the secure flag
✅ confirmed that the pods readiness probe used the http scheme when deployed without any flags
✅ confirmed that without these changes, deploying with only the`--secure` flag in extra args result in a failed deployment because the readiness probe continually fails.

### DIFF
--- a/charts/argo/templates/server-deployment.yaml
+++ b/charts/argo/templates/server-deployment.yaml
@@ -38,6 +38,9 @@ spec:
           {{- if .Values.server.extraArgs }}
           {{- toYaml .Values.server.extraArgs | nindent 10 }}
           {{- end }}
+          {{- if .Values.server.secure }}
+          - "--secure"
+          {{- end }}
           {{- if .Values.singleNamespace }}
           - "--namespaced"
           {{- end }}
@@ -52,7 +55,11 @@ spec:
             httpGet:
               path: /
               port: 2746
+              {{- if .Values.server.secure }}
+              scheme: HTTPS
+              {{- else }}
               scheme: HTTP
+              {{- end }}
             initialDelaySeconds: 10
             periodSeconds: 20
           env:
--- a/charts/argo/values.yaml
+++ b/charts/argo/values.yaml
@@ -201,6 +201,12 @@ server:
   # PriorityClass: system-cluster-critical
   priorityClassName: ""
 
+  # Run the argo server in "secure" mode. Configure this value instead of
+  # "--secure" in extraArgs. See the following documentation for more details
+  # on secure mode:
+  # https://argoproj.github.io/argo/tls/#encrypted
+  secure: false
+
   # Extra arguments to provide to the Argo server binary.
   extraArgs: []
 
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: v2.11.7
 description: A Helm chart for Argo Workflows
 name: argo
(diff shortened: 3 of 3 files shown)


charts/argo/templates/server-deployment.yaml
```diff
@@ -52,7 +52,11 @@ spec:
             httpGet:
               path: /
               port: 2746
+              {{- if has "--secure" .Values.server.extraArgs }}
```
stefansedich, line 55: I wonder if instead we want a value on the chart for turning on secure mode, i.e. `server.secure=true`, this can then add the `--secure` flag and change the scheme of the probe, doing magic things based on `extraArgs` never sits right with me.

josephmcgovern-wf: I would be fine with that

josephmcgovern-wf:
@stefansedich updated with your suggestion, and walked through the (modified) testing steps. I was able to confirm once again that this fixes the readiness probe bug.

charts/argo/values.yaml
```diff
@@ -201,6 +201,12 @@ server:
   # PriorityClass: system-cluster-critical
   priorityClassName: ""
 
+  # Run the argo server in "secure" mode. If this value is set to true, do not
+  # specify "secure" in extraArgs. See the following documentation for more
+  # details on secure mode:
+  # https://argoproj.github.io/argo/tls/#encrypted
+  secure: true
```
stefansedich, line 208: Should we default to `false` here so that we are not secure by default? and change the comment above to say `If this value is set to false, do not # specify "secure" in extraArgs`

josephmcgovern-wf: Yes. I must have accidentally committed this after testing. It should be false - good catch.

josephmcgovern-wf: I think we don't want to set "--secure" anymore at all in extraArgs because of this value. I will update the comment to say that this replaces the need for setting `--secure` in extraArgs.

stefansedich (approved):
LGTM, thanks! This should make it much easier to run the server in secured mode.

josephmcgovern-wf:
@stefansedich anything else we need to merge?

stefansedich:
> @stefansedich anything else we need to merge?

For some reason the builds never reported status, can you maybe bump the version one more time and we can see if it kicks off CI and runs, I don't have access to do that.

josephmcgovern-wf:
@stefansedich I just bumped the version again, any luck?

stefansedich (approved):

stefansedich:
> @stefansedich I just bumped the version again, any luck?

Looks like we have a winner! Thanks.

PR #504, opened 2020-11-23
feat: Make K8s API version configurable for CRD

zdtsw:
Checklist:

* [ ] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [ ] Any new values are backwards compatible and/or have sensible default.
* [ ] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md).
* [ ] I have signed the CLA and the build is green.
* [ ] I will test my changes again once merged to master and published.

Changes are automatically published when merged to `master`. They are not published on branches.

PR #514, opened 2020-12-08
fix: upgraded CRD version to apiextensions.k8s.io/v1

maateen:
I am using Kubernetes `v1.18.9` for one of my clusters. In this version, the kind "apiextensions.k8s.io/v1beta1 CustomResourceDefinition" is deprecated in favor of "apiextensions.k8s.io/v1 CustomResourceDefinition". I get the following error while using Helm3:

```shell
➜ helm lint .
==> Linting .
[WARNING] templates/cluster-workflow-template-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart
[ERROR] templates/cluster-workflow-template-crd.yaml: the kind "apiextensions.k8s.io/v1beta1 CustomResourceDefinition" is deprecated in favor of "apiextensions.k8s.io/v1 CustomResourceDefinition"
[WARNING] templates/cron-workflow-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart
[ERROR] templates/cron-workflow-crd.yaml: the kind "apiextensions.k8s.io/v1beta1 CustomResourceDefinition" is deprecated in favor of "apiextensions.k8s.io/v1 CustomResourceDefinition"
[WARNING] templates/workflow-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart
[ERROR] templates/workflow-crd.yaml: the kind "apiextensions.k8s.io/v1beta1 CustomResourceDefinition" is deprecated in favor of "apiextensions.k8s.io/v1 CustomResourceDefinition"
[WARNING] templates/workflow-template-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart
[ERROR] templates/workflow-template-crd.yaml: the kind "apiextensions.k8s.io/v1beta1 CustomResourceDefinition" is deprecated in favor of "apiextensions.k8s.io/v1 CustomResourceDefinition"

Error: 1 chart(s) linted, 1 chart(s) failed
```

After this fix:

```shell
➜ helm lint .
==> Linting .
[WARNING] templates/cluster-workflow-template-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart
[WARNING] templates/cron-workflow-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart
[WARNING] templates/workflow-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart
[WARNING] templates/workflow-template-crd.yaml: manifest is a crd-install hook. This hook is no longer supported in v3 and all CRDs should also exist the crds/ directory at the top level of the chart

1 chart(s) linted, 0 chart(s) failed
```

Checklist:

### DIFF
--- a/charts/argo-cd/crds/crd-application.yaml
+++ b/charts/argo-cd/crds/crd-application.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   labels:
--- a/charts/argo-cd/crds/crd-project.yaml
+++ b/charts/argo-cd/crds/crd-project.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   labels:
--- a/charts/argo-events/crds/eventbus-crd.yml
+++ b/charts/argo-events/crds/eventbus-crd.yml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: eventbus.argoproj.io
--- a/charts/argo-events/crds/eventsource-crd.yml
+++ b/charts/argo-events/crds/eventsource-crd.yml
@@ -1,5 +1,5 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: eventsources.argoproj.io
--- a/charts/argo-events/crds/sensor-crd.yml
+++ b/charts/argo-events/crds/sensor-crd.yml
@@ -1,5 +1,5 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: sensors.argoproj.io
--- a/charts/argo-events/templates/eventbus-crd.yaml
+++ b/charts/argo-events/templates/eventbus-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: eventbus.argoproj.io
--- a/charts/argo-events/templates/eventsource-crd.yaml
+++ b/charts/argo-events/templates/eventsource-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: eventsources.argoproj.io
--- a/charts/argo-events/values.yaml
+++ b/charts/argo-events/values.yaml
@@ -21,7 +21,7 @@ additionalSaNamespaces: []
 additionalServiceAccountRules:
 - apiGroups:
     - apiextensions.k8s.io
-    - apiextensions.k8s.io/v1beta1
+    - apiextensions.k8s.io/v1
   verbs:
     - create
     - delete
--- a/charts/argo-rollouts/templates/crds/analysis-template-crd.yaml
+++ b/charts/argo-rollouts/templates/crds/analysis-template-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
--- a/charts/argo-rollouts/templates/crds/cluster-analysis-template-crd.yaml
+++ b/charts/argo-rollouts/templates/crds/cluster-analysis-template-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
--- a/charts/argo-rollouts/templates/crds/experiment-crd.yaml
+++ b/charts/argo-rollouts/templates/crds/experiment-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
--- a/charts/argo-rollouts/templates/crds/rollout-crd.yaml
+++ b/charts/argo-rollouts/templates/crds/rollout-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
--- a/charts/argo/crds/cluster-workflow-template-crd.yaml
+++ b/charts/argo/crds/cluster-workflow-template-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: clusterworkflowtemplates.argoproj.io
--- a/charts/argo/crds/cron-workflow-crd.yaml
+++ b/charts/argo/crds/cron-workflow-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: cronworkflows.argoproj.io
--- a/charts/argo/crds/workflow-crd.yaml
+++ b/charts/argo/crds/workflow-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: workflows.argoproj.io
--- a/charts/argo/crds/workflow-eventbinding-crd.yaml
+++ b/charts/argo/crds/workflow-eventbinding-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: workfloweventbindings.argoproj.io
--- a/charts/argo/crds/workflow-template-crd.yaml
+++ b/charts/argo/crds/workflow-template-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: workflowtemplates.argoproj.io
--- a/charts/argo/templates/cluster-workflow-template-crd.yaml
+++ b/charts/argo/templates/cluster-workflow-template-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: clusterworkflowtemplates.argoproj.io
--- a/charts/argo/templates/cron-workflow-crd.yaml
+++ b/charts/argo/templates/cron-workflow-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: cronworkflows.argoproj.io
--- a/charts/argo/templates/workflow-crd.yaml
+++ b/charts/argo/templates/workflow-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: workflows.argoproj.io
--- a/charts/argo/templates/workflow-template-crd.yaml
+++ b/charts/argo/templates/workflow-template-crd.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: workflowtemplates.argoproj.io
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.7.6
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 2.10.0
+version: 2.11.0
 home: https://github.com/argoproj/argo-helm
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 keywords:
--- a/charts/argo-events/Chart.yaml
+++ b/charts/argo-events/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to install Argo-Events in k8s Cluster
 name: argo-events
-version: 1.0.0
+version: 1.1.0
 keywords:
   - argo-events
   - sensor-controller
--- a/charts/argo-events/templates/sensor-crd.yaml
+++ b/charts/argo-events/templates/sensor-crd.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.installCRD }}
 # Define a "sensor" custom resource definition
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: sensors.argoproj.io
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.9.1"
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 0.3.10
+version: 0.4.0
 icon: https://raw.githubusercontent.com/argoproj/argo/master/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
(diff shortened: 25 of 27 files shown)


spencergilbert:
It looks like #504 also tackles this issue, I think the question is if it needs to remain configurable for users on older versions

charts/argo-cd/crds/crd-application.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-cd/crds/crd-project.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-events/crds/eventbus-crd.yml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-events/crds/eventsource-crd.yml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-events/crds/sensor-crd.yml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-events/templates/eventbus-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-events/templates/eventsource-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-events/values.yaml
```diff
@@ -21,7 +21,7 @@ additionalSaNamespaces: []
 additionalServiceAccountRules:
 - apiGroups:
     - apiextensions.k8s.io
-    - apiextensions.k8s.io/v1beta1
+    - apiextensions.k8s.io/v1
```
james-d-elliott, line 24:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-rollouts/templates/crds/analysis-template-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-rollouts/templates/crds/cluster-analysis-template-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-rollouts/templates/crds/experiment-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo-rollouts/templates/crds/rollout-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRDs }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/crds/cluster-workflow-template-crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/crds/cron-workflow-crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/crds/workflow-crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/crds/workflow-eventbinding-crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/crds/workflow-template-crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 1:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/templates/cluster-workflow-template-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/templates/cron-workflow-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/templates/workflow-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

charts/argo/templates/workflow-template-crd.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.installCRD }}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
james-d-elliott, line 2:
```suggestion
{{- if semverCompare "<1.18-0" .Capabilities.KubeVersion.GitVersion }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- else }}
apiVersion: apiextensions.k8s.io/v1
{{- end }}
```

james-d-elliott (commented):
I think this may be a better way to handle this instead of making the user configure it or hard coding it. Thoughts? I'm relatively new to charts but this is how I handled it in one of mine. Just not sure if there are implications I'm not considering.

For your consideration too @stefansedich the other way to handle it may be a helper which checks for a used defined override, or automatically applies it based on k8s version from capabilities. So if you for what ever reason need to target a specific version and override it you have that option.

zdtsw:
The main reason why I did a fully confinable version but still set default to v1beta1 is considering backwards compatibility. I think it is the same thoughts as @james-d-elliott 's suggestion.

I am not a big fan of hardcoded version, on the other hand, design pattern in this repo (or most of the helm charts) likes 
 to follow hardcoded value for apiVersion in all templates. 

As @spencergilbert mentioned, the question is do we need to support old version?  
if not, what @maateen did is good enough;  
if yes, we need to decide detail implementation: use a help function or just leave it to user to set value.

I will have my MR [feat: Make K8s API version configurable for CRD ](https://github.com/argoproj/argo-helm/pull/514) close first and let's keep this one active for further discussion.

seanson (changes requested):
Given that API Capabilities are now stable since 1.5 I think it would be good to use the flag `.Capabilities.APIVersions.Has " apiextensions.k8s.io/v1"` instead of a Kubernetes version check as that seems like it would be a better experience for users in earlier clusters that support the version.

Additionally it looks like upstream is still using `apiextensions.k8s.io/v1beta1` and there's an open issue to address it in https://github.com/argoproj/argo-cd/issues/4280

What's the behaviour when a cluster removes the old CRD and applies the new version? Will deleting and recreating the CRD require recreation of App objects or do we require an additional PATCH action against the API to migrate the existing App objects?

james-d-elliott:
> 
> 
> Given that API Capabilities are now stable since 1.5 I think it would be good to use the flag `.Capabilities.APIVersions.Has " apiextensions.k8s.io/v1"` instead of a Kubernetes version check as that seems like it would be a better experience for users in earlier clusters that support the version.
> 
> Additionally it looks like upstream is still using `apiextensions.k8s.io/v1beta1` and there's an open issue to address it in [argoproj/argo-cd#4280](https://github.com/argoproj/argo-cd/issues/4280)
> 
> What's the behaviour when a cluster removes the old CRD and applies the new version? Will deleting and recreating the CRD require recreation of App objects or do we require an additional PATCH action against the API to migrate the existing App objects?

So I've not had specific experience with this change in ArgoCD, however I manually modified my Traefik CRD's to use v1 instead of v1beta1 (they didn't define the OpenAPI 3 schema at all). After making the changes to the CRD nothing else had to be updated. So I would expect as long as the CRD versions are not getting bumped by this change (which currently they are not), then it should be fine, but should probably be tested either way.

One thing of note that is important - CRD's with .spec.Version are not valid in v1 I don't think, so I can't help but wonder if these manifests have even been tested against an apiserver that supports v1 by either @maateen or @zdtsw and on which apiserver versions?

zdtsw:
> > Given that API Capabilities are now stable since 1.5 I think it would be good to use the flag `.Capabilities.APIVersions.Has " apiextensions.k8s.io/v1"` instead of a Kubernetes version check as that seems like it would be a better experience for users in earlier clusters that support the version.
> > Additionally it looks like upstream is still using `apiextensions.k8s.io/v1beta1` and there's an open issue to address it in [argoproj/argo-cd#4280](https://github.com/argoproj/argo-cd/issues/4280)
> > What's the behaviour when a cluster removes the old CRD and applies the new version? Will deleting and recreating the CRD require recreation of App objects or do we require an additional PATCH action against the API to migrate the existing App objects?
> 
> So I've not had specific experience with this change in ArgoCD, however I manually modified my Traefik CRD's to use v1 instead of v1beta1 (they didn't define the OpenAPI 3 schema at all). After making the changes to the CRD nothing else had to be updated. So I would expect as long as the CRD versions are not getting bumped by this change (which currently they are not), then it should be fine, but should probably be tested either way.
> 
> One thing of note that is important - CRD's with .spec.version are not valid in v1 I don't think, so I can't help but wonder if these manifests have even been tested against an apiserver that supports v1 by either @maateen or @zdtsw and on which apiserver versions?

Just looked at my values and I actually had "installCRDs: false" so the answer to your question is "no". I did not even use CRD in my setups. Only recall that I saw the warning after running helm and initialized PR

james-d-elliott:
```
error: error validating "crd-application.yaml": error validating data: [
ValidationError(CustomResourceDefinition.spec): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, 
ValidationError(CustomResourceDefinition.spec): unknown field "subresources" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, 
ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, 
ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec
]; 

if you choose to ignore these errors, turn validation off with --validate=false
```

```
error: error validating "crd-project.yaml": error validating data: [
ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, 
ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec
]; 
if you choose to ignore these errors, turn validation off with --validate=false
```

maateen:
> ```
> error: error validating "crd-application.yaml": error validating data: [
> ValidationError(CustomResourceDefinition.spec): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, 
> ValidationError(CustomResourceDefinition.spec): unknown field "subresources" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, 
> ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, 
> ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec
> ]; 
> 
> if you choose to ignore these errors, turn validation off with --validate=false
> ```

@james-d-elliott This is because of some parameter changes in the CRD API.

alexec:
I don't think this should be fixed in the Helm charts. It should be fixed in the original repos.

james-d-elliott:
> 
> 
> I don't think this should be fixed in the Helm charts. It should be fixed in the original repos.

Yeah according to [this post](https://github.com/traefik/traefik/issues/5473#issuecomment-763454171) it's going to be done with code generation. I do have a gist with working manifests either way for reference: https://gist.github.com/james-d-elliott/c3323ab66df3f09150b7538a8a8a8aca

charts/argo-cd/crds/crd-application.yaml
vladlosev: These files are in the `crd` directory, not under `templates`. Are you sure the template directives in them will be processed?

cabrinha:
Looks like this is fixed in https://github.com/argoproj/argo-helm/pull/576

PR #517, opened 2020-12-10
feat(argo): Add support for setting env vars for workflow controller.

vladlosev:
This PR adds support for setting custom environment variables on the workflow controller.

Checklist:

### DIFF
--- a/charts/argo/templates/workflow-controller-deployment.yaml
+++ b/charts/argo/templates/workflow-controller-deployment.yaml
@@ -61,6 +61,9 @@ spec:
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.namespace
+          {{- with .Values.controller.env }}
+            {{ toYaml . | nindent 10 }}
+          {{- end }}
           resources:
             {{- toYaml .Values.controller.resources | nindent 12 }}
           {{- if .Values.controller.metricsConfig.enabled }}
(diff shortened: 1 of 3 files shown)


charts/argo/templates/workflow-controller-deployment.yaml
```diff
@@ -61,6 +61,9 @@ spec:
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.namespace
+          {{- with .Values.controller.env }}
```
stefansedich, line 64: Normally `extraEnv` is used for additional env vars in a container.

charts/argo/values.yaml
```diff
@@ -115,6 +115,7 @@ controller:
   # service type `LoadBalancer`
   loadBalancerSourceRanges: []
   resources: {}
+  env: []
```
stefansedich, line 118: Might pay to put an example of what extraEnv could look like.

stefansedich (changes requested):
If we could just change `env` -> `extraEnv` and add an example to the values.

charts/argo/templates/workflow-controller-deployment.yaml
vladlosev: Done.

charts/argo/values.yaml
vladlosev: These values serve as defaults so they are not an appropriate place to put examples. I've added a description in comment instead.

stefansedich (approved):
LGTM

PR #529, opened 2020-12-30
feat(argo): Add argo SSO/RBAC support server cluster role

krrrr38:
## motivation

argo workflow 2.12.0 support SSO + RBAC, and it requires sa list + secret create. ref https://github.com/argoproj/argo/pull/4514.

----

Checklist:

### DIFF
--- a/charts/argo/templates/server-cluster-roles.yaml
+++ b/charts/argo/templates/server-cluster-roles.yaml
@@ -30,6 +30,7 @@ rules:
   - serviceaccounts
   verbs:
   - get
+  - list
 - apiGroups:
   - ""
   resources:
@@ -45,6 +46,22 @@ rules:
   - events
   verbs:
   - watch
+{{- if .Values.server.sso }}
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - sso
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+{{- end }}
 {{- if .Values.controller.persistence }}
 - apiGroups:
   - ""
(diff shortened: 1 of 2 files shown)


charts/argo/templates/server-cluster-roles.yaml
```diff
@@ -45,6 +46,17 @@ rules:
...
+{{- if .Values.server.sso }}
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - sso
+  verbs:
+  - get
+  - create
```
stefansedich, line 58: The create permission cannot be attached to a specific resource name like is done, here will need to allow create on all for create permissions to support.

stefansedich (changes requested):
Will need some changes to the role for things to work.

charts/argo/templates/server-cluster-roles.yaml
krrrr38: fixup it allows to create all secrets

stefansedich: Thanks looks like CI is failing however, I wonder if it is because the old helm repos have finally been taken down?

krrrr38: I tried to fix it, but cannot find a good solution now...
- https://stackoverflow.com/a/65408163
- https://github.com/helm/chart-testing/issues/161

```sh
> docker run -v `pwd`:/workdir --rm gcr.io/kubernetes-charts-ci/test-image:v3.1.0 ct lint --debug --config .circleci/chart-testing.yaml --lint-conf .circleci/lintconf.yaml --helm-repo-extra-args ‘--stable-repo-url=https://charts.helm.sh/stable’
```

krrrr38: I guess that it's time to change ci char-testing image from `gcr.io/kubernetes-charts-ci/test-image:v3.1.0` to `quay.io/helmpack/chart-testing:v3.1.0`. it means we need to upgrade ci helm version from 2.x to 3.x.

krrrr38: fix it in https://github.com/argoproj/argo-helm/pull/537

krrrr38:
https://github.com/argoproj/argo-helm/pull/541 support this feature, close this

PR #563, opened 2021-01-28
chore: Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/argo-helm/commit/299fe87c58ad9546213d999ff4059579e55d5cc5#commitcomment-46477627

The action reports that the changes in this PR would make it happy: https://github.com/argoproj/argo-helm/commit/f3179dd57732fb4cac60ef10ccfdb6efea1f44d9

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

---

Checklist:

---

I'll update the chart versions once the content changes are deemed acceptable (and probably squash). I leave commits as distinct while reviewing because otherwise things get really painful.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
       # We just need an image with `helm` on it. Handily we know of one already.
       - image: quay.io/helmpack/chart-testing:v3.3.1
     steps:
-      # install the additional keys needed to push to Github. Alex Collins owns these keys.
+      # install the additional keys needed to push to GitHub. Alex Collins owns these keys.
       - add_ssh_keys
       - run: git config --global user.email "nobody@circleci.com"
       - run: git config --global user.name "Circle CI Build"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,7 +76,7 @@ Please ensure chart version changes adhere to semantic versioning standards:
 
 ## Testing Charts
 
-As part of the Continuous Intergration system we run Helm's [Chart Testing](https://github.com/helm/chart-testing) tool.
+As part of the Continuous Integration system we run Helm's [Chart Testing](https://github.com/helm/chart-testing) tool.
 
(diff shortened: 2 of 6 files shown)


charts/argo-cd/crds/crd-project.yaml
```diff
@@ -164,7 +164,7 @@ spec:
                       description: Name is a name for this role
                       type: string
                     policies:
-                      description: Policies Stores a list of casbin formated strings that define access policies for the role in the project
```
spencergilbert, line 167: I don't think we should edit this one here, since it's pulled in directly from the upstream CRDs

spencergilbert (changes requested):

charts/argo-cd/crds/crd-project.yaml
yann-soubeyrand: I agree, CRDs should be fixed upstream.

jsoref: Sure, got a handy link? otherwise i'll figure it out.

yann-soubeyrand: I guess the files to modify are here (I don't think they are otherwise generated): https://github.com/argoproj/argo-cd/tree/master/manifests/crds

jsoref: https://github.com/argoproj/argo-cd/pull/5373

yann-soubeyrand (approved):

spencergilbert (approved):
Thanks! Have you looked at the failing CI, I think just fixing the semantic part is OK by me - doesn't seem like any of this _Should_ fail the other CI

jsoref:
The underlying CRD is horribly broken and really should be outside the scope of this change.

I'm going to try to fix it in https://github.com/argoproj/argo-events/pull/1048 --  but really, it's a huge amount of work and it's really unfortunate.

The easiest thing for me to do is to drop the argo-events changes for the time being because that chart effectively isn't buildable (if you tried to build master today, it'd fail for the same reasons).

PR #567, opened 2021-02-02
feat(argocd-notifications): update to support argocd-notification v1.0.1

georgekaz:
Checklist:

* [x] Any new values ~~are backwards compatible and/or~~ have sensible default.


alexmt (approved):
Thanks a lot, @georgekaz .

Reviewed and tested locally - chart perfectly works. LGTM 

Have only one suggestion by lets discuss it first and work on it in next PR:

Instead of commenting out default templates and triggers, I would add a boolean flag e.g. `includeCatalog: false`. If it is set to `true` than config map should include templates and triggers from catalog.

Issue #568, opened 2021-02-09
Allow using custom configmap for rbac


johnkost:
**Is your feature request related to a problem? Please describe.**
We wrap the ArgoCD chart with a custom chart to deploy resources we require alongside ArgoCD. When working on adding RBAC, the only way we can template is RBAC is with Ansible and passing that in the values file because ArgoCD does not allow us to reference our own `argocd-rbac-cm`.

**Describe the solution you'd like**
I would like to be able to have my wrapper chart create the required RBAC configmap and have the ArgoCD chart use that configmap

johnkost:
https://github.com/argoproj/argo-cd/blob/4417cc81804f85e33e60774b574821f0ac930eef/common/common.go#L21-L23

I found that the configmap names are hardcoded into ArgoCD. We could still allow supplying the CM (PR I am opening does this) bu the name would have to be constant

johnkost:
Closed by https://github.com/argoproj/argo-helm/pull/569

PR #569, opened 2021-02-09
feat(argo-cd): Conditionally create the RBAC configmap

johnkost:
Solves https://github.com/argoproj/argo-helm/issues/568. If needed, I can do this for other configmaps that are used as well

Checklist:

### DIFF
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -595,13 +595,9 @@ server:
   ## Annotations to be added to ArgoCD rbac ConfigMap
   rbacConfigAnnotations: {}
 
-  # Name of the configmap used for the RBAC config. Must be set to `argocd-rbac-cm` as the values are hardcoded in
-  # ArgoCD. See https://github.com/argoproj/argo-cd/blob/master/common/common.go
-  rbacConfigMapName: "argocd-rbac-cm"
-
   # Boolean determining whether or not to create the configmap. If false, it is expected tthe configmap will be created
   # by something else. ArgoCD will not work if there is no configMap created with the name above.
-  rbacConfigMapEnabled: true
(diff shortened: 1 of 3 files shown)


charts/argo-cd/values.yaml
```diff
@@ -595,6 +595,14 @@ server:
   ## Annotations to be added to ArgoCD rbac ConfigMap
   rbacConfigAnnotations: {}
 
+  # Name of the configmap used for the RBAC config. Must be set to `argocd-rbac-cm` as the values are hardcoded in
+  # ArgoCD. See https://github.com/argoproj/argo-cd/blob/master/common/common.go
+  rbacConfigMapName: "argocd-rbac-cm"
```
yann-soubeyrand, line 600: I’m not sure we should offer the possibility to configure the ConfigMap name if it’s hardcoded on Argo CD side. It’s error prone in my opinion.

charts/argo-cd/values.yaml
```diff
@@ -595,6 +595,14 @@ server:
...
+
+  # Boolean determining whether or not to create the configmap. If false, it is expected tthe configmap will be created
+  # by something else. ArgoCD will not work if there is no configMap created with the name above.
+  rbacConfigMapEnabled: true
```
yann-soubeyrand, line 604: I would name this value rbacConfigCreate to be coherent with the controller.serviceAccount.create value (we cannot have rbacConfig.create since rbacConfig is already used). What do you think?

charts/argo-cd/values.yaml
johnkost: That makes sense. Will remove

yann-soubeyrand (approved):

PR #545, opened 2021-01-12
fix(argo-cd): Fixes CRD schema.

vladlosev:
This PR makes the schema attributes compatible with `apiextensions.k8s.io/v1` after #541 bumped the version without this adjustment.

Best viewed with whitespace diff [suppressed](https://github.com/argoproj/argo-helm/pull/545/files?w=1).

Signed-off-by: Vlad Losev <vladimir.losev@sage.com>

Checklist:

* [x] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md).
* [x] I have signed the CLA and the build is green.
* [x] I will test my changes again once merged to master and published.

Changes are automatically published when merged to `master`. They are not published on branches.

PR #574, opened 2021-02-09
feat(argo-cd): updated ArgoCD to v1.8.4

chgl:
Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 1.7.11
+appVersion: 1.8.4
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 2.13.0
+version: 2.14.0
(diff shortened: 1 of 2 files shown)


davidkarlsen (approved):

yann-soubeyrand:
Hello @chgl, the CRDs seem to not be in sync with upstream CRDs for the 1.8.4 version. I’m surprised with the amount of differences. Could you verify that the CRDs in this chart are the ones from v1.7.x and update them to the v1.8.4?

chgl:
@yann-soubeyrand  sorry about that. It looks like https://github.com/argoproj/argo-helm/pull/545 changed the CRD API version to v1, however the CRD manifests over at https://github.com/argoproj/argo-cd/blob/v1.8.4/manifests/crds/application-crd.yaml still use v1beta1. This should explain the significant diff.

Not entirely sure what the best way forward is. I'd prefer keeping the chart CRDs in-sync with the argo-cd crds (https://github.com/argoproj/argo-cd/blob/v1.8.4/manifests/crds/). But this would mean downgrading the API version which might break what https://github.com/argoproj/argo-helm/pull/545 tried to achieve.

PR #575, opened 2021-02-10
feat: Allow custom labels in Server Ingress

bakhtin:
Signed-off-by: Artyom Bakhtin <a@bakhtin.net>

We use `cert-manager` to manage SSL certificates inside the EKS cluster. Depending on the value of `labels` in Ingress object, `cert-manager` selects a matching method to solve HTTP01/DNS01 challenge (https://cert-manager.io/docs/installation/upgrading/upgrading-0.7-0.8/#performing-an-incremental-switch-to-the-new-format).

Adding support for custom `labels` in Server Ingress object to allow it to work with `cert-manager`. Or just in case one needs to add custom labels for some other reason.

Checklist:

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.12.5
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.15.5
+version: 0.16.0
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.12.5
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.15.4
+version: 0.15.5
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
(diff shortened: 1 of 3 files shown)


charts/argo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.12.5
 description: A Helm chart for Argo Workflows
 name: argo
-version: 0.15.4
+version: 0.15.5
```
yann-soubeyrand, line 5: You’re adding a “feature” (new value) to the chart, the new version should be 0.16.0 ;-)

charts/argo/values.yaml
```diff
@@ -226,6 +226,11 @@ server:
     #   kubernetes.io/ingress.class: nginx
     #   kubernetes.io/tls-acme: "true"
 
+    ## Labels to be added to the web ingress.
+    ##
+    # labels:
```
yann-soubeyrand, line 231: Should we name it additionalLabels to clearly indicate that it won’t replace the default labels? What do you think @stefansedich @vladlosev?

vladlosev: In this chart, we already have multiple instances of the `podLabels` and `serviceLabels` attributes specifying _additional_ labels on different object. Same goes for additional annotation attributes. So leaving this attribute as just `labels` will be consistent with the existing convention.

charts/argo/Chart.yaml
bakhtin: Bumped to 0.16.0

yann-soubeyrand (approved):

PR #530, opened 2020-12-30
chore(argo-cd): create roles instead of rolebindings when singleNamespace is true

searsaw:
Checklist:

* [x] I have update the chart version in `Chart.yaml` following Semantic Versioning.
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have followed the testing instructions in the [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md).
* [x] I have signed the CLA and the build is green.
* [x] I will test my changes again once merged to master and published.

Changes are automatically published when merged to `master`. They are not published on branches.

PR #593, opened 2021-02-24
fix(argo): Restore cluster template role binding to be ClusterRoleBinding.

vladlosev:
#530 changed the `cluster-template` `ClusterRoleBinding`s into namespace-scoped `RoleBinding`s when `singleNamespace` is set in values. That change is incorrect as the `cluster-template` role grants permissions on cluster-scoped objects, and binding it with a namespace scoped `RoleBinding` will _not_ grant the subjects the expected permissions. This binding has to be `ClusterRoleBinding`.

Checklist:

### DIFF
--- a/charts/argo/Chart.yaml
+++ b/charts/argo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.12.5
 description: A Helm chart for Argo Workflows
 name: argo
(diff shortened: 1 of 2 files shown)


stefansedich (approved):
LGTM this makes sense, the role/binding for access to cluster workflowtemplates will need to be at the cluster level.

PR #594, opened 2021-02-24
fix(argo): Restore server cluster-template to be a ClusterRoleBinding.

vladlosev:
This is a followup on #593, as the server role binding needs to be cluster scope as well. Thanks to @stefansedich for the catch.

Signed-off-by: Vlad Losev <vladimir.losev@sage.com>

Checklist:

### DIFF
--- a/charts/argo/templates/server-cluster-roles.yaml
+++ b/charts/argo/templates/server-cluster-roles.yaml
@@ -111,11 +111,7 @@ rules:
   - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-{{- if .Values.singleNamespace }}
-kind: Role
-{{- else }}
 kind: ClusterRole
-{{- end }}
 metadata:
   name: {{ .Release.Name }}-{{ .Values.server.name }}-cluster-template
 rules:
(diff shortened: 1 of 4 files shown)


charts/argo/templates/server-cluster-roles.yaml
```diff
@@ -111,11 +111,7 @@ rules:
   - delete
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-{{- if .Values.singleNamespace }}
-kind: Role
-{{- else }}
 kind: ClusterRole
```
stefansedich, line 114: Will also need to fix the bindings in https://github.com/argoproj/argo-helm/blob/556011dd4d764eb8c11a8d41eb9526e67e6c186e/charts/argo/templates/server-crb.yaml#L23

stefansedich (changes requested):

charts/argo/templates/server-cluster-roles.yaml
vladlosev: Done.

stefansedich (approved):
LGTM

PR #596, opened 2021-02-25
add crds dir to .helmignore

kostas-theo:
The CRDs in the `./crds` directory were still being created even when the `installCRD` value was set to false. This is now fixed in the `argo` and `argo-events` charts.

I also found another bug while I testing for this change, which was a missing `apiVersion` line in the `argo/templates/workflow-aggregate-roles.yaml` template.

Checklist:

### DIFF
--- a/charts/argo-events/.helmignore
+++ b/charts/argo-events/.helmignore
@@ -19,3 +19,5 @@
 .project
 .idea/
 *.tmproj
+# Custom Resource Definitions for manual creation only
+crds/
--- a/charts/argo-events/Chart.yaml
+++ b/charts/argo-events/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to install Argo-Events in k8s Cluster
 name: argo-events
-version: 1.2.3
+version: 1.2.4
 keywords:
   - argo-events
   - sensor-controller
--- a/charts/argo/.helmignore
+++ b/charts/argo/.helmignore
@@ -19,3 +19,5 @@
 .project
 .idea/
 *.tmproj
+# Custom Resource Definitions for manual creation only
+crds/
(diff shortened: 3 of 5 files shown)


vladlosev:
Ah, good catch about the API version.

With regards to the CRDs, Helm by default installs (but not upgrades) everything in the CRD directory. Please note that Helm has its own `--skip-crds` option which you can utilize:
>      --skip-crds                    if set, no CRDs will be installed. By default, CRDs are installed if not already present

IMHO, it would be against the user's expectations to hide the `crds` directory from Helm. A better approach would be to update the CRDs with the fresh ones from the Argo project itself and then synchronize the ones Argo installs via templates with the ones in the `crds` dir. For example, see how Istio  is [doing that](https://github.com/istio/istio/blob/master/manifests/charts/base/templates/crds.yaml).

vladlosev:
I have created #598 to quickly fix the chart.

kostas-theo:
Thanks for the quick reply @vladlosev . So I had a go at trying to fix something but maybe I should take a step back. First off, the main reason why I was trying to fix this was a persistent error trying to create the CRD, which I have now created a seperate issue for: https://github.com/argoproj/argo-helm/issues/599

Secondly, as a relatively new helm user I'm not really sure on the distinction you make between installing and upgrading everything in the `crds` directory. However, I think that users would expect that if `installCRD: false` was set, no crds would be installed.

vladlosev:
@kostas-theo unfortunately, due to historic reasons, you now have two knobs to turn with regards to the CRD installation. You have to specify both `--skip-crds` and `--set=installCRD=false` if you want to reliably prevent the CRD installation by the chart. If you are interested the background, https://banzaicloud.com/blog/helm3-the-good-the-bad-and-the-ugly/#managing-crds and https://helm.sh/docs/chart_best_practices/custom_resource_definitions/ might give you some details.

kostas-theo:
@vladlosev - that sounds like a mess :(

Thanks for taking the time to explain. I've raised https://github.com/argoproj/argo-helm/pull/605/files to get me out of my underlying issue anyway.

Cheers

PR #614, opened 2021-03-08
chore!: move all charts to apiVersion v2

### LINKED ISSUE #611: [all]: Move Helm ApiVersion v2
Do you any timeline, when you would want to change to Helm's apiVersion v2 or why this has not happened yet. Since Helm apiVersion 1 was deprecated with the Release of Helm v3.0.0 ([Helm 3.0.0 Blog](https://helm.sh/blog/helm-3-released/)) it seems like a good idea to me to move to the new Helm apiVersion. I would be happy to do that contribution for you guys, just wanted to check your thoughts on this.

This would include changing each chart's dependency declaration and said `apiVersion: v2`. If Users have installed ArgoCD with Helm 2.X they would simply need to migrate with the official helm Plugin. More information on that:

  * [https://helm.sh/docs/topics/v2_v3_migration/](https://helm.sh/docs/topics/v2_v3_migration/)

Don't migrate to apiVersion 2, But i don't see a reason why that should be.

**Additional context**

oliverbaehler:
This PR introduces no new changes to the charts. But since the changes breaks compatibility with helm v2, i have bumped to a new major version for all the affected charts (seemed like to obvious way for me). If that's not correct in your opinion let me know. I was also thinking about only incrementing to a new minor version.

fixes #611 

Checklist:


yann-soubeyrand (approved):
Maybe we should create a legacy branch so that people can open pull requests against it to backport fixes? Let’s see what other maintainers think before merging ;-)

cabrinha:
I like this but it appears you have some merge conflicts.

mkilchhofer:
I think we should also update the README.md (especially for the argo-cd chart):

- [ ] Add Upgrading notes
- [ ] Mention helm3 in "Prerequisites" section

Maybe it also makes sense to split the PR into one PR per chart and refresh/refactor some parts of the charts (eg. formatting, code dedup with template functions) while introducing a "breaking change" anyways.

oliverbaehler:
Thanks for your input @mkilchhofer. That with the documentation agree- What should we note in the "Upgrading Notes". I guess the simplest way of this is just using the helm 2to3 plugin. I am just thinking maybe we should do refactoring in a dedicated step?  

@yann-soubeyrand I don't know if we kinda create like two worlds with having two branches but at the same time you are right, there should be some way to be able to backport fixes. What about if we announce that from a certain point we just start supporting v2 (as google helm charts did). Becausethe majority of Helm v3.X is far proofen by this point and it should be avoided having apiVersion v1 for much longer in my opinion.

oliverbaehler:
Thinking about it, this change should be bundled with other changes to a new Major as @mkilchhofer mentioned. Maybe it would make more sense to have an `edge` branch where we release rcs for new majors?

oliverbaehler:
Charts are deprecated or have moved to new apiVersion, therefor resolving this. fixes #611

PR #619, opened 2021-03-09
fix(argo-cd): Mount emptyDir on repoServer when using read-only FS

mkilchhofer:
I wanted to set `containerSecurityContext.readOnlyRootFilesystem: true` in all ArgoCD components. The component `repoServer` then fails to start with this error message:
```log
level=fatal msg="could not create canary: open /app/config/gpg/keys/.argocd-generated: read-only file system"
```
And if I fix the problem for `/app/config/gpg/keys` with an emptyDir volume, it then fails because /tmp is read-only:
```log
level=fatal msg="open /tmp/gpg-key-recipe505421993: read-only file system"
```

So I mount emptyDirs for both two directories when running with a read-only filesystem.

Checklist:


mkilchhofer:
The checklist item regarding "signed the CLA" is done when I read through [contributing guide](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md) and performed the manual tests?

davidkarlsen:
Does it need to be conditional? Having an emptyDir for /tmp makes sense anyway I think. Also if we could run with readOnlyRootfs by default that makes sense too IMHO.

mkilchhofer:
Indeed we can omit the conditional. In the app repo over there, the manifests adds adding the gpg directory by default:
- https://github.com/argoproj/argo-cd/pull/4136/files
- https://github.com/argoproj/argo-cd/blob/master/manifests/base/repo-server/argocd-repo-server-deployment.yaml#L51-L52

Should I proceed directly or wait for other feedbacks?

davidkarlsen:
I think you can proceed changing this PR, and we gather feedback on it, but using `emptyDir` for such temporary data which follows the pod lifecycle makes sense - it's what it was designed for.

mkilchhofer:
> I think you can proceed changing this PR, and we gather feedback on it, but using `emptyDir` for such temporary data which follows the pod lifecycle makes sense - it's what it was designed for.

Ok, I will remove the conditionals here.
What about changing the default regarding read-only root fs?
I found an issue to implement this also in the main repo over there: https://github.com/argoproj/argo-cd/issues/3648

mkilchhofer:
> I like the idea. Could you try to resume the discussion upstream? Is it something you’ve already tried in production?

I bumped the discussion "upstream". Let's see what happen. When my own cluster for my personal workload counts as prod, yes ;-). Logs are smooth for 24h or so. Maybe we can merge this PR without changing the default for now and create another issue oder draft PR?

yann-soubeyrand:
Sounds good to me. Since @davidkarlsen started this review, I’ll let him proceed with the merge if everything is OK for him ;-)

davidkarlsen (approved):

PR #626, opened 2021-03-17
feat(argo-cd): add the ability to specify automountServiceAccountToken for argo-cd

joshuastern:
Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.8.4
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 2.17.0
+version: 2.17.1
 home: https://github.com/argoproj/argo-helm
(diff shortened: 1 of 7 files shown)


davidkarlsen:
What is the motivation for this change?

joshuastern:
> What is the motivation for this change?

As part of following best practices, we want to be able to explicitly configure this field rather than get it for free. When the `automountServiceAccountToken` is not set, a service account token is still mounted.

davidkarlsen:
> > What is the motivation for this change?
> 
> As part of following best practices, we want to be able to explicitly configure this field rather than get it for free. When the `automountServiceAccountToken` is not set, a service account token is still mounted.

sgtm. can you get the semantic pr check to pass so we can merge it?

joshuastern:
> > > What is the motivation for this change?
> > 
> > 
> > As part of following best practices, we want to be able to explicitly configure this field rather than get it for free. When the `automountServiceAccountToken` is not set, a service account token is still mounted.
> 
> sgtm. can you get the semantic pr check to pass so we can merge it?

done

davidkarlsen (approved):

PR #637, opened 2021-03-28
feat(argo-cd): Add compatibility to v1 Ingress

### LINKED ISSUE #535: server ingress apiversion warning 1.19+
when using the included ingress on k8s 1.19+, the following warning is emitted:
`networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress` 

if `networking.k8s.io/v1` api is available, ingress should be generated with correct api. should also add the `ingressClassName` field to the object. 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#ingressspec-v1-networking-k8s-io

mkilchhofer:
This resolves #535 

Checklist:

* [X] I have signed the DCO and the build is green.


charts/argo-cd/templates/argocd-server/ingress-grpc.yaml
```diff
@@ -19,41 +15,66 @@ metadata:
...
-{{- if .Values.server.ingressGrpc.labels }}
-{{- toYaml .Values.server.ingressGrpc.labels | nindent 4 }}
-{{- end }}
+    {{- if .Values.server.ingressGrpc.labels }}
+    {{- toYaml .Values.server.ingressGrpc.labels | nindent 4 }}
+    {{- end }}
 spec:
+  {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+  {{- with .Values.server.ingress.ingressClassName }}
+  ingressClassName: {{ . }}
```
oliverbaehler, line 24:
```suggestion
  {{- if .Values.server.ingress.ingressClassName }}
  ingressClassName: {{ .Values.server.ingress.ingressClassName  }}
```
I had problem when rewriting ingresses that it does not regocnize the field yet. This way we can ensure it only changes if set. Otherwise you have a conflict on the resource bcs you would also try to set it via annotations

charts/argo-cd/templates/argocd-server/ingress-grpc.yaml
```diff
@@ -19,41 +15,66 @@ metadata:
...
           - path: {{ $p }}
+            {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+            pathType: Prefix
+            {{- end }}
             backend:
+              {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+              service:
+                name: {{ $serviceName }}
+                port:
+                  number: {{ $servicePort }}
```
oliverbaehler, line 46:
```suggestion
                {{- if or (kindIs "int" $servicePort) (kindIs "float64" $servicePort }}
                  number: {{ $servicePort }}
                {{- else }}
                  name: {{ $servicePort }}
                {{- end }}         
```
Note that we lose compatibility with this change, we need to evaluate if the port is a number or a port reference.

charts/argo-cd/templates/argocd-server/ingress.yaml
```diff
@@ -19,41 +15,66 @@ metadata:
...
           - path: {{ $p }}
+            {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+            pathType: Prefix
+            {{- end }}
             backend:
+              {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+              service:
+                name: {{ $serviceName }}
+                port:
+                  number: {{ $servicePort }}
```
oliverbaehler, line 69:
```suggestion
                {{- if or (kindIs "int" $servicePort) (kindIs "float64" $servicePort }}
                  number: {{ $servicePort }}
                {{- else }}
                  name: {{ $servicePort }}
                {{- end }}         
```
Same as above

charts/argo-cd/templates/argocd-server/ingress.yaml
```diff
@@ -19,41 +15,66 @@ metadata:
...
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.server.name "name" .Values.server.name) | nindent 4 }}
-{{- if .Values.server.ingress.labels }}
-{{- toYaml .Values.server.ingress.labels | nindent 4 }}
-{{- end }}
+    {{- if .Values.server.ingress.labels }}
+    {{- toYaml .Values.server.ingress.labels | nindent 4 }}
+    {{- end }}
 spec:
+  {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+  {{- with .Values.server.ingress.ingressClassName }}
```
oliverbaehler, line 23:
```suggestion
  {{- if .Values.server.ingress.ingressClassName }}
  ingressClassName: {{ .Values.server.ingress.ingressClassName  }}
```
Same as above

charts/argo-cd/templates/argocd-server/ingress.yaml
```diff
@@ -19,41 +15,66 @@ metadata:
...
           - path: {{ $p }}
+            {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+            pathType: Prefix
+            {{- end }}
             backend:
+              {{- if eq (include "argo-cd.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+              service:
+                name: {{ $serviceName }}
+                port:
+                  number: {{ $servicePort }}
```
oliverbaehler, line 46:
```suggestion
                {{- if or (kindIs "int" $servicePort) (kindIs "float64" $servicePort }}
                  number: {{ $servicePort }}
                {{- else }}
                  name: {{ $servicePort }}
                {{- end }}         
```
Same as above

oliverbaehler (changes requested):

charts/argo-cd/templates/argocd-server/ingress-grpc.yaml
mkilchhofer: I agree but, do we also need to check for `float64`? What's the usecase?

charts/argo-cd/templates/argocd-server/ingress-grpc.yaml
mkilchhofer: Hmm and this is solved by using an `if` statement? The `with` block is not rendered by default (when the value is `nil` or empty string).

charts/argo-cd/templates/argocd-server/ingress-grpc.yaml
oliverbaehler: Should be the question, why we check for `int`? By default these values are float64. And I thought we should also catch int.

DirtyCajunRice: @oliverbaehler There is never a case that you want to test for int when pulling from values.yaml. Example:
testchart/values.yaml
```yaml
tests:
  - 123
  - 123.123
  - "123"
```
templates/test.yaml
```
{{- range $i, $v := .Values.tests -}}
  {{- printf "# test %d KindOf: %s\n" $i (kindOf $v) -}}
{{- end -}}
```
output:
```
# Source: testchart/templates/test.yaml
# test 0 KindOf: float64
# test 1 KindOf: float64
# test 2 KindOf: string

charts/argo-cd/templates/argocd-server/ingress-grpc.yaml
DirtyCajunRice: @mkilchhofer  is correct. With vs If makes no change here.

charts/argo-cd/templates/argocd-server/ingress-grpc.yaml
oliverbaehler: @DirtyCajunRice Thanks for your input. I also tested it and couldn't really find a scenario where it becomes int. I implemented it that way in one of my libraries, it could be that it changes after multiple from/toyaml conversion, idk tbh :D. Anyway I adjusted my suggestions.

Do you have any creative idea how we could catch if a number is between quotes? As you showed in your example in quotes it will always be string and I am sure some people still have their number in quotes and that would be again breaking. I will also try to come up with something.

DirtyCajunRice: @oliverbaehler Yes you could always have templates that try to force conversions but IMO this is not the purpose of helm honestly. We version tag and notate breaking changes to ensure that proper usage, derived from informed updates, yield valid results. Realistically there is a bare minimum expectation of knowledge when someone is deploying applications via a templating interface (helm) derived from a templating language (go template/html + sprig) to an orchestration stack (k8s) over containerization (docker/containerd/etc).

oliverbaehler (approved):
lgtm

PR #645, opened 2021-03-29
fix(argo-cd): Fix breaking dex change in argocd 2.0

jessebye:
Checklist:

These changes should make the helm chart compatible with Argo CD version 2.0.0, per https://argo-cd.readthedocs.io/en/latest/operator-manual/upgrading/1.8-2.0/#dex-tool-migrated-from-argocd-util-to-argocd-dex


ArgTang:
Could this (major chart bump) be a chance to use helm chartverion 2? ref #611 
Argocd v2 will use helmv3 as default, so it might make sense to drop it in the chart as well

jessebye:
@ArgTang definitely! I bumped it to v2 in the Chart.yaml.

aweis89:
LGTM

charts/argo-cd/Chart.yaml
```diff
@@ -1,8 +1,8 @@
-apiVersion: v1
-appVersion: 1.8.4
+apiVersion: v2
+appVersion: 2.0.0
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
```
oliverbaehler, line 4: Thanks for taking this opportunity to change to apiVersion v2. Please note, that we also need to change the dependency model for the redis-ha chart.

I need you to delete thw following two files:
  *  requirements.lock
  * requirements.yaml

In the `Chart.yaml` we need to add the `dependencies` section (replaces the requirement files):

```
dependencies:
  - name: redis-ha
    version: 4.10.1
    repository: https://dandydeveloper.github.io/charts/
    condition: redis-ha.enabled
```
You can also see this changes in my pr - https://github.com/argoproj/argo-helm/pull/614

After that update the dependencies (`helm dependency update`) and remove the charts folder. (We need to update the Chart.lock).

Finally, we need to add a new section in the `Upgrading` heading. Something like this: 

``` 
3.0.0 and above
Helm apiVersion switched to `v2`. Requires Helm `3.0.0` or above to install. [Read More](https://helm.sh/blog/migrate-from-helm-v2-to-helm-v3/) on how to migrate your release from Helm 2 to Helm 3.
```
Since this is the effective change. I don't think your other changes require manual interaction, therefor we don't need to mention them.

If you have done these changes we can approve this pull request. If you need help, let me know. Thanks again.

oliverbaehler (changes requested):
Changes according to comment

jessebye:
@oliverbaehler updated to helm 3 as requested 😁

oliverbaehler (approved):
Changes are ready to merge. I am wondering if we should wait until we have a stable 2.0.0 docker build. I will let that decision to the person merging this pr.

aweis89:
> Changes are ready to merge. I am wondering if we should wait until we have a stable 2.0.0 docker build. I will let that decision to the person merging this pr.

IMO, it should be up to the end-user to decide. Merging this PR doesn't force anyone to upgrade, it just provides everyone the option of doing so. Also the rc2 has been pretty rock solid so far!

jessebye:
Is there anything blocking this from a merge? What is the process to get this shipped?

oliverbaehler:
@jessebye sadly my powers here don't allow me to merge. We will have to wait until one of the owners comes by and merges this. That's why most of the pull requests don't have any activity, sadly. I am sorry this takes a lot of time. I have already offered the owners to be more active here but without merge power I can just approve..

davidkarlsen:
> Is there anything blocking this from a merge? What is the process to get this shipped?

Personally I think we should await a final release, but that's just me. Nothing in the chart signals it's a pre-release and generally a user would expect latest and greatest chart to be a stable one, no?

charts/argo-cd/values.yaml
```diff
@@ -10,7 +10,7 @@ installCRDs: true
 global:
   image:
     repository: argoproj/argocd
-    tag: v1.8.4
+    tag: v2.0.0-rc2
```
mstrzele, line 13: `v2.0.0` is there! 🎉

Cian911:
This is great, is there anything blocking this from being merged?

davidkarlsen:
> This is great, is there anything blocking this from being merged now that `v2.0.0` is released?

bump the value to v2.0.0 now?

charts/argo-cd/values.yaml
davidkarlsen, line 13:
```suggestion
    tag: v2.0.0
```

jessebye: Done!

davidkarlsen (approved):

marcuz:
Just want to confirm it works like a charm. Thank you! :raised_hands:

mhoyer:
If I compare the changes done in the `kustomize` based [`install.yaml`](https://github.com/argoproj/argo-cd/blob/v2.0.0/manifests/install.yaml) between v1.8.7 and v2.0.0 of the [argoproj/argo-cd](https://github.com/argoproj/argo-cd/tree/v2.0.0) repository, I still see some changes not showing up here in this PR.

* the descriptions of the CRDs are different - why not just straight forward copying the [original ones from the argo-cd repo](https://github.com/argoproj/argo-cd/tree/v2.0.0/manifests/crds) here into the helm repo? 
* quay.io was used, but missing in the helm chart
  * see for instance: https://github.com/argoproj/argo-cd/blob/v2.0.0/manifests/install.yaml#L2547 vs. https://github.com/argoproj/argo-helm/blob/1d96f80dac56a6305bd263c6370ba3c38dadbaca/charts/argo-cd/values.yaml#L12
* redis was not updated in the `values.yaml`
  * see for instance: https://github.com/argoproj/argo-cd/blob/v2.0.0/manifests/install.yaml#L2596 vs. https://github.com/argoproj/argo-helm/blob/1d96f80dac56a6305bd263c6370ba3c38dadbaca/charts/argo-cd/values.yaml#L279
* in the `install.yaml` I also found an optional `argocd-repo-server-tls` secret mounted to `application-controller`, `repo-server` and `argocd-server`. This is also missing in the helm chart. Shall we add it too?

Find the diff of the `install.yaml` between ArgoCD 1.8.7 and 2.0.0 here: https://github.com/argoproj/argo-cd/compare/v1.8.7..v2.0.0#diff-085adbc8bea5279ae0da31de54e660d34708509979e708a566d8e5f5339822d8

oliverbaehler:
@mhoyer Thanks for comparing the changes. Would you be open to create a new PR addressing those issues?

jessebye:
@mhoyer I wasn't aware of the kustomize manifests or I would have synced up with them in this PR, good catch!

mhoyer:
I think over the time there were even other things that diverged over time. I wonder if there is a chance to somehow use the `helm template` feature to generate the full stack of k8s resources in a single `.yaml` file and diff it against the generated `install.yaml` in original argo-cd repo? But I assume sorting will make this a tricky thing. 🤷

PR #649, opened 2021-03-31
feat(argo): Add optional minio-operator and raw chart as dependencies

### LINKED ISSUE #639: [argo] Move to minio operator
Per the [minio deprecation notice](https://helm.min.io/) it is recommended to move to the [minio operator ](https://github.com/minio/operator).

Modify argo chart to have a conditional dependency on minio-operator and also optionally install a minio cr to match the current usage.

Not upgrading and continue to use deprecated minio chart.

**Additional context**
After the recent argo chart reference fix to the minio chart [#455](https://github.com/argoproj/argo-helm/issues/455) we realized it would be deprecated next week.

oliverbaehler:
This change fixes #639. It adds the possibility to deploy the minio-operator chart as dependency of argo-cd. Since for my experience, it's not really that common to have an operator as subchart of an application chart (at least in production environments) i have added the raw chart as well. The raw chart allows to add loose kubernetes manifests which will then also be maintained by the argo-cd helm release. This way it's possible to just stage a new minio tenant with the minio-operator being maintained by a different release.

These changes make the raw chart and minio-operator just optional. The old minio chart is still functional (Update the Bugfix version) and therefor this does not cause a breaking change. This way there's the possibility for users to add really any manifests they need additionaly for their argo release.

That's just my take on the issue, i think the solution is nice, but i am open for any input.

Note on DCO:

Checklist:


mkilchhofer:
Hi @oliverbaehler
Chart argo != argo-cd :-). I'd rephrase the description and the title

oliverbaehler:
@mkilchhofer Thanks for noticing. Now everything is fixed and pipelines are ok too :). Ready to be reviewed.

charts/argo/values.yaml
```diff
@@ -322,9 +322,207 @@ artifactRepository:
...
+      ## Refer https://github.com/minio/console
+      console:
+        image: minio/console:v0.6.3
+        replicas: 2
+        consoleSecret:
+          name: console-secret
+        securityContext:
+          runAsUser: 1000
+          runAsGroup: 2000
+          runAsNonRoot: true
```
mkilchhofer, line 528: What I do not like here is that `resources` is a list. When we need to customize _(and we do need this)_ it from external _(eighter `--set` or via `--values`)_, we need to define the whole block again since lists cannot be merged...

charts/argo/values.yaml
```diff
@@ -322,9 +322,207 @@ artifactRepository:
...
+      name: minio-creds-secret
+    type: Opaque
+    data:
+      ## Access Key for MinIO Tenant, base64 encoded (echo -n 'minio' | base64)
+      accesskey: bWluaW8=
+      ## Secret Key for MinIO Tenant, base64 encoded (echo -n 'minio123' | base64)
+      secretkey: bWluaW8xMjM=
+  ## Secret to be used for MinIO Console
+  - apiVersion: v1
+    kind: Secret
```
mkilchhofer, line 361: What I do not like here is that `resources` is a list. When we need to customize _(and we do need this)_ it from external _(eighter `--set` or via `--values`)_, we need to define the whole block again since lists cannot be merged...

oliverbaehler: That assumption is missing the point of this chart. It's not supposed to be configurable by external values, since this just adds to functionality to add lose resources. If you would want to achieve what you are looking for, you would be using `.templates`. So yes, it's intentional that this block must be defined again. and what's with using `-f lose-resources.yaml`.

tsunamishaun: Using raw is too prescriptive but luckily the upstream minio-operator chart includes a single tenant configuration for those wanting out of the box functionality. We should include a comment `tentnant: []` which disables the operator from generating a tenant for those looking to define the cr's in their `templates/` directory. That should handle really all the use-cases and keep values.yaml lean and to the point.

oliverbaehler: That does not make too much sense. Well ofc you can stage with the operator a tenant. But what if you manage the operator with another helm release, as most production environments do. You are not able to just add a tenant without also deploying the operator with the minio-operator chart.  For that use-case we use the raw chart.

I will changed the values to make it more clear.

tsunamishaun: You just pass in a blank tenant list. Please see the default here: https://github.com/minio/operator/blob/master/helm/minio-operator/values.yaml#L31

mkilchhofer: Hi @tsunamishaun,
Also in my opinion it is absolutely correct what @oliverbaehler tries to highlight:
Deploying the operator as a subchart from an application chart is very uncommon. The operator normally operates in a different namespace and acts as a clusterlevel operator. Also the operator creates cluster-level CRDs.
Therefore we disable the operator deployment by default. But disabling the subchart also results that we cannot deploy MinIO Tenats via this subchart. He solves this via a raw chart which is a viable approach. An alternative approach would be to define serveral resources in the `templates` directory of the argo workflow chart.

oliverbaehler: Correct, Is this implementation as is acceptable for you @mkilchhofer ?

tsunamishaun: @oliverbaehler this is shaping up nicely, good work. @mkilchhofer thank you for emphasizing the direction, I myself wouldn't rely on the argo chart to give me cr's for a minio tenant but perhaps others would? To that point though, having 200+ lines of unused raw in values.yaml is simplified quite drastically by the reference to the upstream examples `tenants.yaml` and may be unneeded. There are also 10+ ways to deploy a tenant so keeping that open seems like the best choice? (https://github.com/minio/operator/tree/master/examples)

oliverbaehler: Well, who would rely on maging the minio operator with the argo chart? I wouldn't. But we should always try to achieve a gerneric approach, so that everyone can work with the chart and doesn't have to implement his own patched version or first publiush changes to make it work.I removed the samples to make the values slimer.

chgl:
To add my two cents: I think it's unusual to add a (even optional) dependency to an operator to a chart. Similar to the prometheus operator, a ton of charts provide support for ServiceMonitor crds, but very few -if any- provide the operator itself as a dependency. I think it would be great if we could add an option to create a minio tenant (similar to the ServiceMonitor and the example included under the 'raw' key) which is then used to auto-configure the argo artifacts repo.

oliverbaehler:
@chgl Thanks for your input. I was wondering about the discussion on this thread, because I haven't seen this as well (but I didn't want to stick to it just because I haven't seen it yet). So you would propose rendering a minio tenant and create an own structure so that the user can change these values or the raw chart approach, meaning the users kinda have to do it themselves?

chgl:
@oliverbaehler Yes exactly! Basically what's commonly seen with charts depending on a database chart: if the dependency is enabled, the chart is auto-configured to use this database's connection details. I guess the simplest possible Minio tenant instance.
Then again, this might also be needless effort, since most users are just going to use the "raw" chart or create minio tenant crds within the "templates" dir of a chart using argo as a sub-chart - which is what I'd do.

tsunamishaun:
@chgl just to add to this discussion the possibility of using `alias` down the road for including the upstream chart twice to install the operator and to instantiate a tenant. Unfortunately the operator doesn't have conditionals around the install of its components (yet?) but would be a clean/flexible way for us to consume it down the road.

oliverbaehler:
We will leave this one stale until the referenced Pr Request is merged. Maybe until then minio is providing a helm chart that just deploys a tenant, bcs we are not the only ones with that need:

* https://github.com/minio/operator/issues/578

dirtycajunrice:
This has no place as a subchart of an operator. They are distinct products with no correlation outside of your environment. The design that would most closely resemble your goal would be an umbrella with argo and minio as sibling charts

PR #657, opened 2021-04-07
feat(argo-workflows): add value to avoid creating RBAC related to ClusterWorkflowTemplates

g-linville:
Signed-off-by: g-linville <53102776+g-linville@users.noreply.github.com>

Checklist:

The Argo Workflows chart creates some ClusterRoles and ClusterRoleBindings in order to enable both the workflow controller and the server to use and access ClusterWorkflowTemplates. CWTs are cluster-wide resources and do not need to be used in cases where the `singleNamespace` value is set to true, so this commit makes it so that the CRs and CRBs are not installed if that is the case.

UPDATE: based on feedback below, I added two new values (`server.clusterWorkflowTemplates.enabled` and `controller.clusterWorkflowTemplates.enabled`) which can toggle whether the CRs and CRBs are created.


searsaw:
Only issue here is that the controller tries to list the `ClusterWorkflowTemplate`s whether you use them or not and fails the workflow because it won't have permissions to list them.

searsaw:
Turns out that if this `ClusterRoleBinding` doesn't exist, the logic in the workflow controller tells it that it can get `ClusterRoleTemplates`. But because this is a Cluster-scoped CRD, it tries to get them from the Cluster scope and no longer can. The logic gets this assumption because the Role that is created when `namespaced: true` has [these permissions](https://github.com/argoproj/argo-helm/blob/b8f483fb731e9ce40cec5059954119d288686429/charts/argo/templates/workflow-controller-cluster-roles.yaml#L56-L57) to get, list, and watch the `ClusterRoleTemplates`. So I will add to this MR to remove those two lines which fix this logic error.

vladlosev:
What about the situation when a user istalls Argo scoped to a single namespace to allow running workflows just there but wants to use templates from a central library of ClusterWorkflowTemplates in the cluster? This change will break such users.

searsaw:
@vladlosev ClusterWorkflowTemplates are cluster-scoped resources. So you either want to see ClusterWorkflowTemplates which would require a ClusterRoleBinding to be used, or you don't want to access them. Either way, a Role is the wrong way to give the controller access.

vladlosev:
That's correct - the two lines you mentioned have no effect/are redundant and can be removed. But tying the creation of the cluster roles/bindings allowing access to ClusterWorkflowTemplates to the `singleNamespace` setting is a breaking change.

searsaw:
Ahh I understand. Yeah I guess to me, I'd think `namespaced` means it's siloed to a namespace. So it wouldn't need the roles to get access to cluster-wide objects. Looking at [the logic in the workflow controller](https://github.com/argoproj/argo-workflows/blob/master/workflow/controller/controller.go#L397), my assumption is that the Argo project has agreed with this as well.

g-linville:
> Turns out that if this `ClusterRoleBinding` doesn't exist, the logic in the workflow controller tells it that it can get `ClusterRoleTemplates`. But because this is a Cluster-scoped CRD, it tries to get them from the Cluster scope and no longer can. The logic gets this assumption because the Role that is created when `namespaced: true` has [these permissions](https://github.com/argoproj/argo-helm/blob/b8f483fb731e9ce40cec5059954119d288686429/charts/argo/templates/workflow-controller-cluster-roles.yaml#L56-L57) to get, list, and watch the `ClusterRoleTemplates`. So I will add to this MR to remove those two lines which fix this logic error.

I went ahead and removed these two lines in another commit.

vladlosev:
A better, compatible change will be to introduce `.Values.server.clusterWorkflowTemplates.enable`, which will be `true` by default but can be turned off to disable creation of permissions on `ClusterWorkflowTemplate` objects.

kulinskyvs:
Keeping in mind that #668 is merged I suppose this PR should be rather updated for a new chart: argo-workflows. 
@g-linville , could you please update it considering the recommendations from @vladlosev?

g-linville:
> Keeping in mind that #668 is merged I suppose this PR should be rather updated for a new chart: argo-workflows.
> @g-linville , could you please update it considering the recommendations from @vladlosev?

Sounds good! Hopefully I will have this finished on Monday.

mkilchhofer:
> Sounds good! Hopefully I will have this finished on Monday.

Thank you for updating your PR :-) 
I flipped this PR to draft as it needs and update since #668 was merged. You can click "ready for review" once you are done.

g-linville:
It looks like the CircleCI lint is saying that I need to bump the argo-cd chart version. Why would I need to do that? I didn't modify it.

mkilchhofer:
> It looks like the CircleCI lint is saying that I need to bump the argo-cd chart version. Why would I need to do that? I didn't modify it.

Your base branch is out-of-date. Since you force-pushed recently my "merge master into master" commit is gone again. Please click of "Update branch" and then do a git pull on your side if you want to make further changes :-)

g-linville:
> > It looks like the CircleCI lint is saying that I need to bump the argo-cd chart version. Why would I need to do that? I didn't modify it.
> 
> Your base branch is out-of-date. Since you force-pushed recently, my "merge master into master" commit is gone again. Please click of "Update branch" and then do a git pull on your side if you want to make further changes :-)

Thanks. I figured it out and removed a bad merge commit that was in the middle of my PR. All my commits are signed off and I think it's all good now.

vladlosev (approved):
LGTM

mkilchhofer (approved):

Issue #536, opened 2021-01-08
Automate chart dependency and image tag updates using renovate


chgl:
**Is your feature request related to a problem? Please describe.**

Updating the images used within the charts currently depends on manually created PRs to keep up-to-date with new images.

**Describe the solution you'd like**

Tools like renovatebot (https://github.com/renovatebot/renovate) can automatically create PRs for updating image tags and chart dependencies. To test it, I've enabled it in my fork: https://github.com/chgl/argo-helm/pull/1. For an example PR, see: https://github.com/chgl/argo-helm/pull/3.

**Describe alternatives you've considered**

Continue with ad-hoc updates via PRs from the community.

**Additional context**

The current downside is that it does not automatically bump the `version` field in `Chart.yaml` when updating image tag values. It does when updating sub-charts, though: https://github.com/renovatebot/renovate/issues/4728. But this would at least allow for creating automated notifications if there's an update available upstream.

chgl:
Small update: I've now got a PR goin' at https://github.com/renovatebot/renovate/pull/8240 that would automatically bump the chart.yaml when image tags in values.yaml are updated. This should allow for completely automating the update process.

PR #668, opened 2021-04-11
feat(argo-workflows): added new argo-workflows chart and deprecated argo chart

### LINKED ISSUE #483: CRD "workflows.argoproj.io" incorrectly rendered?
**Describe the bug**
When CRD "workflows.argoproj.io" is created, the "spec.additionalPrinterColumns" became part of "spec.versions"

```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    helm.sh/hook: crd-install
    helm.sh/hook-delete-policy: before-hook-creation
  creationTimestamp: "2020-10-29T21:31:03Z"
  generation: 1
  name: workflows.argoproj.io
  resourceVersion: "610788391"
  selfLink: /apis/apiextensions.k8s.io/v1/customresourcedefinitions/workflows.argoproj.io
  uid: 110646a6-297e-42d6-a7e0-42cb13e4a1b1
spec:
  conversion:
    strategy: None
  group: argoproj.io
  names:
    kind: Workflow
    listKind: WorkflowList
    plural: workflows
    shortNames:
    - wf
    singular: workflow
  preserveUnknownFields: true
  scope: Namespaced
  versions:
  - additionalPrinterColumns:
    - description: Status of the workflow
      jsonPath: .status.phase
      name: Status
      type: string
    - description: When the workflow was started
      format: date-time
      jsonPath: .status.startedAt
      name: Age
      type: date
    name: v1alpha1
    served: true
    storage: true
status:
  acceptedNames:
    kind: Workflow
    listKind: WorkflowList
    plural: workflows
    shortNames:
    - wf
    singular: workflow
  conditions:
  - lastTransitionTime: "2020-10-29T21:31:03Z"
    message: no conflicts found
    reason: NoConflicts
    status: "True"
    type: NamesAccepted
  - lastTransitionTime: "2020-10-29T21:31:03Z"
    message: the initial names have been accepted
    reason: InitialNamesAccepted
    status: "True"
    type: Established
  storedVersions:
  - v1alpha1
```

See the correct definition of CRD "workflows.argoproj.io" https://github.com/argoproj/argo-helm/blob/master/charts/argo/templates/workflow-crd.yaml#L10 and https://github.com/argoproj/argo-helm/blob/master/charts/argo/crds/workflow-crd.yaml#L9

**To Reproduce**
1.  'git clone git@github.com:argoproj/argo-helm.git'
2. Run 'helm install --debug --namespace argo --name argo ./argo --set controller.containerRuntimeExecutor=k8sapi'
3. Run 'kubectl get crd/workflows.argoproj.io -o yaml'
4. Look for "additionalPrinterColumns"

**Expected behavior**
"additionalPrinterColumns" should be a spec of the CRD, not a version.

**Screenshots**

**Additional context**

### LINKED ISSUE #636: Workflow CRDs are outdated and the official repo has supported CRDS
Hey! Appreciate the work yall have done on this project. I've been looking to upgrade my CRDs and I found that the official repo has updated v1 CRDs for Workflows
https://github.com/argoproj/argo-workflows/blob/master/manifests/install.yaml

What do you think about either supporting the new version here or redirecting to the official repo?

### LINKED ISSUE #646: Rename "argo" chart to "argo-workflows" to be consistent with the application repo

The upstream argo repo has recently been renamed to argo-workflows. For consistency, we might rename the helm chart as well.

Deprecate the "argo" helm chart and create a new "argo-workflows" chart. Eventually deleting the old "argo" one.

Not doing this.

**Additional context**

The benefit isn't gigantic except for improving consistency in naming. The new argo-workflows 3.0 release may be a good reason to pool these breaking changes, though.

### LINKED ISSUE #660: [argo-workflows] server.secure=false broken with tag 3.0.1
**Describe the bug**
With argo-workflows [tag 3.0.1](https://github.com/argoproj/argo-workflows/releases/tag/v3.0.1), "The server now starts with TLS enabled by default if a key is available. This can be disabled using --secure=false." 

However, the current Helm chart will only modify the `--secure` argument if `server.secure` is true.  You can see that on [this line](https://github.com/argoproj/argo-helm/blob/master/charts/argo/templates/server-deployment.yaml#L42). But now we now need it to explicitly pass false.

Also the link [at this line](https://github.com/argoproj/argo-helm/blob/master/charts/argo/values.yaml#L209) should be updated to https://argoproj.github.io/argo-workflows/tls/#encrypted

**To Reproduce**
Use the existing helm chart, modify `images.tag` to `v3.0.1` and set `server.secure` false.

**Expected behavior**
Expect `argo-server` to come up... but it doesn't because it's readiness probe is using `HTTP` while the server is erroneously using `HTTPS` (see `secure=true` in the log).

**Additional context**
Workaround is to both set `server.secure` to false AND to add `--secure=false` to any `server.extraArgs`

chgl:
Resolves #660:

uses the workaround of setting `--secure=false` by default: https://github.com/argoproj/argo-workflows/issues/5607

resolves #646:

renames the chart (while keeping the old one around but marking as deprecated)

resolves #636:

copied all the CRDs from https://github.com/argoproj/argo-workflows/tree/master/manifests/base/crds/minimal and updated clusterroles for the new workfloweventbindings resource

resolves #483:

while the apiextension version was bumped to v1, the resource spec was not updated correctly. This is solved by resolving #636 as well.

I've documented the breaking changes in the README. I tried not to change *too* much, but we could definitely break more - now is the time!

I've removed any included minio support for now until https://github.com/argoproj/argo-helm/pull/649 is resolved - so we can add it back later with a clean solution.

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -4,7 +4,7 @@ description: A Helm chart for Argo Workflows
 type: application
 version: 0.1.0
 appVersion: "v3.0.1"
-icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
+icon: https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 sources:
   - https://github.com/argoproj/argo-workflows
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
 version: 0.1.0
-appVersion: "v3.0.1"
+appVersion: "v3.0.2"
 icon: https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 sources:
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,6 +3,7 @@ name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
 version: 0.1.0
+appVersion: "v3.0.1"
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 sources:
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -1,18 +1,14 @@
 apiVersion: v2
-appVersion: v2.12.5
+name: argo-workflows
 description: A Helm chart for Argo Workflows
-name: argo
-version: 0.16.10
+type: application
+version: 0.1.0
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
+sources:
+  - https://github.com/argoproj/argo-workflows
 maintainers:
   - name: alexec
   - name: alexmt
   - name: jessesuen
   - name: benjaminws
-dependencies:
-- name: minio
-  version: 8.0.9
-  repository: https://helm.min.io/
-  condition: minio.install
-deprecated: true
--- a/charts/argo-workflows/templates/NOTES.txt
+++ b/charts/argo-workflows/templates/NOTES.txt
@@ -4,4 +4,4 @@ kubectl --namespace {{ .Release.Namespace }} get services -o wide | grep {{ .Rel
 
 2. Submit the hello-world workflow by running:
 
-argo submit https://raw.githubusercontent.com/argoproj/argo/master/examples/hello-world.yaml --watch
+argo submit https://raw.githubusercontent.com/argoproj/argo-workflows/master/examples/hello-world.yaml --watch
--- a/charts/argo-workflows/templates/NOTES.txt
+++ b/charts/argo-workflows/templates/NOTES.txt
@@ -1,0 +1,7 @@
+1. Get Argo Server external IP/domain by running:
+
+kubectl --namespace {{ .Release.Namespace }} get services -o wide | grep {{ .Release.Name }}-{{ .Values.server.name }}
+
+2. Submit the hello-world workflow by running:
+
+argo submit https://raw.githubusercontent.com/argoproj/argo/master/examples/hello-world.yaml --watch
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -1,0 +1,16 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+{{- define "fullname" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -94,3 +94,21 @@ Return the appropriate apiVersion for ingress
 {{- print "networking.k8s.io/v1" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the appropriate service layout for ingress
+*/}}
+{{- define "argo-workflows.ingress.service" -}}
+{{- if semverCompare "<1.19-0" .kubeVersion -}}
+- backend:
+    serviceName: {{ .serviceName }}
+    servicePort: {{ .servicePort }}
+{{- else -}}
+- backend:
+    service:
+      name: {{ .serviceName }}
+      port: 
+        number: {{ .servicePort }}
+  pathType: ImplementationSpecific
+{{- end -}}
+{{- end -}}
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -86,29 +86,11 @@ Create the name of the controller service account to use
 Return the appropriate apiVersion for ingress
 */}}
 {{- define "argo-workflows.ingress.apiVersion" -}}
-{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.Version -}}
 {{- print "extensions/v1beta1" -}}
-{{- else if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- else if semverCompare "<1.19-0" .Capabilities.KubeVersion.Version -}}
 {{- print "networking.k8s.io/v1beta1" -}}
 {{- else -}}
 {{- print "networking.k8s.io/v1" -}}
-{{- end -}}
-{{- end -}}
-
-{{/*
-Return the appropriate service layout for ingress
-*/}}
-{{- define "argo-workflows.ingress.service" -}}
-{{- if semverCompare "<1.19-0" .kubeVersion -}}
-- backend:
-    serviceName: {{ .serviceName }}
-    servicePort: {{ .servicePort }}
-{{- else -}}
-- backend:
-    service:
-      name: {{ .serviceName }}
-      port:
-        number: {{ .servicePort }}
-  pathType: ImplementationSpecific
 {{- end -}}
 {{- end -}}
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -1,16 +1,96 @@
 {{/* vim: set filetype=mustache: */}}
+
+{{/*
+Create argo workflows server name and version as used by the chart label.
+*/}}
+{{- define "argo-workflows.server.fullname" -}}
+{{- printf "%s-%s" (include "argo-workflows.fullname" .) .Values.server.name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create controller name and version as used by the chart label.
+*/}}
+{{- define "argo-workflows.controller.fullname" -}}
+{{- printf "%s-%s" (include "argo-workflows.fullname" .) .Values.controller.name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
 {{/*
 Expand the name of the chart.
 */}}
-{{- define "name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
-{{- end -}}
+{{- define "argo-workflows.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
 
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
-{{- define "fullname" -}}
+{{- define "argo-workflows.fullname" -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "argo-workflows.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "argo-workflows.labels" -}}
+helm.sh/chart: {{ include "argo-workflows.chart" .context }}
+{{ include "argo-workflows.selectorLabels" (dict "context" .context "component" .component "name" .name) }}
+app.kubernetes.io/managed-by: {{ .context.Release.Service }}
+app.kubernetes.io/part-of: argo-workflows
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "argo-workflows.selectorLabels" -}}
+{{- if .name -}}
+app.kubernetes.io/name: {{ include "argo-workflows.name" .context }}-{{ .name }}
+{{ end -}}
+app.kubernetes.io/instance: {{ .context.Release.Name }}
+{{- if .component }}
+app.kubernetes.io/component: {{ .component }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create the name of the server service account to use
+*/}}
+{{- define "argo-workflows.serverServiceAccountName" -}}
+{{- if .Values.server.createServiceAccount -}}
+    {{ default (include "argo-workflows.fullname" .) .Values.server.serviceAccount }}
+{{- else -}}
+    {{ default "default" .Values.server.serviceAccount }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create the name of the controller service account to use
+*/}}
+{{- define "argo-workflows.controllerServiceAccountName" -}}
+{{- if .Values.controller.createServiceAccount -}}
+    {{ default (include "argo-workflows.fullname" .) .Values.controller.serviceAccount }}
+{{- else -}}
+    {{ default "default" .Values.controller.serviceAccount }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the appropriate apiVersion for ingress
+*/}}
+{{- define "argo-cd.ingress.apiVersion" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- print "extensions/v1beta1" -}}
+{{- else if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- print "networking.k8s.io/v1beta1" -}}
+{{- else -}}
+{{- print "networking.k8s.io/v1" -}}
+{{- end -}}
 {{- end -}}
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -85,7 +85,7 @@ Create the name of the controller service account to use
 {{/*
 Return the appropriate apiVersion for ingress
 */}}
-{{- define "argo-cd.ingress.apiVersion" -}}
+{{- define "argo-workflows.ingress.apiVersion" -}}
 {{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
 {{- print "extensions/v1beta1" -}}
 {{- else if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion -}}
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -64,10 +64,10 @@ app.kubernetes.io/component: {{ .component }}
 Create the name of the server service account to use
 */}}
 {{- define "argo-workflows.serverServiceAccountName" -}}
-{{- if .Values.server.createServiceAccount -}}
-    {{ default (include "argo-workflows.fullname" .) .Values.server.serviceAccount }}
+{{- if .Values.server.serviceAccount.create -}}
+    {{ default (include "argo-workflows.fullname" .) .Values.server.serviceAccount.name }}
 {{- else -}}
-    {{ default "default" .Values.server.serviceAccount }}
+    {{ default "default" .Values.server.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
 
@@ -75,10 +75,10 @@ Create the name of the server service account to use
 Create the name of the controller service account to use
 */}}
 {{- define "argo-workflows.controllerServiceAccountName" -}}
-{{- if .Values.controller.createServiceAccount -}}
-    {{ default (include "argo-workflows.fullname" .) .Values.controller.serviceAccount }}
+{{- if .Values.controller.serviceAccount.create -}}
+    {{ default (include "argo-workflows.fullname" .) .Values.controller.serviceAccount.name }}
 {{- else -}}
-    {{ default "default" .Values.controller.serviceAccount }}
+    {{ default "default" .Values.controller.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
 
@@ -107,7 +107,7 @@ Return the appropriate service layout for ingress
 - backend:
     service:
       name: {{ .serviceName }}
-      port: 
+      port:
         number: {{ .servicePort }}
   pathType: ImplementationSpecific
 {{- end -}}
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -15,8 +15,8 @@ spec:
       labels:
         {{- include "argo-workflows.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 8 }}
         app.kubernetes.io/version: {{ default .Chart.AppVersion .Values.controller.image.tag | quote }}
-        {{- if .Values.controller.podLabels }}
-        {{ toYaml .Values.controller.podLabels | nindent 8}}
+        {{- with.Values.controller.podLabels }}
+          {{- toYaml . | nindent 8 }}
         {{- end }}
       {{- with .Values.controller.podAnnotations }}
       annotations:
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -32,14 +32,14 @@ spec:
       {{- end }}
       containers:
         - name: controller
-          image: "{{ .Values.controller.image.registry }}/{{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag }}"
+          image: "{{ .Values.controller.image.registry }}/{{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.images.pullPolicy }}
           command: [ "workflow-controller" ]
           args:
           - "--configmap"
           - "{{ .Release.Name }}-{{ .Values.controller.name}}-configmap"
           - "--executor-image"
-          - "{{ .Values.executor.image.registry }}/{{ .Values.executor.image.repository }}:{{ .Values.executor.image.tag }}"
+          - "{{ .Values.executor.image.registry }}/{{ .Values.executor.image.repository }}:{{ .Values.executor.image.tag | default .Chart.AppVersion }}"
           - "--loglevel"
           - "{{ .Values.controller.logging.level }}"
           - "--gloglevel"
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -99,6 +99,6 @@ spec:
       affinity:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- if .Values.controller.priorityClassName }}
-      priorityClassName: {{ .Values.controller.priorityClassName }}
+      {{- with .Values.controller.priorityClassName }}
+      priorityClassName: {{ . }}
       {{- end }}
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -76,23 +76,15 @@ spec:
           {{- end }}
           resources:
             {{- toYaml .Values.controller.resources | nindent 12 }}
-          {{- if .Values.controller.metricsConfig.enabled }}
           ports:
             - name: metrics
-              containerPort: {{- .Values.controller.metricsConfig.port }}
+              containerPort: {{ .Values.controller.metricsConfig.port }}
           livenessProbe:
             httpGet:
               port: metrics
-              path: {{- .Values.controller.metricsConfig.path }}
+              path: {{ .Values.controller.metricsConfig.path }}
             initialDelaySeconds: 30
             periodSeconds: 30
-          readinessProbe:
-            httpGet:
-              port: metrics
-              path: {{- .Values.controller.metricsConfig.path }}
-            initialDelaySeconds: 30
-            periodSeconds: 30
-          {{- end }}
       {{- with .Values.images.pullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -53,8 +53,8 @@ spec:
           - "--pod-workers"
           - {{ . | quote }}
           {{- end }}
-          {{- if .Values.controller.extraArgs }}
-          {{- toYaml .Values.controller.extraArgs | nindent 10 }}
+          {{- with .Values.controller.extraArgs }}
+            {{- toYaml . | nindent 10 }}
           {{- end }}
           securityContext:
             {{- toYaml .Values.controller.securityContext | nindent 12 }}
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -24,9 +24,9 @@ spec:
       {{- end }}
     spec:
       serviceAccountName: {{ template "argo-workflows.controllerServiceAccountName" . }}
-      {{- if .Values.controller.podSecurityContext }}
+      {{- with .Values.controller.podSecurityContext }}
       securityContext:
-        {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
       {{- end }}
       containers:
         - name: controller
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -18,9 +18,10 @@ spec:
         {{- if .Values.controller.podLabels }}
         {{ toYaml .Values.controller.podLabels | nindent 8}}
         {{- end }}
-      {{- if .Values.controller.podAnnotations }}
+      {{- with .Values.controller.podAnnotations }}
       annotations:
-{{ toYaml .Values.controller.podAnnotations | indent 8}}{{- end }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
     spec:
       serviceAccountName: {{ template "argo-workflows.controllerServiceAccountName" . }}
       {{- if .Values.controller.podSecurityContext }}
--- a/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
@@ -1,0 +1,114 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}-{{ .Values.controller.name}}
+  labels:
+    app: {{ .Release.Name }}-{{ .Values.controller.name}}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  replicas: {{ .Values.controller.replicas }}
+  selector:
+    matchLabels:
+      app: {{ .Release.Name }}-{{ .Values.controller.name}}
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}-{{ .Values.controller.name}}
+        release: {{ .Release.Name }}
+        {{- if .Values.controller.podLabels }}
+        {{ toYaml .Values.controller.podLabels | nindent 8}}
+        {{- end }}
+      {{- if .Values.controller.podAnnotations }}
+      annotations:
+{{ toYaml .Values.controller.podAnnotations | indent 8}}{{- end }}
+    spec:
+      serviceAccountName: {{ .Values.controller.serviceAccount | quote }}
+      {{- if .Values.controller.podSecurityContext }}
+      securityContext:
+        {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: controller
+          image: "{{ .Values.controller.image.registry }}/{{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag }}"
+          imagePullPolicy: {{ .Values.images.pullPolicy }}
+          command: [ "workflow-controller" ]
+          args:
+          - "--configmap"
+          - "{{ .Release.Name }}-{{ .Values.controller.name}}-configmap"
+          - "--executor-image"
+          - "{{ .Values.executor.image.registry }}/{{ .Values.executor.image.repository }}:{{ .Values.executor.image.tag }}"
+          - "--loglevel"
+          - "{{ .Values.controller.logging.level }}"
+          - "--gloglevel"
+          - "{{ .Values.controller.logging.globallevel }}"
+          {{- if .Values.singleNamespace }}
+          - "--namespaced"
+          {{- end }}
+          {{- with .Values.controller.workflowWorkers }}
+          - "--workflow-workers"
+          - {{ . | quote }}
+          {{- end }}
+          {{- with .Values.controller.podWorkers }}
+          - "--pod-workers"
+          - {{ . | quote }}
+          {{- end }}
+          {{- if .Values.controller.extraArgs }}
+          {{- toYaml .Values.controller.extraArgs | nindent 10 }}
+          {{- end }}
+          securityContext:
+            {{- toYaml .Values.controller.securityContext | nindent 12 }}
+          env:
+            - name: ARGO_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.namespace
+            - name: LEADER_ELECTION_IDENTITY
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.name
+          {{- with .Values.controller.extraEnv }}
+            {{ toYaml . | nindent 10 }}
+          {{- end }}
+          resources:
+            {{- toYaml .Values.controller.resources | nindent 12 }}
+          {{- if .Values.controller.metricsConfig.enabled }}
+          ports:
+            - name: metrics
+              containerPort: {{- .Values.controller.metricsConfig.port }}
+          livenessProbe:
+            httpGet:
+              port: metrics
+              path: {{- .Values.controller.metricsConfig.path }}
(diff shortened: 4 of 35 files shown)


charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+                  fieldPath: metadata.name
+          {{- with .Values.controller.extraEnv }}
+            {{ toYaml . | nindent 10 }}
+          {{- end }}
+          resources:
+            {{- toYaml .Values.controller.resources | nindent 12 }}
+          {{- if .Values.controller.metricsConfig.enabled }}
+          ports:
+            - name: metrics
+              containerPort: {{- .Values.controller.metricsConfig.port }}
```
ChiefAlexander, line 82: You need to remove the `-` to avoid templating issues.

```suggestion
              containerPort: {{ .Values.controller.metricsConfig.port }}
```

charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+          livenessProbe:
+            httpGet:
+              port: metrics
+              path: {{- .Values.controller.metricsConfig.path }}
```
ChiefAlexander, line 86:
```suggestion
              path: {{ .Values.controller.metricsConfig.path }}
```

charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+            initialDelaySeconds: 30
+            periodSeconds: 30
+          readinessProbe:
+            httpGet:
+              port: metrics
+              path: {{- .Values.controller.metricsConfig.path }}
```
ChiefAlexander, line 92:
```suggestion
              path: {{ .Values.controller.metricsConfig.path }}
```

ChiefAlexander (changes requested):

charts/argo-workflows/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v2
+name: argo-workflows
```
psmit, line 2: Just out of curiosity; you are not defining an appVersion and set the tag right away in the values.yaml. Does this have a specific reason?

charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
chgl: argh, thanks for noticing. Will fix asap. In general, I am not too happy with the way the metricsConfig currently works - I kept the changes minimal but I think it may be better to enable the metrics port at all times and not depending on whether the `metricsConfig.enabled` is set. The major advantage is that liveness and readiness probes are always enabled.

charts/argo-workflows/Chart.yaml
chgl: The main motivation (or why I am personally always doing it) is to support automated image tag updates in the future, see https://github.com/argoproj/argo-helm/issues/536. If all container images are set consistently from within the values.yaml, then updates are possible out-of-the-box (at least for renovatebot should we ever consider using it).

It might make sense to set `appVersion: "v3"`?

charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
ChiefAlexander: I missed that the probes relayed on metrics. I agree this should always be on so that the pods have their probes to determine health.

charts/argo-workflows/Chart.yaml
psmit: I see that tools like renovatebot might have trouble with the appVersion; but I believe it would be "the other way around" to now for this chart to not adopt the industry-standard practices, instead of finding a way for the other tools to support the appVersion attribute.

I would propose to use the helm-standard "appVersion" to contain the argo version and let image.tag default to .Chart.AppVersion

charts/argo-workflows/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: A Helm chart for Argo Workflows
+type: application
+version: 0.1.0
+icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
```
psmit, line 6: This link should be updated to https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png

charts/argo-workflows/templates/NOTES.txt
```diff
@@ -0,0 +1,7 @@
+1. Get Argo Server external IP/domain by running:
+
+kubectl --namespace {{ .Release.Namespace }} get services -o wide | grep {{ .Release.Name }}-{{ .Values.server.name }}
+
+2. Submit the hello-world workflow by running:
+
+argo submit https://raw.githubusercontent.com/argoproj/argo/master/examples/hello-world.yaml --watch
```
psmit, line 7: This link should be updated to https://raw.githubusercontent.com/argoproj/argo-workflows/master/examples/hello-world.yaml

charts/argo-workflows/values.yaml
```diff
@@ -0,0 +1,325 @@
...
+  ##
+  nodeSelector:
+    kubernetes.io/os: linux
+  tolerations: []
+  affinity: {}
+  # Leverage a PriorityClass to ensure your pods survive resource shortages
+  # ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
+  # PriorityClass: system-cluster-critical
+  priorityClassName: ""
+  # https://argoproj.github.io/argo/links/
```
psmit, line 141: This link needs to be updated to https://argoproj.github.io/argo-workflows/links/

charts/argo-workflows/values.yaml
```diff
@@ -0,0 +1,325 @@
...
+  # Adds environment variables for the executor.
+  env: {}
+  # sets security context for the executor container
+  securityContext: {}
+
+server:
+  enabled: true
+  # only updates base url of resources on client side,
+  # it's expected that a proxy server rewrites the request URL and gets rid of this prefix
+  # https://github.com/argoproj/argo/issues/716#issuecomment-433213190
```
psmit, line 160: Should be updated to https://github.com/argoproj/argo-workflows/issues/716#issuecomment-433213190

charts/argo-workflows/Chart.yaml
chgl: I'm not completely convinced yet, but since my use case is at this point entirely hypothetical I have reverted back to using appVersion.

charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
```diff
@@ -0,0 +1,106 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}-{{ .Values.controller.name}}
+  labels:
+    app: {{ .Release.Name }}-{{ .Values.controller.name}}
```
chgl, line 6: the deployments are currently using the "old-style" labels while newer deployments/charts generally use the recommended labels such as `app.kubernetes.io/name` (see https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

Should we update these as well?

oliverbaehler: I was wondering if we should add a library chart that has the same labels/ selectors etc. for all charts. We recently implemented the new-style labels in the argo-cd chart. I would suggest for now we leave it old style but i will create an issue addressing this for all charts.

chgl: Sounds good!

ChiefAlexander: I'd personally like to see these updated to the recommended style. However doing so is a breaking change as selector labels are immutable.

chgl: You're right! I'm OK with either approach. I've looked at https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/_helpers.tpl and could definitely copy-and-paste them over to this chart for now until we have a library chart available. Since this is the first version of this chart and a 0.* release, most people will need to freshly install anyway.

chgl:
I can't access the circleci logs unfortunately, but running the lint.sh script locally yields `Error: Looks like "https://kubernetes-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: Failed to fetch https://kubernetes-charts.storage.googleapis.com/index.yaml : 403 Forbidden` - running `helm init --stable-repo-url=https://charts.helm.sh/stable --client-only` seems to fix it though. But since other PRs are fine I think I'm doing something else wrong.

oliverbaehler:
@chgl You should be able to view the circle-ci logs. Click on details and then login via Github. It requires a version bump of the old `argo` chart:

```
 ✖︎ argo => (version: "0.16.9", path: "charts/argo") > Chart version not ok. Needs a version bump!
``` 
Well that's what chart-linter is causing to fail.

chgl:
Ah, thank you! I just linted argo-workflows instead of the entire charts/ dir locally. 

I'm a bit hesitant to grant circleci access, so I'm definitely looking forward to your PR https://github.com/argoproj/argo-helm/pull/685 being merged!

oliverbaehler:
@chgl Understandable. Doesn't look like anyone wants to drive it forward but let's hope for the best ^^

ionutz89:
Can we merge this in master? I am getting issue with the current helm chart for workflow. It seems the dependencies of minio helm chart makes deploy unsuccessful.  I tested this PR and it is working for me as a new fresh install via ArgoCD.

oliverbaehler:
@ionutz89 I will try to review and merge it in the next days

charts/argo-workflows/templates/server/server-deployment.yaml
```diff
@@ -0,0 +1,104 @@
...
+        {{- end }}
+      {{- if .Values.server.podAnnotations }}
+      annotations:
+{{ toYaml .Values.server.podAnnotations | indent 8}}{{- end }}
+    spec:
+      serviceAccountName: {{ .Values.server.serviceAccount | quote }}
+      {{- if .Values.server.podSecurityContext }}
+      securityContext:
+        {{- toYaml .Values.server.podSecurityContext | nindent 8 }}
+      {{- end }}
```
oliverbaehler, line 30:
```suggestion
      {{- with .Values.server.podSecurityContext }}
      securityContext:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

charts/argo-workflows/templates/server/server-deployment.yaml
```diff
@@ -0,0 +1,104 @@
...
+          image: "{{ .Values.server.image.registry }}/{{ .Values.server.image.repository }}:{{ .Values.server.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.images.pullPolicy }}
+          securityContext:
+            {{- toYaml .Values.server.securityContext | nindent 12 }}
+          args:
+          - server
+          - --configmap={{ .Release.Name }}-{{ .Values.controller.name }}-configmap
+          {{- if .Values.server.extraArgs }}
+          {{- toYaml .Values.server.extraArgs | nindent 10 }}
+          {{- end }}
```
oliverbaehler, line 42:
```suggestion
          {{- with .Values.server.extraArgs }}
            {{- toYaml . | nindent 10 }}
          {{- end }}
```

charts/argo-workflows/templates/server/server-deployment.yaml
```diff
@@ -0,0 +1,104 @@
...
+      {{- with .Values.server.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.server.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if .Values.server.priorityClassName }}
+      priorityClassName: {{ .Values.server.priorityClassName }}
```
oliverbaehler, line 102:
```suggestion
      {{- with .Values.server.priorityClassName }}
      priorityClassName: {{ . }}
```

charts/argo-workflows/templates/server/server-deployment.yaml
```diff
@@ -0,0 +1,104 @@
...
+    matchLabels:
+      {{- include "argo-workflows.selectorLabels" (dict "context" . "name" .Values.server.name) | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "argo-workflows.labels" (dict "context" . "component" .Values.server.name "name" .Values.server.name) | nindent 8 }}
+        app.kubernetes.io/version: {{ default .Chart.AppVersion .Values.server.image.tag | quote }}
+        {{- if .Values.server.podLabels }}
+        {{- toYaml .Values.server.podLabels | nindent 8 }}
+        {{- end }}
```
oliverbaehler, line 21:
```suggestion
        {{- with .Values.server.podLabels }}
          {{- toYaml . | nindent 8 }}
        {{- end }}
```

charts/argo-workflows/templates/server/server-sa.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if and .Values.server.enabled .Values.server.createServiceAccount -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "argo-workflows.serverServiceAccountName" . }}
+  annotations:
+{{ toYaml .Values.server.serviceAccountAnnotations | indent 4 }}
```
oliverbaehler, line 7:
```suggestion
  {{- with .Values.server.serviceAccountAnnotations  }}
  annotations:
    {{- toYaml . | indent 4 }}
  {{- end }}
```

charts/argo-workflows/templates/server/server-service.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "argo-workflows.server.fullname" . }}
+  labels:
+    {{- include "argo-workflows.labels" (dict "context" . "component" .Values.server.name "name" .Values.server.name) | nindent 4 }}
+    app.kubernetes.io/version: {{ default .Chart.AppVersion .Values.server.image.tag | quote }}
+  {{- if .Values.server.serviceAnnotations }}
+  annotations:
+{{ toYaml .Values.server.serviceAnnotations | indent 4}}{{- end }}
```
oliverbaehler, line 11:
```suggestion
  {{- with .Values.server.serviceAnnotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

charts/argo-workflows/templates/server/server-service.yaml
```diff
@@ -0,0 +1,29 @@
...
+  selector:
+    {{- include "argo-workflows.selectorLabels" (dict "context" . "name" .Values.server.name) | nindent 4 }}
+  sessionAffinity: None
+  type: {{ .Values.server.serviceType }}
+  {{- if and (eq .Values.server.serviceType "LoadBalancer") .Values.server.loadBalancerIP }}
+  loadBalancerIP: {{ .Values.server.loadBalancerIP | quote }}
+  {{- end }}
+  {{- if and (eq .Values.server.serviceType "LoadBalancer") .Values.server.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges:
+{{ toYaml .Values.server.loadBalancerSourceRanges | indent 4 }}{{- end }}
```
oliverbaehler, line 28:
```suggestion
  loadBalancerSourceRanges:
    {{- toYaml .Values.server.loadBalancerSourceRanges | nindent 4 }}
  {{- end }}
```

charts/argo-workflows/templates/server/server-service.yaml
```diff
@@ -0,0 +1,29 @@
...
+    {{- include "argo-workflows.labels" (dict "context" . "component" .Values.server.name "name" .Values.server.name) | nindent 4 }}
+    app.kubernetes.io/version: {{ default .Chart.AppVersion .Values.server.image.tag | quote }}
+  {{- if .Values.server.serviceAnnotations }}
+  annotations:
+{{ toYaml .Values.server.serviceAnnotations | indent 4}}{{- end }}
+spec:
+  ports:
+  - port: {{ .Values.server.servicePort }}
+    {{- if .Values.server.servicePortName }}
+    name: {{ .Values.server.servicePortName }}
```
oliverbaehler, line 16:
```suggestion
    {{- with .Values.server.servicePortName }}
    name: {{ . }}
```

charts/argo/Chart.yaml
```diff
@@ -5,13 +5,9 @@ name: argo
 version: 0.16.10
```
oliverbaehler, line 5: We also bumped to a new major when deprecating argo-ci, so i would do the same here

```suggestion
version: 1.0.0
```

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,39 @@
...
+          {{- end }}
+          - backend:
+              serviceName: {{ $serviceName }}
+              servicePort: {{ $servicePort }}
+    {{- end -}}
+  {{- if .Values.server.ingress.tls }}
+  tls:
+{{ toYaml .Values.server.ingress.tls | indent 4 }}
+  {{- end -}}
+{{- end -}}
```
oliverbaehler, line 38:
```suggestion
    {{- with .Values.server.ingress.tls }}
  tls:
      {{- toYaml . | nindent 4 }}
    {{- end -}}
  {{- end -}}
```

charts/argo-workflows/templates/controller/workflow-controller-servicemonitor.yaml
```diff
@@ -0,0 +1,29 @@
...
+spec:
+  endpoints:
+  {{- if .Values.controller.metricsConfig.enabled }}
+    - port: metrics
+      path: {{ .Values.controller.metricsConfig.path }}
+      interval: 30s
+  {{- end }}
+  {{- if .Values.controller.telemetryConfig.enabled }}
+    - port: telemetry
+      path: {{ .Values.controller.telemetryConfig.path }}
```
oliverbaehler, line 20:
```suggestion
  {{- with .Values.controller.telemetryConfig.enabled }}
    - port: telemetry
      path: {{ . }}
```

charts/argo-workflows/templates/controller/workflow-controller-servicemonitor.yaml
```diff
@@ -0,0 +1,29 @@
...
+  labels:
+    {{- include "argo-workflows.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
+    {{- if .Values.controller.serviceMonitor.additionalLabels }}
+{{ toYaml .Values.controller.serviceMonitor.additionalLabels | indent 4 }}
+    {{- end }}
+spec:
+  endpoints:
+  {{- if .Values.controller.metricsConfig.enabled }}
+    - port: metrics
+      path: {{ .Values.controller.metricsConfig.path }}
```
oliverbaehler, line 15:
```suggestion
  {{- with .Values.controller.metricsConfig.enabled }}
    - port: metrics
      path: {{ . }}
```

charts/argo-workflows/templates/controller/workflow-controller-sa.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "argo-workflows.controllerServiceAccountName" . }}
+  annotations:
+{{ toYaml .Values.controller.serviceAccountAnnotations | indent 4 }}
```
oliverbaehler, line 6:
```suggestion
  {{ with .Values.controller.serviceAccountAnnotations }}
  annotations:
    {{- toYaml .| nindent 4 }}
  {{- end }}
```

charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
```diff
@@ -0,0 +1,103 @@
...
+        {{- end }}
+      {{- if .Values.controller.podAnnotations }}
+      annotations:
+{{ toYaml .Values.controller.podAnnotations | indent 8}}{{- end }}
+    spec:
+      serviceAccountName: {{ template "argo-workflows.controllerServiceAccountName" . }}
+      {{- if .Values.controller.podSecurityContext }}
+      securityContext:
+        {{- toYaml .Values.controller.podSecurityContext | nindent 8 }}
+      {{- end }}
```
oliverbaehler, line 29:
```suggestion
      {{- with .Values.controller.podSecurityContext }}
      securityContext:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

oliverbaehler (changes requested):
@chgl I hope you don't hate me too much for this. I have corrected all the spots where if was used instead of `with` to keep a better consistency. I corrected some indendting and replaced indent with nindent. I think there's one proposal in the rbac which changes coniditions but the reast is just cleaning up. I thought it would be a good opportunity and further contributions now face a cleaner code. What do you think?

oliverbaehler:
@chgl Also, don't you think ye just should keep the minio dependency for now? As mentioned in the related pr to the issue there's a demand for a minio tenant chart. But they are still missing stuff like default Bucketsetc. on minio tenants. And I would wait minio releases a tenant chart themselves. As far as I am concerned I would keep the minio dependency.

@ionutz89 What problem are you seeing with the minio dependency? Because currently you won't  have a minio dependency at all, which is also already supported via values

ionutz89:
> @ionutz89 What problem are you seeing with the minio dependency? Because currently you won't have a minio dependency at all, which is also already supported via values

@oliverbaehler I am getting this error:
`rpc error: code = Unknown desc = helm dependency build failed exit status 1: Error: no repository definition for https://helm.min.io/. Please add the missing repos via 'helm repo add'` via ArgoCD install with all defaults settings. 

In value.yaml file we have this currently:
```
minio:
  install: false
  ```
  
 It's set to false. I tried to overwrite this value but no luck.

**Comment by oliverbaehler** [comment]
@ionutz89 I am aware of that, since I am working on fixing this lifecycle. However this does not instantly deprecate the old dependency and we are not gonna deploy an operator as dependency of argo. Before the following issues aren't handled by minio we are not gonna change to the operator based minio instances.

> via ArgoCD install with all defaults settings

Not exactly sure what that means, but for me it works fine with deploying the default values (there's no reason to contact the minio repository with the default values):

```
 helm install my-argo argo/argo --version 0.16.10 -n argo-work
W0427 21:40:19.627094  232593 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:40:19.632717  232593 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:40:19.638812  232593 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:40:19.644851  232593 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:40:19.650380  232593 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
NAME: my-argo
LAST DEPLOYED: Tue Apr 27 21:40:19 2021
NAMESPACE: argo-work
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
1. Get Argo Server external IP/domain by running:

kubectl --namespace argo-work get services -o wide | grep my-argo-server

2. Submit the hello-world workflow by running:

argo submit https://raw.githubusercontent.com/argoproj/argo/master/examples/hello-world.yaml --watc
```

And with Minio enabled:

``` 
W0427 21:42:45.244461  242888 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:42:45.250265  242888 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:42:45.256166  242888 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:42:45.262123  242888 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 21:42:45.268457  242888 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
NAME: my-argo
LAST DEPLOYED: Tue Apr 27 21:42:45 2021
NAMESPACE: argo-work
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
1. Get Argo Server external IP/domain by running:

kubectl --namespace argo-work get services -o wide | grep my-argo-server

2. Submit the hello-world workflow by running:

argo submit https://raw.githubusercontent.com/argoproj/argo/master/examples/hello-world.yaml --watch



3. Access Minio UI and create bucket 'argo-artifacts'. Minio UI is available on port 9000 and available via external URL. URL might be retrieved using following
command:

kubectl --namespace argo-work get services -o wide | grep my-argo-minio
```
So your problem seems to be somewhere else.

**Comment by ionutz89** [comment]
@oliverbaehler thank you for checking out. Appreciated your very quick reply. 🍺 

The output from the previous command was from the `ArgoCD UI` interface.
If it is required, I can provide screenshot.

This is config.yaml file tested:

```
---
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: argocd-workflow
  namespace: argocd
  finalizers:
    - resources-finalizer.argocd.argoproj.io
spec:
  destination:
    namespace: argocd
    server: https://kubernetes.default.svc
  project: default
  source:
    path: charts/argo
    repoURL: https://github.com/argoproj/argo-helm.git
  syncPolicy:
    automated:
      selfHeal: true
```
I am just creating a new app via Argo CD with, not via CLI. So ArgoCD handles all helm installations. 
Via CLI is working fine even with your command. So I am not sure what I am doing wrong. 🤔 

````
 helm install my-argo argo/argo --version 0.16.10 -n argo-work
W0427 20:54:08.811837 2398436 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 20:54:08.877059 2398436 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 20:54:08.958523 2398436 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 20:54:09.032538 2398436 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0427 20:54:09.076939 2398436 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
NAME: my-argo
LAST DEPLOYED: Tue Apr 27 20:54:16 2021
NAMESPACE: argo-work
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
1. Get Argo Server external IP/domain by running:

kubectl --namespace argo-work get services -o wide | grep my-argo-server

2. Submit the hello-world workflow by running:

argo submit https://raw.githubusercontent.com/argoproj/argo/master/examples/hello-world.yaml --watch
❯ kubectl  get po -n argo-work
NAME                                           READY   STATUS              RESTARTS   AGE
my-argo-server-7c695959c6-frdff                0/1     ContainerCreating   0          16s
my-argo-workflow-controller-6cb5797dcd-q6d4j   0/1     ContainerCreating   0          16s

**Comment by oliverbaehler** [comment]
@ionutz89 Oh I see. You need to add minio as helm repository when deploying that way. See When you build it from the git repository, the dependency is missing and therefor it tries building it (And that's why you are getting that error message). But with the helm packages we release (and are installed via cli) the reference to the helm repository is hardcoded and therefor it's always able to fetch the chart.

What you need to do is adding the minio repository (based on [argo-cd repositories values](https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml#L571)) would look something like this:

```
...
repositories: |
  - type: helm
    url: https://helm.min.io/
    name: minio
....
```

Try that and it should work. Another way would be instead of referencing the Git Repository the Helm Repository in your deployed application.

File: charts/argo-workflows/templates/_helpers.tpl [hunk 24]
```diff
@@ -0,0 +1,96 @@
...
+    {{ default (include "argo-workflows.fullname" .) .Values.controller.serviceAccount }}
+{{- else -}}
+    {{ default "default" .Values.controller.serviceAccount }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the appropriate apiVersion for ingress
+*/}}
+{{- define "argo-cd.ingress.apiVersion" -}}
```
bmayfi3ld, line 88: Should this be 

```suggestion
{{- define "argo-workflows.ingress.apiVersion" -}}
```

I wasn't able to get it to work otherwise with below error.

`UPGRADE FAILED: template: argo-workflows/templates/server/server-ingress.yaml:5:15: executing \"argo-workflows/templates/server/server-ingress.yaml\" at <include \"argo-workflows.ingress.apiVersion\" .>: error calling include: template: no template \"argo-workflows.ingress.apiVersion\" associated with template \"gotpl\"\n",`

server-ingress.yaml references it as

chgl: That's embarassing, you are correct of course!

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.server.enabled -}}
+{{- if .Values.server.ingress.enabled -}}
+{{- $serviceName := include "argo-workflows.server.fullname" . -}}
+{{- $servicePort := .Values.server.servicePort -}}
+apiVersion: {{ include "argo-workflows.ingress.apiVersion" . }}
+metadata:
+  name: {{ template "argo-workflows.server.fullname" . }}
+  labels:
+    {{- include "argo-workflows.labels" $ | nindent 4 }}
```
bmayfi3ld, line 9:
```suggestion
   {{- include "argo-workflows.labels" (dict "context" . "component" .Values.server.name "name" .Values.server.name) | nindent 4 }}
```

Making this same as other references to this template. Was getting an error otherwise.
error calling include: template: argo-workflows/templates/_helpers.tpl:57:39: executing \"argo-`workflows.selectorLabels\" at <.context.Release.Name>: nil pointer evaluating interface {}.Release"`

bmayfi3ld (commented):
Didn't mean to make this a review, but similar error as in the template file.

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.server.enabled -}}
+{{- if .Values.server.ingress.enabled -}}
+{{- $serviceName := include "argo-workflows.server.fullname" . -}}
+{{- $servicePort := .Values.server.servicePort -}}
+apiVersion: {{ include "argo-workflows.ingress.apiVersion" . }}
+metadata:
```
bmayfi3ld, line 6:
```suggestion
kind: Ingress
metadata:
```

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,39 @@
...
+          {{- if $.Values.server.ingress.paths }}
+          {{- range $.Values.server.ingress.paths }}
+          - backend:
+              serviceName: {{ .serviceName }}
+              servicePort: {{ .servicePort }}
+          {{- end }}
+          {{- end }}
+          - backend:
+              serviceName: {{ $serviceName }}
+              servicePort: {{ $servicePort }}
```
bmayfi3ld, line 32:
```suggestion
    {{- $kubeVersion := .Capabilities.KubeVersion.GitVersion}}
    {{- range .Values.server.ingress.hosts }}
    - host: {{ . }}
      http:
        paths:
          {{- if $.Values.server.ingress.paths }}
          {{- range $.Values.server.ingress.paths }}
          {{- include "argo-workflows.ingress.service" (dict "kubeVersion" $kubeVersion "serviceName" .serviceName "servicePort" .servicePort) | nindent 10 }}
          {{- end }}
          {{- end }}
          {{- include "argo-workflows.ingress.service" (dict "kubeVersion" $kubeVersion "serviceName" $serviceName "servicePort" $servicePort) | nindent 10 }}
```

Also see new helper function. The new ingress api changed the fields layout. Since the api field in here was attempting to be backward compatible. I suggested a function to correctly build the backend field.

charts/argo-workflows/templates/_helpers.tpl
```diff
@@ -0,0 +1,96 @@
...
+*/}}
+{{- define "argo-workflows.ingress.apiVersion" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- print "extensions/v1beta1" -}}
+{{- else if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- print "networking.k8s.io/v1beta1" -}}
+{{- else -}}
+{{- print "networking.k8s.io/v1" -}}
+{{- end -}}
+{{- end -}}
```
bmayfi3ld, line 96:
```suggestion
{{- end -}}

{{/*
Return the appropriate service layout for ingress
*/}}
{{- define "argo-workflows.ingress.service" -}}
{{- if semverCompare "<1.19-0" .kubeVersion -}}
- backend:
    serviceName: {{ .serviceName }}
    servicePort: {{ .servicePort }}
{{- else -}}
- backend:
    service:
      name: {{ .serviceName }}
      port: 
        number: {{ .servicePort }}
  pathType: ImplementationSpecific
{{- end -}}
{{- end -}}
```

To compliment fixing errors in ingress.

bmayfi3ld:
Added several changes for the server's ingress resource. It didn't seem to be functional. But I had been using it in the argo chart.

chgl:
> Added several changes for the server's ingress resource. It didn't seem to be functional. But I had been using it in the argo chart.

@bmayfi3ld Thank you! I've now re-used the ingress definition from the argo-cd chart and also added a values file for chart-testing.

charts/argo-workflows/templates/server/server-deployment.yaml
```diff
@@ -0,0 +1,105 @@
...
+      {{- end }}
+      containers:
+        - name: argo-server
+          image: "{{ .Values.server.image.registry }}/{{ .Values.server.image.repository }}:{{ .Values.server.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.images.pullPolicy }}
+          securityContext:
+            {{- toYaml .Values.server.securityContext | nindent 12 }}
+          args:
+          - server
+          - --configmap={{ .Release.Name }}-{{ .Values.controller.name }}-configmap
```
bmayfi3ld, line 40:
```suggestion
          - --configmap={{ template "argo-workflows.controller.fullname" . }}-configmap
```

I believe this is the name of the configmap

oliverbaehler:
@chgl @bmayfi3ld We encounter currently a problem with said ingress condition in argo-cd #703. Currently my best idea would be the add a global value which allows to overwrite the kubeVersion. But i might have to sleep over it.

charts/argo-workflows/templates/controller/workflow-controller-servicemonitor.yaml
```diff
@@ -0,0 +1,29 @@
...
+  labels:
+    {{- include "argo-workflows.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
+    {{- with .Values.controller.serviceMonitor.additionalLabels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  endpoints:
+  {{- with .Values.controller.metricsConfig.enabled }}
+    - port: metrics
+      path: {{ . }}
```
tyrken, line 15: The path value should be from `.Values.controller.metricsConfig.path` - not the `true` that comes from the `with` expression.  Not sure if `with` really helps here or if it's better going back to the old fashioned `if`.

charts/argo-workflows/templates/controller/workflow-controller-servicemonitor.yaml
```diff
@@ -0,0 +1,29 @@
...
+      interval: 30s
+  {{- end }}
+  {{- with .Values.controller.telemetryConfig.enabled }}
+    - port: telemetry
+      path: {{ . }}
```
tyrken, line 20: Same `path` issue as above, but needs to be `.Values.controller.telemetryConfig.path`

tyrken:
Just found a couple of minor issues in the controller's ServiceMonitor, commented inline.  Otherwise looks good & works when I disabled them.  Would be nice to bump up the appVersion to `3.0.2` given it's been out a couple of weeks now.

Thanks to all for your efforts!

charts/argo-workflows/templates/controller/workflow-controller-servicemonitor.yaml
chgl: thanks! Agreed, here an `if` is definitely clearer.

chgl:
Thank you @tyrken! I've corrected it and also moved the previously known as `telemetryServicePort` inside the `telemetryConfig` as `telemetryConfig.servicePort` - same for `metricsConfig`.

I've additionally aligned the configuration of serviceAccounts with the argo-cd chart, ie: what used to be `server.createServiceAccount` is now `server.serviceAccount.create`, etc.  I've also found and corrected a few places in the RoleBindings where I hadn't used the `server/controllerServiceAccountName` templates.

abhinavkhanna-sf:
Thank you @chgl for the PR. I am actually tracking this PR everyday.
it still shows changes request by @oliverbaehler . Is it ready for review again?

abhinavkhanna-sf:
@oliverbaehler can you please review this PR?

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,88 @@
...
+          {{- if $extraPaths }}
+          {{- toYaml $extraPaths | nindent 10 }}
+          {{- end }}
+          {{- range $p := $paths }}
+          - path: {{ $p }}
+            {{- if eq (include "argo-workflows.ingress.apiVersion" $) "networking.k8s.io/v1" }}
+            pathType: Prefix
+            {{- end }}
+            backend:
+              {{- if eq (include "argo-workflows.ingress.apiVersion" $) "networking.k8s.io/v1" }}
```
oliverbaehler, line 69:
```suggestion
              {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,88 @@
...
+    {{- end -}}
+  {{- else }}
+    - http:
+        paths:
+          {{- if $extraPaths }}
+          {{- toYaml $extraPaths | nindent 10 }}
+          {{- end }}
+          {{- range $p := $paths }}
+          - path: {{ $p }}
+            {{- if eq (include "argo-workflows.ingress.apiVersion" $) "networking.k8s.io/v1" }}
```
oliverbaehler, line 65:
```suggestion
            {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,88 @@
...
+            pathType: Prefix
+            {{- end }}
+            backend:
+              {{- if eq (include "argo-workflows.ingress.apiVersion" $) "networking.k8s.io/v1" }}
```
oliverbaehler, line 42:
```suggestion
              {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,88 @@
...
+    {{- range $host := .Values.server.ingress.hosts }}
+    - host: {{ $host }}
+      http:
+        paths:
+          {{- if $extraPaths }}
+          {{- toYaml $extraPaths | nindent 10 }}
+          {{- end }}
+          {{- range $p := $paths }}
+          - path: {{ $p }}
+            {{- if eq (include "argo-workflows.ingress.apiVersion" $) "networking.k8s.io/v1" }}
```
oliverbaehler, line 38:
```suggestion
            {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```

charts/argo-workflows/templates/server/server-ingress.yaml
```diff
@@ -0,0 +1,88 @@
...
+  {{- end }}
+{{- end }}
+  name: {{ template "argo-workflows.server.fullname" . }}
+  labels:
+    {{- include "argo-workflows.labels" (dict "context" . "component" .Values.server.name "name" .Values.server.name) | nindent 4 }}
+    {{- if .Values.server.ingress.labels }}
+    {{- toYaml .Values.server.ingress.labels | nindent 4 }}
+    {{- end }}
+spec:
+  {{- if eq (include "argo-workflows.ingress.apiVersion" $) "networking.k8s.io/v1" }}
```
oliverbaehler, line 22:
```suggestion
  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```

charts/argo-workflows/templates/_helpers.tpl
```diff
@@ -0,0 +1,114 @@
...
+    servicePort: {{ .servicePort }}
+{{- else -}}
+- backend:
+    service:
+      name: {{ .serviceName }}
+      port:
+        number: {{ .servicePort }}
+  pathType: ImplementationSpecific
+{{- end -}}
+{{- end -}}
```
oliverbaehler, line 114: Remove this

oliverbaehler (changes requested):
@chgl Final Changes, after that I will be satisfied and further fixes changes should be done with additional prs, because we are kinda losing the overview here. Anyway, thanks for the work!

charts/argo-workflows/templates/server/server-ingress.yaml
chgl: Thanks! I added `/Ingress"` to make sure it checks for the availability of the Ingress version specifically - there was a comment here somewhere indicating the need for it.

abhinavkhanna-sf:
@oliverbaehler Can you please review this ?

oliverbaehler (approved):
lgtm

oliverbaehler:
@chgl Thanks for all your efforts! :heart: 
The Chart is now also available on Artifacthub: https://artifacthub.io/packages/helm/argo/argo-workflows

Please test & open new prs if any errors occur.

chgl:
Thanks again @oliverbaehler for your time (and patience) reviewing this PR!

PR #683, opened 2021-04-20
feat(argo-events): commit signing failure

kostas-theo:
This is a small feature to allow the passing in of Kubernetes Service Account (KSA) annotations into the `argo-events-sa` service account. Currently this is not possible using this chart. This is desired to be able to use the already created KSA with GKE workload identity, as well as any other use cases people may have using annotations.

Checklist:


charts/argo-events/templates/argo-events-sa.yaml
```diff
@@ -14,5 +18,8 @@ kind: ServiceAccount
 metadata:
   name: {{ $sa }}
   namespace: {{ $namespace }}
+  {{- if $annotations }}
+  annotations: {{ toYaml $annotations | nindent 4 }}
```
oliverbaehler, line 22:
```suggestion
  annotations: {{- toYaml $annotations | nindent 4 }}
```

oliverbaehler (changes requested):

charts/argo-events/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to install Argo-Events in k8s Cluster
 name: argo-events
-version: 1.2.4
+version: 1.3.4
```
oliverbaehler, line 4: That's not the proper version path accorging to [semver](https://semver.org/lang/de/). Changing to a new minor always resets the bgufix version. This rule applies always when the higher version number hierarchy is increased, the following version numbers. Because for example the bugfix version correspond the the patches icluding the new feature.

```suggestion
version: 1.3.0
```

oliverbaehler (changes requested):

charts/argo-events/Chart.yaml
kostas-theo: Hey @oliverbaehler - I was getting confused with all the git commit signing and looks like I messed something up, sorry about that. Dunno what's going on with the diff, but the current version in master branch for argo-events is `1.3.3`. As this is a bugfix I would have though the new version would be `1.3.4` - definitely not `1.3.0`.

kostas-theo:
new pr created with clean commit history for this work

PR #687, opened 2021-04-21
feat(argocd-applicationset): Lookup referenced Configmaps on Install

oliverbaehler:
Signed-off-by: Oliver Bähler <oliverbaehler@hotmail.com>

With this feature we ensure that configmaps that don't exist are not mounted and therefor improve user experience. Also we can more easily install the chart in linting scenarios and no longer need to set the variables to false.

Note on DCO:

Checklist:

### DIFF
--- a/charts/argocd-applicationset/Chart.yaml
+++ b/charts/argocd-applicationset/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argocd-applicationset
 description: A Helm chart for installing ArgoCD ApplicationSet
 type: application
-version: 0.1.3
+version: 0.2.0
 appVersion: "v0.1.0"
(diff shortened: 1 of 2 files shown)


dirtycajunrice:
lookup will break argocd even deploying argocd. it requires creds the workers dont have.

mkilchhofer:
> lookup will break argocd even deploying argocd. it requires creds the workers dont have.

I also think that using `lookup` in context of Argo is not a good idea. The `lookup` requires a correct install/upgrade (`helm upgrade <foo>` or `helm install <foo>`. ArgoCD only executes `helm template <foo>` and therefore lookup always returns and empty string.

oliverbaehler:
Fixed by not installing the Chart in the CI Run

PR #704, opened 2021-04-29
fix(argo-cd): Add option to override kubeVersion

### LINKED ISSUE #703: ArgoCD can't update itself to 3.2.2
**Describe the bug**
I have a self-managed ArgoCD installation and it can't update itself to 3.2.2. Currently running 3.1.2.

The sync fails saying "one or more synchronization tasks are not valid" and below "Ingress.networking.k8s.io "" not found"

**To Reproduce**
I don't know if this is some problem with [my `values`](https://github.com/valerauko/sut/blob/a51d4a3/apps/20-argo-cd.yaml) or the chart itself. I couldn't see any changes in #659 (that bumped 3.2.2) related to ingress.

**Expected behavior**
ArgoCD can update itself without issue.

**Screenshots**
image

**Additional context**
-

oliverbaehler:
Signed-off-by: Oliver Bähler <oliverbaehler@hotmail.com>

Note on DCO:

Checklist:

Resolves #703 but requires user interaction. Only other option would be reverting but i believe we will encounter this problem in the future anyway.


oliverbaehler:
@davidkarlsen if you could take a look at this and approve if it, it would help a lot =).

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -144,11 +144,19 @@ app.kubernetes.io/component: {{ .component }}
...
 {{- print "networking.k8s.io/v1beta1" -}}
 {{- else -}}
 {{- print "networking.k8s.io/v1" -}}
 {{- end -}}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{/*
+Return the appropriate apiVersion for ingress
```
mkilchhofer, line 157:
```suggestion
Return the target Kubernetes version
```

charts/argo-cd/README.md
```diff
@@ -34,6 +34,20 @@ Changes in the `CustomResourceDefinition` resources shall be fixed easily by cop
...
+If you are running a cluster version prior to `1.19` you can avoid this issue by directly installing chart version `3.3.0` and setting `global.kubeCapabilities` like:
+
+``` 
+global:
+  kubeCapabilities: "1.18.0"
+``` 
+
+Then you should no longer encounter this issue.
+
+
```
mkilchhofer, line 50: We should highlight that this only happens when we deploy the chart with Argo. For all other cases where the `argo-cd` chart is deployed by other infrastructure provisioning tools (eg. Terraform, Flux, scripts), this can be ignored.

mkilchhofer (commented):
I added a small comment. Additionally we maybe think about the variable name this PR introduces. What about reusing what the [bitnami guys](https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl#L8) use (`.Values.global.kubeVersion`)?

mkilchhofer (approved):
LGTM

lbpdt:
The latest `argo-workflows` chart seems to suffer from the same problem FYI.

PR #759, opened 2021-05-26
fix(argo-workflows): use template for all resource names

### LINKED ISSUE #521: Support as a subchart
When using argo as a dependency of another helm chart, the application name takes from the parent.
For example id Argo is installed as a subchart of mychart, argo-server is deployed as mychart-server. There are also references to argo-server (service account and deployment) so argo won't work in that case.

The name I can see is coming from the use in templates of the release name as in `{{ .Release.Name }}-{{ .Values.server.name}}`
Maybe it could be changed directly to  `argo-{{ .Values.server.name}}` rather than using the release name, which is not really adding anything crucial (we can install as many argos as we want in different namespaces or change server.name variable).

### LINKED ISSUE #762: Wrong name for Controller ClusterRole in Controller CRB for `argo-workflows`
Location: https://github.com/argoproj/argo-helm/blob/f6069848f53cddb33f5be6e04b1eaebe824a3a5f/charts/argo-workflows/templates/controller/workflow-controller-crb.yaml#L46

Does not match the name of the generated CR located here: https://github.com/argoproj/argo-helm/blob/f6069848f53cddb33f5be6e04b1eaebe824a3a5f/charts/argo-workflows/templates/controller/workflow-controller-cluster-roles.yaml#L138

causes the binding to mismatch , leading to 403 errors & crashloop on the controller when running a workflow

mkilchhofer:
/cc: @chgl 

This should resolve #521. It also resolves #762

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-workflows/templates/controller/workflow-aggregate-roles.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-aggregate-roles.yaml
@@ -5,7 +5,7 @@ metadata:
   annotations:
     helm.sh/hook: pre-install
     helm.sh/hook-delete-policy: before-hook-creation
-  name: argo-workflows-aggregate-to-view
+  name: {{ template "argo-workflows.fullname" . }}-view
   labels:
     rbac.authorization.k8s.io/aggregate-to-view: "true"
 rules:
@@ -33,7 +33,7 @@ metadata:
   annotations:
     helm.sh/hook: pre-install
     helm.sh/hook-delete-policy: before-hook-creation
-  name: argo-workflows-aggregate-to-edit
+  name: {{ template "argo-workflows.fullname" . }}-edit
   labels:
     rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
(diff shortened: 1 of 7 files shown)


charts/argo-workflows/templates/controller/workflow-aggregate-roles.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   annotations:
     helm.sh/hook: pre-install
     helm.sh/hook-delete-policy: before-hook-creation
-  name: argo-workflows-aggregate-to-view
+  name: {{ template "argo-workflows.fullname" . }}-view
```
vladlosev, line 8: These cluster roles are now specific to the release and as such don't have to be managed as hooks, so the hook annotations can be dropped.

vladlosev (approved):

charts/argo-workflows/templates/controller/workflow-aggregate-roles.yaml
mkilchhofer: you are right, I dropped these now.

vladlosev (approved):
LGTM

PR #760, opened 2021-05-26
feat(argo-cd): Add ability to override API versions

AtzeDeVries:
Signed-off-by: Atze de Vries <devries@zorgdomein.nl>

Two use cases.

* If you run `helm template` the helm `.Has.Capablities` is not validated to the Kubernetes API server by default, sometimes this is also not what you want. This can cause incompatible resources. This way you can render the templates for the correct version without contacting the Kubernetes API
* ArgoCD can render helm template but it does not validate against api server, so ArgoCD cannot manage it self if you use this Helm Chart. For example if you use ArgoCD 2.0.1 it uses helm 3.5.2, which has by default the k8s 1.19 api's which for ingress causes incompatibility if you run k8s 1.18 or lower.  (check https://github.com/argoproj/argo-cd/issues/3594)

Checklist:


AtzeDeVries:
This is a bit of controversial fix but it is a problem which really exists, although it depends a bit on the use case for this helm chart, see the use cases in the topic start.

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -144,7 +144,9 @@ app.kubernetes.io/component: {{ .component }}
 Return the appropriate apiVersion for ingress
 */}}
 {{- define "argo-cd.ingress.apiVersion" -}}
-{{- if semverCompare "<1.14-0" (include "argo-cd.kubeVersion" $) -}}
+{{- if .Values.APIOverrides -}}
```
oliverbaehler, line 147:
```suggestion
{{- if .Values.APIOverrides.enabled -}}
```

charts/argo-cd/templates/argocd-server/certificate.yaml
```diff
@@ -1,5 +1,7 @@
 {{- if .Values.server.certificate.enabled -}}
-{{- if .Capabilities.APIVersions.Has "cert-manager.io/v1" }}
+{{- if .Values.APIOverrides -}}
```
oliverbaehler, line 2:
```suggestion
{{- if .Values.APIOverrides.enabled -}}
```

oliverbaehler (changes requested):

charts/argo-cd/values.yaml
```diff
@@ -23,6 +23,14 @@ global:
...
   #   - git.myhostname
 
+# Override APIVersions
+# If you want to template helm charts but cannot access k8s API server
+# you can set api versions here
+APIOverrides:
+  enabled: false
+  certmanager: cert-manager.io/v1
+  ingress: networking.k8s.io/v1beta1
+
```
mkilchhofer, line 33: Why do we need an `enabled` flag here? It would be simpler to leave the defaults on `""` or `<nil>`/`~`. Otherwise the if/else if/else condition does not work properly anymore.

AtzeDeVries: yes might make sense, now you can set APIoverrides enabled and you then need to override them for both certmanager and ingress. I'll rebuild it based on the vars being emtpy

charts/argo-cd/templates/_helpers.tpl
AtzeDeVries: correct suggestion, but i've solved it differently in the end, see comment of mkilchhofer

charts/argo-cd/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.0.1
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.6.2
+version: 3.6.3
```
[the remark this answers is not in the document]
AtzeDeVries: ok

AtzeDeVries:
@oliverbaehler can you check if you change requests are still valid?

AtzeDeVries:
@mkilchhofer how can we move forward with this?

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -144,7 +144,9 @@ app.kubernetes.io/component: {{ .component }}
 Return the appropriate apiVersion for ingress
 */}}
 {{- define "argo-cd.ingress.apiVersion" -}}
-{{- if semverCompare "<1.14-0" (include "argo-cd.kubeVersion" $) -}}
+{{- if ne .Values.apiVersionOverrides.ingress "" -}}
```
mkilchhofer, line 147: This would be sufficient:
```suggestion
{{- if .Values.apiVersionOverrides.ingress -}}
```

charts/argo-cd/templates/argocd-server/certificate.yaml
```diff
@@ -1,5 +1,7 @@
 {{- if .Values.server.certificate.enabled -}}
-{{- if .Capabilities.APIVersions.Has "cert-manager.io/v1" }}
+{{-  if ne .Values.apiVersionOverrides.certmanager "" -}}
```
mkilchhofer, line 2:
```suggestion
{{-  if .Values.apiVersionOverrides.certmanager -}}
```

mkilchhofer (changes requested):
See the two comments.
Additional, please update the README with the new values. Then it it looks okay :-)

mkilchhofer (approved):
lgtm

AtzeDeVries (commented):
implemented / reworked all requested changes

mecampbellsoup (commented):
just curious - why wasn't this PR done "generically" such that users could override _any_ `apiVersion` for any resource they need?

PR #770, opened 2021-05-29
chore: Remove old workflow(s) in '.argo' directory

mkilchhofer:
I am not sure about the purpose of these two files, but:
- the container image `argoproj/argoscm:v2.0` is almost 4 years old and never got patched
  https://hub.docker.com/r/argoproj/argoscm/tags?page=1&ordering=last_updated
- the container image `hypnoglow/kubernetes-helm:v2.6.1` also is almost 4 years old
  https://hub.docker.com/layers/hypnoglow/kubernetes-helm/v2.6.1/images/sha256-4551fbee9375a6dacd8e0c02bf74a22fae7a4aaacdcaef09c8163ae137993d5e?context=explore

Since the workflow for linting AND releasing the charts now only using GitHub Actions, I think we can simply drop this.

WDYT @alexec and @alexmt?


oliverbaehler (approved):
lgtm

PR #775, opened 2021-05-31
fix(argo-rollouts): Allow metrics service without ServiceMonitor

PedroArvela:
With the upgrade to version 1.0.0 of the chart, a regression was introduced.

Before, even if you had not installed in your cluster the CoreOS Service Monitor, the chart would allow you to expose a metrics service for your metrics provider to read.

Now, if you do not set `controller.metrics.serviceMonitor.enabled` on your release, the service will not be created, breaking metric exports.

But, if you set `controller.metrics.serviceMonitor.enabled`, it will attempt to create a `ServiceMonitor.monitoring.coreos.com/v1` resource, which, if you do not have it installed in your cluster, breaks the release.

This makes it so the metrics Service is always deployed, allowing people who do not have the CoreOS ServiceMonitor to also make use of these metrics.

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "v1.0.1"
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 1.0.0
+version: 1.0.1
 icon: https://raw.githubusercontent.com/argoproj/argo/master/argo.png
 home: https://github.com/argoproj/argo-helm
 maintainers:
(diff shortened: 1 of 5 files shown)


mkilchhofer:
Hi @PedroArvela 
Thanks for pointing to this use case. I am sorry for introducing this regression.
What about adding a new variable for this? I suggest: `.Values. controller.metrics.enabled` as we use in the `argo-cd` chart:
- https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-server/metrics-service.yaml#L1
- https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-server/servicemonitor.yaml#L1

PedroArvela:
Hi @mkilchhofer 

Sounds like a good idea, I've just updated the PR to match the format in Argo CD :) 

I've tested it locally, and it is creating the service correctly, as expected.

Let me know if you need anything else.

mkilchhofer (approved):
lgtm

PR #782, opened 2021-06-03
docs(argo-workflows): Add missing value on values.yaml

yu-croco:
Signed-off-by: yu-croco <yuki.kita22@gmail.com>

In this PR https://github.com/argoproj/argo-helm/pull/769 , `fullnameOverride` was added but it was not defined on `values.yaml`, so I added it. 🙋 

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-workflows/values.yaml
+++ b/charts/argo-workflows/values.yaml
@@ -14,11 +14,11 @@ createAggregateRoles: true
 
 ## String to partially override "argo-workflows.fullname" template
 ##
-# nameOverride:
+nameOverride:
 
 ## String to fully override "argo-workflows.fullname" template
 ##
-# fullnameOverride:
+fullnameOverride:
 
 # Restrict Argo to only deploy into a single namespace by apply Roles and RoleBindings instead of the Cluster equivalents,
 # and start argo-cli with the --namespaced flag. Use it in clusters with strict access policy.
--- a/charts/argo-workflows/values.yaml
+++ b/charts/argo-workflows/values.yaml
@@ -12,9 +12,12 @@ init:
 
 createAggregateRoles: true
 
-# Override app name
-# Internally, this will be truncated at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
-fullnameOverride: ""
+## String to partially override "argo-workflows.fullname" template
+##
+# nameOverride:
+## String to fully override "argo-workflows.fullname" template
+##
+# fullnameOverride:
 
 # Restrict Argo to only deploy into a single namespace by apply Roles and RoleBindings instead of the Cluster equivalents,
 # and start argo-cli with the --namespaced flag. Use it in clusters with strict access policy.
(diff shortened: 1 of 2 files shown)


charts/argo-workflows/values.yaml
```diff
@@ -12,6 +12,10 @@ init:
 
 createAggregateRoles: true
 
+# Override app name
+# Internally, this will be truncated at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+fullnameOverride: ""
+
```
yu-croco, line 18: This is used below. 🙋 
https://github.com/argoproj/argo-helm/blob/master/charts/argo-workflows/templates/_helpers.tpl#L24-L39

mkilchhofer, line 18:
```suggestion
## String to partially override "argo-workflows.fullname" template
##
# nameOverride:

## String to fully override "argo-workflows.fullname" template
##
# fullnameOverride:
```

yu-croco: Thank you for reviewing! I agree with it.

> Also nameOverride was missing. WDYT?

I completely forgot it... 😇 You are right.

I will fix as you suggested.

yu-croco: I fixed on this commit 🙋 
https://github.com/argoproj/argo-helm/pull/782/commits/81e2e5d7708196997a5454394e0a86d11595c427

charts/argo-workflows/values.yaml
```diff
@@ -12,6 +12,13 @@ init:
 
 createAggregateRoles: true
 
+## String to partially override "argo-workflows.fullname" template
+##
+# nameOverride:
+## String to fully override "argo-workflows.fullname" template
```
mkilchhofer, line 18:
```suggestion

## String to fully override "argo-workflows.fullname" template
```

charts/argo-workflows/values.yaml
```diff
@@ -12,6 +12,14 @@ init:
 
 createAggregateRoles: true
 
+## String to partially override "argo-workflows.fullname" template
+##
+# nameOverride:
```
mkilchhofer, line 17: I think, we'd better use this without being a comment, so tools like ArgoCD detect this value in the UI :-) Sorry for that, as it was my suggestion to use it like this.
```suggestion
nameOverride:
```

charts/argo-workflows/values.yaml
```diff
@@ -12,6 +12,14 @@ init:
...
+
+## String to fully override "argo-workflows.fullname" template
+##
+# fullnameOverride:
```
mkilchhofer, line 21: I think, we'd better use this without being a comment, so tools like ArgoCD detect this value in the UI :-) Sorry for that, as it was my suggestion to use it like this.
```suggestion
fullnameOverride:
```

mkilchhofer (approved):

PR #783, opened 2021-06-03
feat(argocd-notifications): Add metrics service labels and annotations

sergeyshaykhullin:
Note on DCO:

Checklist:


charts/argocd-notifications/templates/service-metrics.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   name: {{ include "argocd-notifications.name" . }}-metrics
   labels:
     {{- include "argocd-notifications.metrics.labels" . | nindent 4 }}
+  annotations:
+    {{- include "argocd-notifications.metrics.annotations" . | nindent 4 }}
```
mkilchhofer, line 9: Hmm I don't think we should introduce more functions only used once. Please use:
```suggestion
  {{- with .Values.metrics.service.annotations  }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

charts/argocd-notifications/templates/service-metrics.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   name: {{ include "argocd-notifications.name" . }}-metrics
   labels:
     {{- include "argocd-notifications.metrics.labels" . | nindent 4 }}
```
mkilchhofer, line 7: Also here, do not extend the template function as these labels are also set on the ServiceMonitor object, not only on the `Service` as a user would suppose.
```suggestion
    {{- include "argocd-notifications.metrics.labels" . | nindent 4 }}
    {{- with .Values.metrics.service.labels }}
      {{- toYaml . | nindent 4 }}
    {{- end }}
```

mkilchhofer (changes requested):
Please remove the file `charts/argocd-notifications/templated.yaml` and restore the `_helpers.tpl`.

mkilchhofer (approved):

PR #784, opened 2021-06-03
feat: add encrypt values for secret section

akastav:
Problem:
we prepare passwords via terraform and use bcrypt to encrypt the password (example of the encrypted password is "$2a$10$H1a30nMr9v2QE2nkyz0BoOD2J0I6FQFMtHS0csEg12RBWzfRuuoE6". When ArgoCD creates a secret and tries to encrypt in base64, characters like $ are interpreted by the shell as variables and breaks the password.

Solutions:
This will help to pass the prepared secrets encrypted via bcrypt to the chart parameters. If ".Values.configs.secret.encrypt" = false, used stringData without base64 encrypt procedure.

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/templates/argocd-configs/argocd-secret.yaml
+++ b/charts/argo-cd/templates/argocd-configs/argocd-secret.yaml
@@ -20,30 +20,34 @@ data:
 stringData:
 {{- end }}
   {{- if .Values.configs.secret.githubSecret }}
-  webhook.github.secret: {{ .Values.configs.secret.githubSecret | b64enc }}
+  webhook.github.secret: {{- if .Values.configs.secret.encrypt }}{{ .Values.configs.secret.githubSecret | b64enc }} {{ else }} {{ .Values.configs.secret.githubSecret }} {{ end }}
   {{- end }}
   {{- if .Values.configs.secret.gitlabSecret }}
-  webhook.gitlab.secret: {{ .Values.configs.secret.gitlabSecret | b64enc }}
+  webhook.gitlab.secret: {{- if .Values.configs.secret.encrypt }}{{ .Values.configs.secret.gitlabSecret | b64enc }} {{ else }} {{ .Values.configs.secret.gitlabSecret }} {{ end }}
   {{- end }}
   {{- if .Values.configs.secret.bitbucketServerSecret }}
-  webhook.bitbucketserver.secret: {{ .Values.configs.secret.bitbucketServerSecret | b64enc }}
+  webhook.bitbucketserver.secret: {{- if .Values.configs.secret.encrypt }}{{ .Values.configs.secret.bitbucketServerSecret | b64enc }} {{ else }} {{ .Values.configs.secret.bitbucketServerSecret }} {{ end }}
   {{- end }}
   {{- if .Values.configs.secret.bitbucketUUID }}
-  webhook.bitbucket.uuid: {{ .Values.configs.secret.bitbucketUUID | b64enc }}
+  webhook.bitbucket.uuid: {{- if .Values.configs.secret.encrypt }}{{ .Values.configs.secret.bitbucketUUID | b64enc }} {{ else }} {{ .Values.configs.secret.bitbucketUUID }} {{ end }}
   {{- end }}
   {{- if .Values.configs.secret.gogsSecret }}
-  webhook.gogs.secret: {{ .Values.configs.secret.gogsSecret | b64enc }}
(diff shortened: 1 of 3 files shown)


charts/argo-cd/templates/argocd-configs/argocd-secret.yaml
```diff
@@ -14,7 +14,11 @@ metadata:
 type: Opaque
 {{- if or .Values.configs.secret.githubSecret (or .Values.configs.secret.gitlabSecret .Values.configs.secret.bitbucketUUID .Values.configs.secret.bitbucketServerSecret .Values.configs.secret.gogsSecret .Values.configs.secret.argocdServerAdminPassword .Values.configs.secret.argocdServerTlsConfig .Values.configs.secret.extra) }}
 # Setting a blank data again will wipe admin password/key/cert
+{{- if .Values.configs.secret.encrypt }}
 data:
+{{- else }}
+stringData:
```
mkilchhofer, line 20: This will not work, look at some lines below. We use `| b64enc` everywhere.

mkilchhofer:
Hi @akastav 
How do you pass that TF encrypted bcrypt value? Do you use the helm provider? We also use TF to deploy this chart and do not have this issue:
```terraform
data "template_file" "oidc_config_template" {
  template = file("${path.module}/helm_values.yaml")
  vars = {
    // (..)
    argocd_password              = bcrypt(var.argocd_password)
    // (..)
  }
}

resource "helm_release" "argocd" {
  name       = var.argocd_helm_release_name
  repository = var.argocd_helm_repository
  chart      = var.argocd_helm_chart_name
  version    = var.argocd_helm_chart_version
  namespace  = var.argocd_namespace

  values = [data.template_file.oidc_config_template.rendered]

  depends_on = [kubernetes_namespace.argocd]

}
```

akastav:
Hi @mkilchhofer! 
We don't use helm provider, in our TF project Swiss-army-kube we create random pass https://github.com/provectus/sak-argocd/blob/master/main.tf#L129 and store it in system manager, param store.  For decrypt pass, we used https://github.com/provectus/sak-argocd/blob/master/main.tf#L262 python code as configManagementPlugins. After all, we create ArgoCD spec and push it to git.  Some kind of self-managed ArgoCD

charts/argo-cd/templates/argocd-configs/argocd-secret.yaml
akastav: Maybe use something like
`admin.password: {{- if .Values.configs.secret.encrypt }}{{ .Values.configs.secret.argocdServerAdminPassword | b64enc }} {{else}} {{ .Values.configs.secret.argocdServerAdminPassword }}`

mkilchhofer:
> Hi @mkilchhofer!
> We don't use helm provider, in our TF project Swiss-army-kube we create random pass https://github.com/provectus/sak-argocd/blob/master/main.tf#L129 and store it in system manager, param store. For decrypt pass, we used https://github.com/provectus/sak-argocd/blob/master/main.tf#L262 python code as configManagementPlugins. After all, we create ArgoCD spec and push it to git. Some kind of self-managed ArgoCD

Oh I see. This have to do with `values` vs. `parameters`:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: argocd-testing
spec:
  destination:
    namespace: testing
    server: https://kubernetes.default.svc
  project: default
  source:
    chart: argo-cd
    helm:
      parameters:
      - name: configs.secret.argocdServerAdminPasswordDD
        value: $2a$10$H1a30nMr9v2QE2nkyz0BoOD2J0I6FQFMtHS0csEg12RBWzfRuuoE6
      values: |
        configs:
          secret:
            argocdServerAdminPassword: $2a$10$H1a30nMr9v2QE2nkyz0BoOD2J0I6FQFMtHS0csEg12RBWzfRuuoE6
```
It works in `values` but not in `parameters` as parameters supports variable substitution.

But I think your solution would not help. This is a feature of ArgoCD which unfortunately introduces this.
What about escaping the dollar sign with two or using the `values` attribute?

mkilchhofer:
Ah! This indeed would help. Base64 cannot contain a `$`

akastav:
Thanks for the hint about values - I'll definitely try it

mkilchhofer:
> Thanks for the hint about values - I'll definitely try it
> 
> @mkilchhofer Thx it works for us! This way much easy and will not allow you to compromise other accidentally unencrypted passwords. You can close this PR, but add this hint to the FAQ ;)

👍 I created an issue so we do not forget this: https://github.com/argoproj/argo-helm/issues/785

Issue #753, opened 2021-05-25
[argo-workflows] Helm .Capabilities not representing k8s cluster


marekadare:
**Describe the bug**
In one of the latest changes to argo-workflows argo-workflows/templates/_helpers.tpl was added. The helper argo-workflows.ingress.apiVersion is using .Capabilities.KubeVersion.Version but that is not representing the current k8s cluster version.

If you use argocd to deploy your Helm Chart the .Capabilities.KubeVersion.Version is v1.20.0 but the actual k8s cluster version is v1.18.0

This leads to an unusable Helm Chart, because I don't know any way to configure that version.
Updating the k8s cluster is not an option at the moment due to production usage.

**To Reproduce**
1. Install v1.18.0 Kubernetes cluster
1. Install argocd
2. Create new app for argo-workflows in argocd
3. Sync

**Expected behavior**
I don't know how but argocd or the argoworkflow Helm Chart helpers should check the k8s-cluster version beforehand.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/16648900/119487140-4b6b6300-bd59-11eb-911a-1b4593f20443.png)

kulinskyvs:
Looks, I have similar behavior with Helmfile. 

```yaml
server:
  ingress:
    enabled: true
    hosts:
      - myhost.mydomain.com
```

I have K8s cluster version 1.19 and with the values.yaml listed above I'm facing the following issue

```
STDERR:
  Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "service" in io.k8s.api.networking.v1beta1.IngressBackend

COMBINED OUTPUT:
  Release "argo-workflows" does not exist. Installing it now.
  Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "service" in io.k8s.api.networking.v1beta1.IngressBackend
```

mkilchhofer:
We recently added a new value to workaround this in `argo-cd`: https://github.com/argoproj/argo-helm/pull/704

Related issues:
- `argo-cd` chart: #703 
- argo-cd bug: https://github.com/argoproj/argo-cd/issues/3594

mkilchhofer:
https://github.com/argoproj/argo-cd/issues/5833 is implemented now. let's see if this helps as soon as this fix got released.

[...]

PR #795, opened 2021-06-07
fix(argo-workflows): use template for ingress .Capabilities

### LINKED ISSUE #794: argo-workflow: Ingress templating is misbehaving for <1.19 clusters
**Describe the bug**
Since https://github.com/argoproj/argo-helm/pull/747, the `Ingress` templating will not work properly on <1.19 K8s clusters (tested on 1.18.8). The ingress manifest is generated with `networking.k8s.io/v1` API but it's not available until 1.19 for `Ingress`. I'm not sure about older clusters than 1.18, the problem seems to be that when `networking.k8s.io/v1` is available on the cluster, either for `Ingress` or for anything else, the template will generate a `networking.k8s.io/v1` Ingress.

**To Reproduce**
1. Install the chart on a <1.19 cluster
2. The install will fail with a validation error

**Expected behavior**
The `Ingress` manifest should be generated for `networking.k8s.io/v1beta1` API.

Byh0ki:
This PR fixes #794.

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.2.4
+version: 0.2.5
 appVersion: "v3.0.2"
 icon: https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -94,11 +94,18 @@ Create the name of the controller service account to use
 Return the appropriate apiVersion for ingress
 */}}
 {{- define "argo-workflows.ingress.apiVersion" -}}
-{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.Version -}}
+{{- if semverCompare "<1.14-0" (include "argo-workflows.kubeVersion" $) -}}
 {{- print "extensions/v1beta1" -}}
-{{- else if semverCompare "<1.19-0" .Capabilities.KubeVersion.Version -}}
+{{- else if semverCompare "<1.19-0" (include "argo-workflows.kubeVersion" $) -}}
 {{- print "networking.k8s.io/v1beta1" -}}
 {{- else -}}
 {{- print "networking.k8s.io/v1" -}}
 {{- end -}}
+{{- end -}}
+
+{{/*
+Return the target Kubernetes version
+*/}}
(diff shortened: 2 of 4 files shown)


mkilchhofer (commented):
This will not fix the behaviour described in #794 but it consolidates multiple evaluations to a central point to implement a workaround for the real issue. Real issue is that when deploying this chart through a tool like ArgoCD (and I think also Pulumi), helm is only used for templating and not for the real deployment.

When only using "helm template", helm does not contact the API server by design. There are fixes in helm 3.6.0 and argo-cd PR https://github.com/argoproj/argo-cd/pull/6359. Please see duplicated issue: https://github.com/argoproj/argo-helm/issues/753

mkilchhofer:
Would you mind to implement this here also?:
- https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml#L6
- https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/_helpers.tpl#L143-L161

mkilchhofer:
> It's fixing the issue when using a simple `helm install`. I'm not currently using argo-cd to deploy this so I can't test it for this use case. That's why I missed #753, I thought that it was specific to argo-cd and not exactly the same use case, ence my issue 😅

Hmm, `helm install` did fail with the previous installation. Interesting! Maybe this `Capabilities.APIVersions.Has` does not work properly in helm v3.
Can you test if it still works with my addition? If so, I will approve and merge once you give green light from your side :-)

Byh0ki:
Hello,

> Hmm, helm install did fail with the previous implementation? Interesting! Maybe this Capabilities.APIVersions.Has does not work properly in helm v3.

Yes, as stated in my original issue (#794), the issue comes from the fact that the APIVersion `networking.k8s.io/v1` is available on 1.18 cluster for the `NetworkPolicies` resource. So the test:
```
if .Capabilities.APIVersions.Has "networking.k8s.io/v1"
```
will always evaluate at `true` on a 1.18 cluster, even if the Ingress is only available as a `networking.k8s.io/v1beta` resource. That's why my original fix was enough for me, it uses the same template as here https://github.com/argoproj/argo-helm/blob/87af20fd06f87b8a05b14ad33738244067102d21/charts/argo-workflows/templates/server/server-ingress.yaml#L6 to get the correct APIVersion. The argo-cd chart is already using the same template. https://github.com/argoproj/argo-helm/blob/87af20fd06f87b8a05b14ad33738244067102d21/charts/argo-cd/templates/argocd-server/ingress.yaml#L22

> Can you test if it still works with my addition? If so, I will approve and merge once you give green light from your side :-)

Will test today and report back here :)

Byh0ki:
`helm install` works as intended on 1.18 k8s cluster with the changes in this PR :)

mkilchhofer (approved):

PR #798, opened 2021-06-08
feat(argo-rollouts): add dashboard

huikang:
credit: the implemention is inspired by the work from Andrii Perenesenko <andrii.perenesenko@gmail.com> (https://github.com/argoproj/argo-rollouts/pull/1240).

Signed-off-by: Hui Kang <hui.kang@salesforce.com>

Note on DCO:

Checklist:

* [ ] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#versioning)
* [ ] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#changelog).
* [ ] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/tree/master/community#contributing-to-argo).


mkilchhofer:
Relates to https://github.com/argoproj/argo-rollouts/pull/1240

mkilchhofer (changes requested):
Let's use shorter filenames and create a subdirectory:
- `templates/dashboard/deployment.yaml`
- `templates/dashboard/service.yaml`

I think we also need a dedicated serviceAccount/RBAC for the dashboard?
```yaml
# Source: https://raw.githubusercontent.com/argoproj/argo-rollouts/d736636ff184f7c6f302c832351711a04565fbe2/manifests/dashboard-install/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- dashboard-clusterrolebinding.yaml # <--
- dashboard-clusterrole.yaml        # <--
- dashboard-deployment.yaml
- dashboard-service.yaml
- dashboard-serviceaccount.yaml     # <--
images:
- name: quay.io/argoproj/kubectl-argo-rollouts
  newTag: latest
```

huikang:
> Let's use shorter filenames and create a subdirectory:
> 
>     * `templates/dashboard/deployment.yaml`
> 
>     * `templates/dashboard/service.yaml`
> 

Agree. Will create a sub-directory and move the file there.

> 
> I think we also need a dedicated serviceAccount/RBAC for the dashboard?

My only concern about the dedicated sa is that we may have to create another set of clusterrole, rolebinding, etc, which seems an overkill. How about letting the same sa and its role to be shared by the controller and the dashboard for now. Later if there are user requests, we can create a separate sa. What do you think?

> 
> ```yaml
> # Source: https://raw.githubusercontent.com/argoproj/argo-rollouts/d736636ff184f7c6f302c832351711a04565fbe2/manifests/dashboard-install/kustomization.yaml
> apiVersion: kustomize.config.k8s.io/v1beta1
> kind: Kustomization
> 
> resources:
> - dashboard-clusterrolebinding.yaml # <--
> - dashboard-clusterrole.yaml        # <--
> - dashboard-deployment.yaml
> - dashboard-service.yaml
> - dashboard-serviceaccount.yaml     # <--
> images:
> - name: quay.io/argoproj/kubectl-argo-rollouts
>   newTag: latest
> ```

mkilchhofer:
> My only concern about the dedicated sa is that we may have to create another set of clusterrole, rolebinding, etc, which seems an overkill. How about letting the same sa and its role to be shared by the controller and the dashboard for now. Later if there are user requests, we can create a separate sa. What do you think?

Hmm maybe we should wait until the PR (https://github.com/argoproj/argo-rollouts/pull/1240) is merged. Then I'd implement the same resources here.

huikang:
> Hmm maybe we should wait until the PR ([argoproj/argo-rollouts#1240](https://github.com/argoproj/argo-rollouts/pull/1240)) is merged. Then I'd implement the same resources here.

Yeah, sure. I will update the PR after the referenced one merged.

charts/argo-rollouts/templates/_helpers.tpl
```diff
@@ -52,6 +52,14 @@ app.kubernetes.io/name: {{ include "argo-rollouts.name" . }}
...
 {{- end }}
 
+{{/*
+Selector labels: dashboard
+*/}}
+{{- define "argo-rollouts.selectorLabels-dashboard" -}}
+app.kubernetes.io/name: {{ include "argo-rollouts.name" . }}-dashboard
+app.kubernetes.io/instance: {{ .Release.Name }}-dashboard
+{{- end }}
+
```
mkilchhofer, line 62: I think we do not need a separate helper for this as we need this only once?

I'd take:
```yaml
spec:
  selector:
    matchLabels:
      {{- include "argo-rollouts.selectorLabels" . | nindent 6 }}
      app.kubernetes.io/component: {{ .Values.dashboard.component }}
```
WDYT?

mkilchhofer (changes requested):
See comments

charts/argo-rollouts/templates/_helpers.tpl
huikang: Hi, @mkilchhofer , thanks for your suggestions; dashboard service uses the `argo-rollouts.selectorLabels-dashboard` as well. When I made the following

```yaml
{{- if .Values.dashboard.enabled -}}
apiVersion: v1
kind: Service
metadata:
  name: {{ include "argo-rollouts.fullname" . }}-dashboard
  labels:
    app.kubernetes.io/component: {{ .Values.dashboard.component }}
    {{- include "argo-rollouts.labels" . | nindent 4 }}
  {{- with .Values.serviceAnnotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
spec:
  ports:
  - name: dashboard
    protocol: TCP
    port: 3100
    targetPort: 3100
  selector:
    {{- include "argo-rollouts.selectorLabels" . | nindent 6 }}                         # <--------- Use the shared label
    app.kubernetes.io/component: {{ .Values.dashboard.component }}         # <--------- Additional selector label to distinguish between dashboard and rollout deployment
{{- end }}
```

the above template gives me the error `Error: YAML parse error on argo-rollouts/templates/dashboard/service.yaml: error converting YAML to JSON: yaml: line 21: did not find expected key`. Any idea to fix?

mkilchhofer: Because the selector on the Service is only nindent 4 :-)

huikang: Thanks! Will send an updated PR.

charts/argo-rollouts/templates/argo-rollouts-deployment.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
 spec:
   selector:
     matchLabels:
+      app.kubernetes.io/component: {{ .Values.controller.component }}
```
mkilchhofer, line 11: This one here is a breaking change and requires the `--force`  parameter when using `helm upgrade`. I am unsure how we solve this.
Options:
- bump major chart version
- place a upgrade note inside the README
- ...?

Any ideas? @oliverbaehler , @davidkarlsen ?

oliverbaehler: Bump Major and add a Upgrading Section to the README.md (like https://github.com/argoproj/argo-helm/tree/master/charts/argo-cd#upgrading)

huikang: @mkilchhofer , the major version is bumped with added upgrading note. Please take a look. Thanks.

charts/argo-rollouts/values.yaml
```diff
@@ -94,3 +94,29 @@ podLabels: {}
...
+  podSecurityContext: {}
+  #  Enable runAsNonRoot after image version v1.0.1
+  #  runAsNonRoot: true
+  containerSecurityContext: {}
+  serviceAccount:
+    create: true
+    annotations: {}
+    # The name of the service account to use.
+    # If not set and create is true, a name is generated using the fullname template
+    name: ""
```
mkilchhofer, line 122: Could you also place documentation for these values into the README?

huikang: Done

charts/argo-rollouts/values.yaml
```diff
@@ -25,7 +25,7 @@ controller:
   #    memory: 64Mi
 
   metrics:
-    enabled: false
+    enabled: true
```
mkilchhofer, line 28: Is that change required? If so, please update the README :-)

charts/argo-rollouts/values.yaml
mkilchhofer: Perfect. Thank you so much! 🙏

charts/argo-rollouts/values.yaml
huikang: @mkilchhofer , it was enabled for test the PR. I reverted it back to `false` in the update PR. Thanks.

charts/argo-rollouts/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "v1.0.2"
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 1.0.4
+version: 2.0.0
```
mkilchhofer, line 5: Please update changelog on L14-15 with something like
```diff
   artifacthub.io/changes: |
-     - "[Fixed]: Add missing liveness and readiness probes"
-     - "[Changed]: Fix icon url"
+     - "[Added]: Dashboard manifests"
```

mkilchhofer (changes requested):
Last small nits

charts/argo-rollouts/Chart.yaml
huikang: Hi, @mkilchhofer , just want to confirm with you that the following lines need to be removed

```
-     - "[Fixed]: Add missing liveness and readiness probes"
-     - "[Changed]: Fix icon url"
```

mkilchhofer: Yup, they need to be removed.

huikang: Got it. Thanks.

mkilchhofer:
I added a commit which should turn on the dashboard deployment during chart testing. So we have some small automated testing on this area as well.

mkilchhofer (approved):
lgtm

PR #807, opened 2021-06-15
feat(argo-workflows): Multiple workflow namespaces

reinvantveer:
Following question https://cloud-native.slack.com/archives/C02097VUKUM/p1623678731036500 (as yet unresolved). I spent some time on this earlier this year so my commits were a bit stale (it was still in the good old `argo` chart days, before the rename to `argo-workflows` chart) so I had some catching up to do.

This PR intends to solve the following use case. In cases where one would want to run workflows in multiple namespaces (say, one for each customer) a set of RBAC resources is required in each namespace for the workflows to execute. Some workflows may be able to run, but they will run into `wait` container trouble with messages comparable to
```
User "system:serviceaccount:some-customer-workflows:default" cannot get resource "pods" in API group "" in the namespace "some-customer-workflows"
```

By allowing a "slice" (array) of workflow.namespaces in values.yaml, the proposed changes can iterate over the supplied namespaces and create RBAC resources accordingly.

Checklist:


reinvantveer:
Testing in separate namespaces: 
image
image

PR #828, opened 2021-07-10
fix(argo-cd): Add ServiceAccount for redis

### LINKED ISSUE #809: Can not create a service account for redis in argo-cd helm chart
**Describe the bug**
If i want to setup a service account for redis i have to select the redis-ha option. In a normal installation of argo-cd a service account for redis is automatically created. The default behavior for the helm chart should be the same.

**To Reproduce**
1. Install argo-cd like this: https://argo-cd.readthedocs.io/en/stable/getting_started/#1-install-argo-cd
2. Service account for redis is created
3. Install argo-cd via helm chart
4. No option to create a service account

**Expected behavior**
It should be possible to create a service account (role and role binding) via values.yaml the same way as for argocd-repo-server.

mkilchhofer:
Resolves #809

While here, I noticed that there might exist clashing ServiceAccount resources when the default names of `controller`, `dex`, `server` and `repoServer` are cleared. This is an edge case but its fixed now :-)

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.0.4
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.8.0
+version: 3.8.1
 home: https://github.com/argoproj/argo-helm
 icon: https://raw.githubusercontent.com/argoproj/argo/master/docs/assets/argo.png
 keywords:
@@ -21,4 +21,4 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Ability to override API versions"
+    - "[Fixed]: Cannot create a service account for redis"
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -89,6 +89,17 @@ Create the name of the dex service account to use
 {{- end -}}
 
 {{/*
+Create the name of the redis service account to use
+*/}}
+{{- define "argo-cd.redisServiceAccountName" -}}
+{{- if .Values.redis.serviceAccount.create -}}
+    {{ default (include "argo-cd.redis.fullname" .) .Values.redis.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.redis.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Create the name of the ArgoCD server service account to use
 */}}
 {{- define "argo-cd.serverServiceAccountName" -}}
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -71,7 +71,7 @@ Create the name of the controller service account to use
 */}}
 {{- define "argo-cd.controllerServiceAccountName" -}}
 {{- if .Values.controller.serviceAccount.create -}}
-    {{ default (include "argo-cd.fullname" .) .Values.controller.serviceAccount.name }}
+    {{ default (include "argo-cd.controller.fullname" .) .Values.controller.serviceAccount.name }}
 {{- else -}}
     {{ default "default" .Values.controller.serviceAccount.name }}
 {{- end -}}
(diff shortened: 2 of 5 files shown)


mkilchhofer:
@alexec Does @mbevc1 have correct permissions? He's in the CODEOWNERS file but Github does not pick him up as a reviewer.

mbevc1:
:thinking:

mbevc1:
Seems I might need push access to the repo :thinking: 
https://github.com/isaacs/github/issues/989#issuecomment-320475904

davidkarlsen (approved):

mkilchhofer:
> Seems I might need push access to the repo 🤔
> [isaacs/github#989 (comment)](https://github.com/isaacs/github/issues/989#issuecomment-320475904)

Created an issue for this (#829) so we can move on with this PR here and not forget to grant access for you.

mbevc1:
Cheers @mkilchhofer :+1:

chillum:
the changing of reposerver's and server's serviceaccount names is a breaking change (causing some existing configurations stopping working)

it would be nice if this was mentioned in the changelog.

chillum:
also please note that this change violated this checkbox from the checklist:
> Any new values are backwards compatible and/or have sensible default.

a sudden renaming of service accounts is not a backwards-compatible change, it literally breaks existing configurations that depend on service account's names (like GKE clusters with WorkloadIdentity).

mkilchhofer:
Hi @chillum,

Thanks for your attention. But these template functions are currently used as a default/fallback if someone override values to `nil` or `""` (empty string):

```yaml
controller:
  # (..)
  serviceAccount:
    create: true
    name: argocd-application-controller

dex:
  serviceAccount:
    create: true
    name: argocd-dex-server

server:
  serviceAccount:
    create: true
    name: argocd-server

repoServer:
  # (..)
  serviceAccount:
    create: false
    #  name: argocd-repo-server
```
Okay, I missed that the `repoServer`'s `name` parameter was only a comment. But for every other component, this template function does not change something as there are values set for the names.
Additional the repoServer normally does not need a serviceAccount as it does not talk to the K8s APIserver.

WDYT? Shall we file a PR to add some hints to the README or does my explanation make sense?

chillum:
so, you changed the default reposerver's serviceaccount name from `argo-cd-argocd` to `argocd-repo-server` and don't think it deserves a mention in the changelog?

this change does break existing configurations, I've been debugging one today.

mkilchhofer:
This change got at least 4 eyes before it got merged.

Please be constructive with your critique
--> what do you suggest to do now? The change did already happen.

Revert? Additional PR? Update the Changelog in Artifacthub?

chillum:
if in previous versions you'd set just:
```yaml
repoServer:
  serviceAccount:
    create: true
```

the user would end up with following service accounts:
* argo-cd-argocd
* argocd-application-controller
* argocd-dex-server
* argocd-server

the 3.8.1 makes a different result:
* argocd-repo-server
* argocd-application-controller
* argocd-dex-server
* argocd-server

yeah, now it looks cleaner, but this is a breaking change. so I'd say the documentation on this had changed is a must. it's been only several hours since the release and first users already found this problem and I'd expect more to come.

revert? no for sure. we could rename the fallback `argocd-repo-server` account back to `argo-cd-argocd` (but keep the new name as comment), but it would introduce complexity. I'd say this depends whether we prefer sane defaults or backward compatibility, but both are important. frankly not sure how much users would be affected though, so I think I'd keep the new name. but documented it.

mkilchhofer:
Added a draft PR to address this: https://github.com/argoproj/argo-helm/pull/830 Please let me know if this is sufficient for you :-)

chillum:
yeah, that looks great, many thanks.
one more idea, if the point of 3.8.1 was ServiceAccount for Redis.

maybe we shouldn't set `redis.serviceAccount.name = ""` in the defaults (and uncomment the repo-server stuff to do this change even more prominent and obvious). they both fallback to defaults, but setting them explicitly in the defaults file (like the chart does with other service accounts) provides better observability: one can look at default `values.yaml` and read the service accounts' name just from there. or copy and edit.

mkilchhofer:
Okay, but this would introduce the next breaking change 🙈 We'd better wait for the next breaking change and let the chart generate all the ServiceAccount names by default. WDYT?

I already think about another breaking change: Implement what was promised:
>### 2.10.x to 2.11.0
>
>The application controller is now available as a `StatefulSet` when the `controller.enableStatefulSet` flag is set to true. >Depending on your Helm deployment this may be a downtime or breaking change if enabled when using HA and will become the default in 3.x.

chillum:
oh, yes, looked up the [redis naming template](https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/_helpers.tpl#L44) and it's indeed complicated, so my proposal will be another breaking change. =)

PR #833, opened 2021-07-13
fix(argo-workflows): Fix controller Role/ClusterRole to work with InstanceID

kobejn-jb:
Signed-off-by: kobejn-jb jakubbielawski89@gmail.com

This PR comes with the following changes:

    It allows for the controller to work with InstanceID while using leases
    It fixes a controller Role/ClusterRole, now it allows to read and update leases with names ending with -<your instance ID> when controller is set to use InstanceID

This PR intends to solve the following use case. In case when controller is set to use InstanceID controller role does not allow to get/watch/update/path/delete leases with names different from "workflow-controller-lease" and "workflow-controller-lease" but controller tries to use lease named "workflow-controller-lease-<provided InstanceID>". Now trying to use controller with InstanceID it results with error:

`E0713 13:43:58.014735       1 leaderelection.go:325] error retrieving resource lock argo-test/workflow-controller-argo-controller: leases.coordination.k8s.io "workflow-controller-argo-controller" is forbidden: User "system:serviceaccount:argo-test:argo" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "argo-test"`

This PR changes Role/ClusterRole template, when InstanceID is enabled names of leases listed in permissions in Roles will contain that name at the end

Checklist:

* [X] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#changelog).

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.2.7
+version: 0.2.8
 appVersion: "v3.0.7"
 icon: https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
@@ -15,4 +15,4 @@ maintainers:
   - name: benjaminws
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Initialize Changelog"
+    - "[Fixed]: Controller Role/ClusterRole to work with InstanceID"
--- a/charts/argo-workflows/templates/controller/workflow-controller-cluster-roles.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-cluster-roles.yaml
@@ -121,8 +121,18 @@ rules:
   resources:
   - leases
   resourceNames:
+  {{- if .Values.controller.instanceID.enabled }}
+    {{- if .Values.controller.instanceID.useReleaseName }}
+  - workflow-controller-{{ .Release.Name }}
+  - workflow-controller-lease-{{ .Release.Name }}
+    {{- else }}
+  - workflow-controller-{{ .Values.controller.instanceID.explicitID }}
+  - workflow-controller-lease-{{ .Values.controller.instanceID.explicitID }}
+    {{- end }}
+  {{- else }}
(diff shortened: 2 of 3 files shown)


charts/argo-workflows/Chart.yaml
```diff
@@ -16,3 +16,4 @@ maintainers:
 annotations:
   artifacthub.io/changes: |
     - "[Added]: Initialize Changelog"
+    - "[Fixed]: Controller ClusterRole to work with InstanceID"
```
vladlosev, line 19: As I understand from https://artifacthub.io/docs/topics/annotations/helm/, you should replace the old value of the annotation, not append to it.

charts/argo-workflows/templates/controller/workflow-controller-cluster-roles.yaml
```diff
@@ -121,8 +121,18 @@ rules:
   resources:
   - leases
   resourceNames:
+  {{- if .Values.controller.instanceID.enabled }}
+  {{- if .Values.controller.instanceID.useReleaseName }}
```
vladlosev, line 125: Please indent the template flow control tags. It makes the code easier to read and is consistent with the rest of the file.

charts/argo-workflows/Chart.yaml
kobejn-jb: Previously I've misunderstood how it works, fixed.

charts/argo-workflows/templates/controller/workflow-controller-cluster-roles.yaml
kobejn-jb: I've added indent but there is very similar flow control in configMap template:
https://github.com/argoproj/argo-helm/blob/922799081d6037d71971baf3fecf3519c35326e9/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml#L9-L11
Should I also change it to make files more consistent?

vladlosev: That would be nice, please do.

kobejn-jb: Done

vladlosev (approved):
LGTM

kobejn-jb:
What is the procedure here? When can I expect this to be merged and released?

PR #850, opened 2021-07-29
feat(argo-cd): Use semverCompare to set server.staticAssets.enabled

### LINKED ISSUE #843: [argo-cd] Deprecated argument as default
**Describe the bug**
Im using the version 3.0.0 of the chart. That version as default take the latest version from argocd-server and broken because an argument called `--staticassets` was removed.

**Expected behavior**
I would like removed that argument as default.

**Additional context**
The commit relates to this bug on argo-cd is https://github.com/argoproj/argo-cd/commit/561452ac943a8f83e17c66eb72200f4a66ee3ae3

hairmare:
Now that [Argo CD 2.1.0-rc1 is tagged](https://github.com/argoproj/argo-cd/releases/tag/v2.1.0-rc1) and contains the static assets change, we can move forward with supporting it in the Helm chart.

This defaults `server.staticAssets.enabled` to `nil` and uses `semverCompare` to decide if the `--staticassets` flag needs to be added when it is `nil`. Setting it to `true`/`false` will work like before for backwards compatibility.

I'm not 100% sure if `toString` based matching to figure out if the flag is `nil` is the way to go. Let me know if you have any better suggestions. The main alternative I came up with is dropping the flag alltogether and bumping the chart to 4.0.0.

Fixes #843 

Checklist:

* [ ] simplify else/else if clause https://github.com/argoproj/argo-helm/pull/850#discussion_r679391798

### DIFF
--- a/charts/argo-cd/templates/argocd-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-server/deployment.yaml
@@ -42,10 +42,15 @@ spec:
         imagePullPolicy: {{ default .Values.global.image.imagePullPolicy .Values.server.image.imagePullPolicy }}
         command:
         - argocd-server
-        {{ if .Values.server.staticAssets.enabled }}
+        {{- if .Values.server.staticAssets.enabled }}
         - --staticassets
         - /shared/app
-        {{ end }}
+        {{- else if eq (toString .Values.server.staticAssets.enabled) "<nil>" }}
+        {{- if semverCompare "< 2.1.0-0" (default .Values.global.image.tag .Values.server.image.tag) }}
+        - --staticassets
+        - /shared/app
+        {{- end }}
+        {{- end }}
         - --repo-server
         - {{ template "argo-cd.repoServer.fullname" . }}:{{ .Values.repoServer.service.port }}
         {{- if .Values.dex.enabled }}
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.0.5
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.11.1
+version: 3.12.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-cd/assets/logo.png
 keywords:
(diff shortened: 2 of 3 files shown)


charts/argo-cd/templates/argocd-server/deployment.yaml
```diff
@@ -42,10 +42,15 @@ spec:
         imagePullPolicy: {{ default .Values.global.image.imagePullPolicy .Values.server.image.imagePullPolicy }}
         command:
         - argocd-server
-        {{ if .Values.server.staticAssets.enabled }}
+        {{- if .Values.server.staticAssets.enabled }}
         - --staticassets
         - /shared/app
-        {{ end }}
+        {{- else if eq (toString .Values.server.staticAssets.enabled) "<nil>" }}
```
mbevc1, line 48: Perhaps something like this could be clearer :thinking: 
```suggestion
        {{- else }}
```

mbevc1:
Also kind on the fence here as well to wait for the 2.1 and just bump Chart version and remove deprecated argument :slightly_smiling_face: . Happy to hear other thoughts.

charts/argo-cd/templates/argocd-server/deployment.yaml
hairmare: if i remove the weird `toString` condition you won't be able to specify `false` without triggering the `semverCompare` clause

maybe the `server.staticAssets.enabled` flag is the bug all along and we can just replace it with just the semver check

charts/argo-cd/templates/argocd-server/deployment.yaml
```diff
@@ -42,10 +42,15 @@ spec:
...
+        {{- if semverCompare "< 2.1.0-0" (default .Values.global.image.tag .Values.server.image.tag) }}
```
aditmeno, line 49: The problem with semverCompare is if I pass master or latest, this will break with an error 
`error calling semverCompare: Invalid Semantic Version`

hairmare: You can set `server.staticAssets.enabled` to `false` and it won't call `semverCompare`. While it's inconvenient, there is a way to use `latest`.

The alternative #851 drops the pattern completely and would need to be updated to support using `master` or `latest`. Would it also make sense to consider the `release-x.y` tags?

aditmeno: Yes, I'm pretty sure there'd be users out there who use that tag variant as well

hairmare:
As discussed in #851, it's easier to keep the flag and drop it in the next major refactor.

PR #846, opened 2021-07-27
fix(argo-cd): Deprecate the static assets flag passed to argocd-server

aditmeno:
This PR seeks to deprecate the `--staticassets` flag which was  deprecated  in this PR (https://github.com/argoproj/argo-cd/pull/6361) 

PR solves issue #834 

Note on DCO:

If the DCO action in the integration test fails, one or more of your commits are not signed off. Please click on the *Details* link next to the DCO action for instructions on how to resolve this.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#versioning)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/tree/master/community#contributing-to-argo).
* [ ] My build is green ([troubleshooting builds](https://argoproj.github.io/argo-cd/developer-guide/ci/)).

Changes are automatically published when merged to `master`. They are not published on branches.

PR #851, opened 2021-07-29
fix(argo-cd): Replace server.staticAssets.enabled with automation

### LINKED ISSUE #843: [argo-cd] Deprecated argument as default

hairmare:
This PR considers introducing the `server.staticAssets.enable` flag a bug and doesn't try to keep backwards compatibility ala #850.

Given that the flag was only available for a short time and the `semverCompare` solution is what i would consider the safe way forward, this might be a better fix for #843.

I put it in it's own pr since both versions would be viable from my POV.

Fixes #843
Fixes #850

### DIFF
--- a/charts/argo-cd/templates/argocd-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-server/deployment.yaml
@@ -42,7 +42,7 @@ spec:
         imagePullPolicy: {{ default .Values.global.image.imagePullPolicy .Values.server.image.imagePullPolicy }}
         command:
         - argocd-server
-        {{ if .Values.server.staticAssets.enabled }}
+        {{- if semverCompare "< 2.1.0-0" (default .Values.global.image.tag .Values.server.image.tag) }}
         - --staticassets
         - /shared/app
         {{ end }}
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.0.5
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.11.1
+version: 3.11.2
 home: https://github.com/argoproj/argo-helm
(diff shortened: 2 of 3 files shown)


hairmare:
Having introduced the flag with #846, maybe @mkilchhofer wants to weigh in on this.

charts/argo-cd/templates/argocd-server/deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
         imagePullPolicy: {{ default .Values.global.image.imagePullPolicy .Values.server.image.imagePullPolicy }}
         command:
         - argocd-server
-        {{ if .Values.server.staticAssets.enabled }}
+        {{- if semverCompare "< 2.1.0-0" (default .Values.global.image.tag .Values.server.image.tag) }}
```
aditmeno, line 45: This will break if we pass a non semver tag e.g - `master` or `latest`

mkilchhofer:
> Having introduced the flag with #846, maybe @mkilchhofer wants to weigh in on this.

Hmm I am a bit unhappy with this conditional at all. Why we not remove it completely as soon as we update our image tag defaults to 2.1.x (we could then place a not in the readme)?
The kustomize manifests also dropped it:
image
Ref: https://github.com/argoproj/argo-cd/compare/v2.0.5...v2.1.0-rc1#diff-50b74651337cc5fd88deab50cd140da7583e76148a3aa3e7031fbe3aa5949ffe

Maybe this 2.1.x release would be the next milestone to refactor our helm manifests a bit and maybe release a new major version of the chart?

mbevc1:
Yeah, as mentioned in initial attempt, I reckon we cut a new major release and just drop it following upstream 👍

hairmare:
Dropping it in the next refactor makes sense, closing this PR.

PR #861, opened 2021-08-08
fix(argo-cd): Set type of service for gRPC as NodePort.

emanuelflp:
To use AWS Load Balancer Controller, you need to set the service type as NodePort.

Note on DCO:

Checklist:

* [X] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#versioning)
* [X] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#changelog).
* [X] My build is green ([troubleshooting builds](https://argoproj.github.io/argo-cd/developer-guide/ci/)).

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -22,3 +22,4 @@ dependencies:
 annotations:
   artifacthub.io/changes: |
     - "[Added]: Add server.service.sessionAffinity setting of the Service into account when deciding which backend Pod to use"
+    - "[Fix]: Set type of service for grpc as NodePort because this is the default of ALB ingress Controller"
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -21,5 +21,4 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Add server.service.sessionAffinity setting of the Service into account when deciding which backend Pod to use"
     - "[Fix]: Set type of service for grpc as NodePort because this is the default of ALB ingress Controller"
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.0.5
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.11.2
+version: 3.11.3
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-cd/assets/logo.png
 keywords:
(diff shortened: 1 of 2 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -22,3 +22,4 @@ dependencies:
 annotations:
   artifacthub.io/changes: |
     - "[Added]: Add server.service.sessionAffinity setting of the Service into account when deciding which backend Pod to use"
```
mbevc1, line 24: Old comment needs removing as it's concatenating it upstream.

charts/argo-cd/templates/argocd-server/alb-grpc-service.yaml
```diff
@@ -20,5 +20,5 @@ spec:
   selector:
     {{- include "argo-cd.selectorLabels" (dict "context" . "name" .Values.server.name) | nindent 4 }}
   sessionAffinity: None
-  type: ClusterIP
+  type: NodePort
```
mbevc1, line 23: How come we'd need NodePort here? Upstream is not mentioning this: https://argoproj.github.io/argo-cd/operator-manual/ingress/#aws-application-load-balancers-albs-and-classic-elb-http-mode :thinking:

emanuelflp: AWS Load Balancer controller in the latest version return an error saying that can't map a targetGroup to a service without a port. This occur because ClusterIP type don't open a port in hosts to be used as destination to targetGroup. Because of this, AWS Controller stop ALB creation.

```
{"level":"error","ts":1628462993.0268402,"logger":"controller","msg":"Reconciler error","reconcilerGroup":"elbv2.k8s.aws","reconcilerKind":"TargetGroupBinding","controller":"targetGroupBinding","name":"k8s-argocd-argocdse-380b120770","namespace":"argocd"
,"error":"service type must be either 'NodePort' or 'LoadBalancer': argocd/argocd-server-grpc"}
```

charts/argo-cd/Chart.yaml
emanuelflp: Ok. Thanks. I will remove this other.

charts/argo-cd/templates/argocd-server/alb-grpc-service.yaml
emanuelflp: Maybe we need to update that docs. I will discuss about this in the argo-cd repo.

emanuelflp: Seeing here, the docs in master branch already updated, but not available in latest page.

https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/ingress.md#aws-application-load-balancers-albs-and-classic-elb-http-mode

mbevc1: Well spotted :+1:

mbevc1 (approved):

PR #877, opened 2021-08-18
chore: Migrate old probot:stale config to github actions

mkilchhofer:
This repo used Probot to handle stale issues and PRs. But commit ca3d8d3 introduced stale via GitHub actions. Let's migrate the old config from probot to the new GH actions based stale workflow and cleanup the obsolete one.


mbevc1 (approved):

davidkarlsen:
Out of curiosity: why is action preferred over app?

mkilchhofer:
The app seems disabled for quite a while. App needs to be enabled from the repo owners.

PR #878, opened 2021-08-18
feat: Enable HA deployment for Dex component

pdrastil:
This PR allows to scale Dex deployment for high availability. It also adds missing liveness / readiness probes so K8s can restart container in case Dex server is not responding for some reason. Implementation is inspired by [Dex chart](https://github.com/dexidp/helm-charts/tree/master/charts/dex)

## Open questions

When looking into deployment I was wondering on following implementation. If you can please respond to following.

1) Why are the container ports configurable via values.yaml? Only reason I can see is in case user would provide custom image, however init container that uses [argocd-dex](https://github.com/argoproj/argo-cd/blob/b37eee1054e42c873699460dd5e2447c2f9fe5a6/util/dex/config.go#L12) will always generate configuration with hardcoded port numbers. Maybe fixed port numbers should be used in deployment instead so user cannot make deployment non-functioning.

2) Why is port for metrics / telemetry hidden in deployment via `metrics.enable`? This endpoint is always running in container so I would expect that the port is also present in Deployment / Service. Maybe following approach to auto-detect Prometheus on ServiceMonitor can be used as another improvement?

```yaml
{{- if and metrics.enabled .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
```

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 2.0.5
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.12.1
+version: 3.13.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-cd/assets/logo.png
 keywords:
@@ -21,4 +21,6 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Fixed]: redis-ha.nameOverride / redis-ha.fullnameOverride breaks the ArgoCD helm chart"
+    - "[Added]: Allow to set number of Dex server replicas"
+    - "[Added]: Allow to enable auto-scaling for Dex server"
+    - "[Fixed]: Dex server uses liveness and readiness probe"
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -68,6 +68,13 @@ Create argocd repo-server name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Create argocd dex-server name and version as used by the chart label.
+*/}}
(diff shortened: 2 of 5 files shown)


pdrastil:
Another question to upstream maintainers. Why did you actually decided to use / maintain your custom Dex deployment compared to dependency sub-chart from dexidp.io? Just because of argocd-dex util?

mkilchhofer:
Hi 👋 , thanks for your contribution.

HA for Dex is not supported:

> The argocd-dex-server uses an in-memory database, and two or more instances would have inconsistent data. argocd-redis is pre-configured with the understanding of only three total redis servers/sentinels.
(Ref: https://argoproj.github.io/argo-cd/operator-manual/high_availability/#argocd-dex-server-argocd-redis)

This repo here is community maintained, we only implement things which is supported by the upstream project(s). Please have a look in the upstream repo for HA manifests and file a bug/feature request over there: https://github.com/argoproj/argo-cd/tree/master/manifests/ha

pdrastil:
@mkilchhofer Thanks for the information. Can I open new PR that adds just a probes? We found out that Dex container fails at least once a month and is not properly restarted which was the main motivation behind this PR.

mkilchhofer:
Does our dex version support health endpoints already? I think they were missing for quite a long time

pdrastil:
Yes dex itself supports this via telemetry (metrics) port on `/healthz/live` and `/healthz/ready` endpoints. It was included in this PR.

mkilchhofer:
Ah yup, I doublechecked it: Dex implemented these endpoints in version 2.28.0 and we use 2.28.1, but the official/upstream manifests for ArgoCD 2.0.5 uses dex-2.27.0:
https://github.com/argoproj/argo-cd/blob/v2.0.5/manifests/base/dex/argocd-dex-server-deployment.yaml

Yeah, feel free to implement it and file a PR. I think implement it as optional for now would be great.

pdrastil:
@mkilchhofer - Thank - new PR for probes is here https://github.com/argoproj/argo-helm/pull/880

PR #880, opened 2021-08-19
feat(argo-cd): Add probes for Dex server >= 2.28.0

pdrastil:
Add liveness and readiness probes for Dex server >= 2.28.0. For custom image tags where version cannot be determined are probes disabled by default

PR #890, opened 2021-08-23
feat(argo-cd): Add probes for Dex server >= 2.28.0

pdrastil:
Re-opened after latest changes

Checklist:

### DIFF
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -261,20 +261,20 @@ dex:
 
   ## Probes for Dex server
   ## Supported from Dex >= 2.28.0
-  probes:
+  livenessProbe:
     enabled: false
-    livenessProbe:
-      failureThreshold: 3
-      initialDelaySeconds: 10
-      periodSeconds: 10
-      successThreshold: 1
-      timeoutSeconds: 1
-    readinessProbe:
-      failureThreshold: 3
-      initialDelaySeconds: 10
-      periodSeconds: 10
-      successThreshold: 1
-      timeoutSeconds: 1
+    failureThreshold: 3
+    initialDelaySeconds: 10
+    periodSeconds: 10
+    successThreshold: 1
+    timeoutSeconds: 1
+  readinessProbe:
+    enabled: false
+    failureThreshold: 3
+    initialDelaySeconds: 10
+    periodSeconds: 10
+    successThreshold: 1
+    timeoutSeconds: 1
 
   serviceAccount:
     create: true
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -259,6 +259,19 @@ dex:
   ##
   podLabels: {}
 
+  ## Probes for Dex server
+  ## Supported from Dex >= 2.28.0
+  probes:
+    enabled: false
+    livenessProbe:
+      httpGet:
+        path: /healthz/live
+        port: 5558
+    readinessProbe:
+      httpGet:
+        path: /healthz/ready
+        port: 5558
+
   serviceAccount:
     create: true
     name: argocd-dex-server
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -264,13 +264,17 @@ dex:
   probes:
     enabled: false
     livenessProbe:
-      httpGet:
-        path: /healthz/live
-        port: 5558
+      failureThreshold: 3
+      initialDelaySeconds: 10
+      periodSeconds: 10
+      successThreshold: 1
+      timeoutSeconds: 1
     readinessProbe:
-      httpGet:
-        path: /healthz/ready
-        port: 5558
+      failureThreshold: 3
+      initialDelaySeconds: 10
+      periodSeconds: 10
+      successThreshold: 1
+      timeoutSeconds: 1
 
   serviceAccount:
     create: true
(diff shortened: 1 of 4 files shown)


pdrastil:
@mkilchhofer - Used suggested implementation from PR https://github.com/argoproj/argo-helm/pull/880, however please take a look into my comments for the change.

charts/argo-cd/values.yaml
```diff
@@ -259,6 +259,19 @@ dex:
...
   podLabels: {}
 
+  ## Probes for Dex server
+  ## Supported from Dex >= 2.28.0
+  probes:
+    enabled: false
+    livenessProbe:
+      httpGet:
+        path: /healthz/live
+        port: 5558
```
pdrastil, line 269: @mkilchhofer This kinda feels like a code smell to me. As a user I need to know which container port exposes the probe. I have similar feel towards `containerPortHttp` and other values because Dex is initialized via `argocd-dex` utility that hardcodes ports. Is there any real use case when user should be able to redefine ports for this deployment? Another subjective code smell is hiding the metrics container port based on configuration. This port still exists and is used by the container even when you have `dex.metrics.enabled: false`

pdrastil: If I can propose another change for different PR that relates to metrics port and how `metrics.enable` and `serviceMonitor.enable` currently behaves I would go with Helm 3 API capabilities check that autodetects installed Prometheus and kept metrics port on deployment always enabled / used with alias name instead of port number. What do you think?

```yaml
{{- if and .metrics.enabled .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
{{- end }}
```

mkilchhofer: Hmm. Using `.Capabilities` in context of ArgoCD is not a good idea. We had several issues, where users deployed the argo-cd chart through another ArgoCD instance. ArgoCD only uses the `helm template` subcommand which does not contact the real kubernetes API. Therefore the `.Capabilities` dict isn't populated with values reflecting the kubernetes cluster.

mkilchhofer: What about removing the conditional `{{- if .Values.dex.metrics.enabled }}` in the deployment.yaml and reference the  probe like this:
```yaml
      httpGet:
        path: /healthz/live
        port: metrics
```
?

pdrastil: Sure we can reference it with name instead of port number. What about the ports for Dex? Do you want to keep them configurable via values.yaml even when they cannot be changed at all?

Regarding API capabilities - we are using this approach on our production servers to detect Prometheus installations automatically with Argo CD 2.0.5 and it works fine. Maybe this was problem on some older version and now Argo invokes `helm template` with  `--api-versions` switch.  I'll try to take a look into this when I have a moment.

pdrastil: It's supported - see here: https://github.com/argoproj/argo-cd/blob/9476ab5e18be34a8cfc81150c94d08eefe3cee25/util/helm/cmd.go#L321

pdrastil:
Took little bit different approach - let's hardcode endpoint in deployment and let user configure just probe options

mkilchhofer (changes requested):
Looks good now. Could you update the README with the newly added variables, I then give an approve :-)

pdrastil:
@mkilchhofer Updated - thanks for the approve :)

mkilchhofer:
Looks good now.
What do you think about moving the probes to component top-level and use the `omit` function for the enabled/disabled feature? I just asks this because ever other component has 
- repoServer.livenessProbe.*
- controller.readinessProbe.*
- server.livenessProbe.*

And dex now has
- dex.**probes**.readinessProbe

pdrastil:
Top level would work fine for the parameter configuration but it would mean each probe has it's own enabled switch. Does this makes sense? There is similar section with global feature toggle for serviceAccount / monitoring. On the other hand if the Dex is upgraded on upstream manifest we can simply remove / deprecate this toggle and make it enabled by default. Your call.

pdrastil:
From the upgrade perspective the top-level makes most sense because user don't have to move section around and we can just remove enabled switch in the future.

mkilchhofer:
> Top level would work fine for the parameter configuration but it would mean each probe has it's own enabled switch. Does this makes sense? There is similar section with global feature toggle for serviceAccount / monitoring.

Yup, this is a downside but I think we can go with this one.

> On the other hand if the Dex is upgraded on upstream manifest we can simply remove / deprecate this toggle and make it enabled by default. 

What do you mean with `upstream manifest`? I think we cannot migrate to dex as a subchart because ArgoCD uses a helper binary from an init-container to generate the config file and start the dex server.

pdrastil:
> What do you mean with upstream manifest? I think we cannot migrate to dex as a subchart because ArgoCD uses a helper binary from an init-container to generate the config file and start the dex server.

I mean when they upgrade installation yamls with Dex >= 2.28.0 for some time and it will be default for most installations. Regarding the utility I was asking in previous thread if it makes sense to have `dex.containerPortHttp` and other values configurable in chart because port numbers are [hardcoded](https://github.com/argoproj/argo-cd/blob/b37eee1054e42c873699460dd5e2447c2f9fe5a6/util/dex/config.go#L12) in this utility and they cannot be changed. If someone modifies them via parameter he will get non-functioning Dex server.

mkilchhofer (approved):

PR #919, opened 2021-09-02
feat(argo-cd): Ability to provide cluster role to repo-server

ahalay:
Note on DCO:

Checklist:


mkilchhofer:
Hi @ahalay 👋
Can you explain the use case for this feature?

mkilchhofer (commented):
Hi @ahalay 👋
Can you explain the use case for this feature? I don't think reposerver needs cluster access for whatever use case ;-)

ahalay:
Hello @mkilchhofer 
In my case, it is for the custom helmfile plugin. Some helmfile hooks contain a logic with access to Kubernetes resources. But it probably would be useful for any custom plugin with similar behavior.

mkilchhofer:
> Hello @mkilchhofer
> In my case, it is for the custom helmfile plugin. Some helmfile hooks contain a logic with access to Kubernetes resources. But it probably would be useful for any custom plugin with similar behavior.

Okay 👌  this seems like a valid use case :-) Thanks for explaining that. I will have a look into your changes.

charts/argo-cd/templates/argocd-repo-server/clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.repoServer.serviceAccount.create }}
+{{- if .Values.repoServer.clusterAdminAccess.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "argo-cd.repoServer.fullname" . }}
+  labels:
+    {{- include "argo-cd.labels" (dict "context" . "component" .Values.repoServer.name "name" .Values.repoServer.name) | nindent 4 }}
+rules:
```
mkilchhofer, line 9: What do you think about reusing what we use for the controller?
https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml#L222-L229

charts/argo-cd/templates/argocd-repo-server/clusterrolebinding.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.repoServer.serviceAccount.create }}
+{{- if .Values.repoServer.clusterAdminAccess.enabled }}
```
mkilchhofer, line 2: Use `and`

charts/argo-cd/templates/argocd-repo-server/clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.repoServer.serviceAccount.create }}
+{{- if .Values.repoServer.clusterAdminAccess.enabled }}
```
mkilchhofer, line 2: Use `and`

mkilchhofer (changes requested):

charts/argo-cd/templates/argocd-repo-server/clusterrole.yaml
ahalay: I hope I understood your idea correctly 😅

ahalay (commented):
Requested changes have been made

charts/argo-cd/templates/argocd-repo-server/clusterrole.yaml
mkilchhofer: yep, looks good. will have a deeper look into it as soon as possible

mkilchhofer (approved):
lgtm. Thank you for implementing my requested changes 🙏😎

ErikLundJensen:
I have just seen the implementation. I would really have liked two flags for creating ClusterRole and ClusterRoleBinding.
Often you would like the ClusterRole, however, create RoleBinding for each namespace that ArgoCD operates on.
In new implementation we can't op-out of the ClusterRoleBinding.

PR #929, opened 2021-09-13
feat(argo-cd): Support for additional labels

nlamirault:
Note on DCO:

Checklist:


charts/argo-cd/templates/redis/deployment.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
   name: {{ template "argo-cd.redis.fullname" . }}
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.redis.name "name" .Values.redis.name) | nindent 4 }}
-    app.kubernetes.io/version: {{ .Values.redis.image.tag | quote }}
```
mkilchhofer, line 9: Please restore the version label change. Redis uses its own version which does not belong to the chart version

mkilchhofer (changes requested):

charts/argo-cd/templates/redis/deployment.yaml
nlamirault: done.

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -137,6 +137,10 @@ helm.sh/chart: {{ include "argo-cd.chart" .context }}
 {{ include "argo-cd.selectorLabels" (dict "context" .context "component" .component "name" .name) }}
 app.kubernetes.io/managed-by: {{ .context.Release.Service }}
 app.kubernetes.io/part-of: argocd
+app.kubernetes.io/version: {{ .context.Chart.AppVersion | quote }}
```
mkilchhofer, line 140: We cannot use the version in the template function. The version label is handled separately in every deployment.yaml

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -137,6 +137,10 @@ helm.sh/chart: {{ include "argo-cd.chart" .context }}
...
+{{- if .context.Values.global.additionalLabels }}
+{{ toYaml .context.Values.global.additionalLabels }}
```
mkilchhofer, line 142:
```suggestion
{{- with .context.Values.global.additionalLabels }}
{{ toYaml . }}
```

mkilchhofer (changes requested):
Additional to the comments below, I'd like to ask if `global.podLabels` would solve your need. This is already present in code.

mbevc1 (approved):

mbevc1 (changes requested):
As posted before we already have parameter that could be used.

mkilchhofer:
> As posted before we already have parameter that could be used.

There might be a use case for this but idk😅.
The parameter we already have is only on the PodSpec. This feature here would be generic for all object metadata.

mbevc1:
Ah, goatcha :)

mbevc1 (approved):

mkilchhofer (approved):

PR #962, opened 2021-10-12
fix(argo-workflows): RBAC for namespaced workflows

tiwarisanjay:
Note on DCO:

I tried to solve the DCO but still showing up as not signed off. 

Checklist:

* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog]
* [x] (https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#changelog).
* [x] I have signed off all my commits as required by [DCO]  (https://github.com/argoproj/argoproj/tree/master/community#contributing-to-argo).

Test the Function by Changing values.yaml  as following : 
```
workflow:
  # namespace: workflows # Deprecated, for backwards compatibility: specify a single namespace to run workflows in
  serviceAccount:
    create: true # Specifies whether a service account should be created
    annotations: {}
    name: "argo-workflow" # Service account which is used to run workflows
  rbac:
    # Adds Role and RoleBinding for the above specified service account to be able to run workflows
    # A Role and Rolebinding pair is also created for each namespace in controller.workflowNamespaces (see below)
    create: true
```
```
controller.extraArgs: ["--managed-namespace","workflows"]
```
```
controller.workflowNamespaces:  ["workflows"]
```

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,9 +2,9 @@ apiVersion: v2
 appVersion: 2.1.3
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.25.0
+version: 3.25.1
 home: https://github.com/argoproj/argo-helm
-icon: https://argoproj.github.io/argo-cd/assets/logo.png
+icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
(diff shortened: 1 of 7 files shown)


vladlosev:
Actually @neerfri you are right - the `controller.workflowNamespaces` setting should be considered when `singleNamespace` is false. It's my bad not paying close enough attention to this change. We will need to revert it and improve the explanations in values.yaml.

PR #987, opened 2021-10-20
feat(argocd-image-updater): Add ability to mount authentication scripts

nielstenboom:
This PR adds the ability to specify an authentication script. This is needed to be able to support ECR (https://github.com/argoproj-labs/argocd-image-updater/issues/112) and the current chart did not support it yet.

Checklist:


mkilchhofer (commented):
I think we should use an additional configmap for this and use a generic structure to define multiple auth scripts.

charts/argocd-image-updater/values.yaml
```diff
@@ -72,6 +72,14 @@ config:
     #   prefix: quay.io
     #   credentials: env:REGISTRY_SECRET
 
+# whether to create an authentication script, if enabled, an authentication script will be mounted on /script/auth.sh that can be used to authenticate with registries (ECR)
+# refer to https://argocd-image-updater.readthedocs.io/en/stable/configuration/registries/#specifying-credentials-for-accessing-container-registries for more info
+authScript:
+  enabled: false
+  script: |
+    #!/bin/sh
+    echo "add script here"
```
huynhsontung, line 81: Echoing the previous comment, instead of having a single `script` field and an `enabled` flag, maybe we can have a map of file name and content like so

```yaml
authScripts:
  auth1.sh: |
    #!/bin/sh
    echo "auth script 1 here"
  auth2.sh: |
    #!/bin/sh
    echo "auth script 2 here"
```

nielstenboom:
Sorry for the spam guys, messed up by signing off all the commits that were merged from the master branch ✌️ 

@mkilchhofer & @huynhsontung looks okay like this?

mbevc1 (approved):

huynhsontung (approved):

charts/argocd-image-updater/templates/deployment.yaml
```diff
@@ -84,13 +84,30 @@ spec:
...
           name: argocd-image-updater-config
-        name: registries-conf
+      {{- if .Values.authScripts.enabled }}
+      {{- range $k, $v := .Values.authScripts.scripts }}
+      - name: {{ $k | replace "." "-" }}
+        configMap:
+          defaultMode: 0777
+          items:
+          - key: {{ $k }}
+            path: {{ $k }}
```
mkilchhofer, line 107: Why do we need this loop here? What about just adding the configmap directly under /scripts?

charts/argocd-image-updater/templates/deployment.yaml
```diff
@@ -84,13 +84,30 @@ spec:
...
             name: tmp-dir
+          {{- if .Values.authScripts.enabled }}
+          {{- range $k, $v := .Values.authScripts.scripts }}
+          - mountPath: /scripts
+            name: {{ $k | replace "." "-" }}
+          {{- end }}
+          {{- end }}
       volumes:
-      - configMap:
+      - name: registries-conf
```
mkilchhofer, line 94: Could you please restore the alphabetical order and leave the `name:` attribute where it was?

mkilchhofer (changes requested):

charts/argocd-image-updater/templates/deployment.yaml
nielstenboom: Wasn't aware that you could mount multiple files from a single configmap, thanks!

mkilchhofer (approved):
Lgtm

charts/argocd-image-updater/values.yaml
```diff
@@ -71,6 +71,25 @@ config:
...
+    #   ping: yes
+    #   insecure: no
+    #   credentials: ext:/scripts/auth1.sh
+    #   credsexpire: 10h
+
+# whether to mount authentication scripts, if enabled, the authentication scripts will be mounted on /scripts that can be used to authenticate with registries (ECR)
+# refer to https://argocd-image-updater.readthedocs.io/en/stable/configuration/registries/#specifying-credentials-for-accessing-container-registries for more info
+authScripts:
+  enabled: false
+  scripts: {}
```
mkilchhofer, line 86:
```suggestion
  # -- Map of key-value pairs where the key consists of the name of the script and the value the contents
  scripts: {}
```

charts/argocd-image-updater/README.md
```diff
@@ -60,6 +60,8 @@ The `config.registries` value can be used exactly as it looks in the documentati
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Kubernetes affinity settings for the deployment |
+| authScripts.enabled | bool | `false` | Whether to mount the defined scripts that can be used to authenticate with a registry, the scripts will be mounted at `/scripts/yourfilename.sh` |
```
mkilchhofer, line 63:
```suggestion
| authScripts.enabled | bool | `false` | Whether to mount the defined scripts that can be used to authenticate with a registry, the scripts will be mounted at `/scripts` |
```

charts/argocd-image-updater/values.yaml
```diff
@@ -71,6 +71,25 @@ config:
...
+    #   prefix: 123456789.dkr.ecr.eu-west-1.amazonaws.com
+    #   ping: yes
+    #   insecure: no
+    #   credentials: ext:/scripts/auth1.sh
+    #   credsexpire: 10h
+
+# whether to mount authentication scripts, if enabled, the authentication scripts will be mounted on /scripts that can be used to authenticate with registries (ECR)
+# refer to https://argocd-image-updater.readthedocs.io/en/stable/configuration/registries/#specifying-credentials-for-accessing-container-registries for more info
+authScripts:
+  enabled: false
```
mkilchhofer, line 85:
```suggestion
  # -- Whether to mount the defined scripts that can be used to authenticate with a registry, the scripts will be mounted at `/scripts`
  enabled: false
```

mkilchhofer (changes requested):

mkilchhofer (approved):

Issue #696, opened 2021-04-24
[workflows]: Add Helm-Docs Support


oliverbaehler:
**Is your feature request related to a problem? Please describe.**
Currently we don't document values in the README.md for each chart. This since the charts are mostly very complex and they get out of sync very fast. 

**Describe the solution you'd like**
[Helm-Docs](https://github.com/norwoodj/helm-docs) synces README content based on values files. If we could add helm-docs to our github workflows we could simply document all values.

**Describe alternatives you've considered**

**Additional context**

chgl:
https://github.com/kubepack/chart-doc-gen is pretty neat as well. It's effectively the same as helm-docs but doesn't require a custom prefix for comments (although I tried making this configurable but haven't succeeded yet https://github.com/norwoodj/helm-docs/issues/67 ...). I'm using it in my chart repo over at https://github.com/chgl/charts.

mkilchhofer:
not stale

PR #1003, opened 2021-11-02
docs(argo-workflows): Use helm docs for workflows

tuananh:
Related issue: #696 

add a step to check if docs is staled and need update. it looks like this

image

if docs is up to date, it will look like this

image

Signed-off-by: Tuan Anh Tran <me@tuananh.org>

Note on DCO:

Checklist:


charts/argo-workflows/values.yaml
```diff
@@ -248,50 +248,48 @@ server:
...
-  ## Additional volumes to the server main container.
+  # -- Additional volumes to the server main container.
   volumeMounts: []
   volumes: []
 
-  ## Ingress configuration.
-  ## ref: https://kubernetes.io/docs/user-guide/ingress/
-  ##
+  # -- Ingress configuration.
+  # ref: https://kubernetes.io/docs/user-guide/ingress/
```
vladlosev, line 283: Putting the comment here appears to generate a huge default value in the output table, which is hard to parse. Please remove it here and add comments to the individual fields instead.

vladlosev (commented):
This is a very nice change, thanks! But as it's submitted I see a big maintenance issue: not all people making changes to the chart will be commenting the values in the proper way or running the helm-docs before each commit. Eventually the `README.md` will go out seriously of sync with `values.yaml`. So in order for this to be a viable change going forward, two things need to be done:
- Give users an easy way to re-generate `README.md` when updating the values `values.yaml`.
- Require the users to do that when submitting the pull request.

The first requirement can be fulfilled by providing a script to bootstrap and invoke helm-docs for a given chart. The second on can met by amending the CI configuration by adding a step that runs helm-docs on charts that have `README.md.gotmpl` provided, comparing results with the existing `README.md`, and failing if there is a mismatch. This will force people to make sure they always run helm-docs before submitting the PR.

charts/argo-workflows/values.yaml
tuananh: i will add a step in CI to check whether the readme is stale and need to re-generate

mkilchhofer: Hi @tuananh. Thanks for this great contribution. I like this and I also converted several charts to use helm-docs here.
Can you prefix the unnecessary comment line with two `##` so it doesn't get included in the README? This way we can get rid of the long and unnecessary information in the README.

tuananh: @mkilchhofer updated. i also add a step in CI to check if the docs is up to date or not.

.github/workflows/lint-and-test.yml
```diff
@@ -35,6 +35,12 @@ jobs:
           fi
       - name: Run chart-testing (lint)
         run: ct lint --debug --config ./.github/configs/ct-lint.yaml --lint-conf ./.github/configs/lintconf.yaml
+      
+      - name: Run docs-testing (helm-docs)
+        uses: buttahtoast/helm-release-action@v2.0.1
+        with:
+          charts: "${{ steps.list-changed.outputs.changed_charts }}"
```
vladlosev, line 42: As far as I can tell, this will pick up and try to compare the results of`helm-docs` on all the charts, not just the ones that are helm-docs enabled (have `README.md.gotmpl`). That will break the build on a change to any chart that is not yet enabled for helm-docs. One way around it is to generate a separate list of changed charts with helm-docs support and run the check only on them. For example:
```bash
          changed=$(ct --config ./.github/configs/ct-lint.yaml list-changed)
          charts=$(echo "$changed" | tr '\n' ' ' | xargs)
          if [[ -n "$changed" ]]; then
            echo "::set-output name=changed::true"
            echo "::set-output name=changed_charts::$charts"
          fi
          changed_helm_docs=$(for chart in $changed ; do if test -f "$chart/README.md.gotmpl" ; then echo "$chart" ; done)
          charts_helm_docs=$(echo "$changed_helm_docs" | tr '\n' ' ' | xargs)
          if test -n "$changed_helm_docs" ; then
            echo "::set-output name=changed_helm_docs::true"
            echo "::set-output name=changed_helm_docs_charts::$charts_helm_docs"
          fi
```

vladlosev (commented):
One other issue is you need to provide or document an easy way to setup and run `helm-docs`. Without that many people will be avoid either updating values.yaml or making changes to the chart altogether.

.github/workflows/lint-and-test.yml
mkilchhofer: We could place the same note like it exists in the image-updaters README

mkilchhofer:
Ah I will add a commit with the remaining (undocumented/missing) values asap

.github/workflows/lint-and-test.yml
mkilchhofer: Add the remaining charts to a file .helmdocsignore?

mkilchhofer: I added the .helmdocsignore. Are you okay with that, @vladlosev ?

vladlosev: Yes, looks like it will work. The only downside is the check will not be automatically enabled for a chart if one adds the template there - one will also have to remember (or to know) to delete the directory from .helmdocsignore.

mkilchhofer: Okay but in my eyes this sounds reasonable as such a contributor will be familiar with helm-docs.

vladlosev: I'll trust your opinion here, then.

mkilchhofer (approved):

PR #1043, opened 2021-11-27
feat(argo-cd): support initContainers on all pods

jMarkP:
Signed-off-by: Mark Pim <j.mark.pim@gmail.com>

This adds support for custom initContainers for all pods, not just the repo-server. I think giving users control over initContainers for all deployments adds some handy flexibility at a low cost, hopefully you agree!

My particular use case, for the record, is we have a custom K8s auth provider for our clusters, and so I need to deploy that auth provider binary to the Application Controller and Server pods, so that it's available in the container when ArgoCD attempts to connect to the remote clusters. So far I've been building a custom ArgoCD image containing our binary, but that makes the process of upgrading ArgoCD more work, and so copying it at runtime via an initContainer would simplify that.

---

Checklist:

### DIFF
--- a/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
@@ -179,7 +179,7 @@ spec:
         name: tmp-dir
       {{- if .Values.repoServer.initContainers }}
       initContainers:
-{{- toYaml .Values.repoServer.initContainers | nindent 6 }}
+      {{- toYaml .Values.repoServer.initContainers | nindent 6 }}
       {{- end }}
 {{- if .Values.repoServer.priorityClassName }}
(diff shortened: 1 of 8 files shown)


charts/argo-cd/templates/argocd-repo-server/deployment.yaml
```diff
@@ -179,7 +179,7 @@ spec:
         name: tmp-dir
       {{- if .Values.repoServer.initContainers }}
       initContainers:
-{{- toYaml .Values.repoServer.initContainers | nindent 6 }}
+      {{- toYaml .Values.repoServer.initContainers | nindent 6 }}
```
jMarkP, line 182: Changed the indentation here, which is a no-op in the output (thanks to `{{-` plus `nindent`) and seemed to match the convention elsewhere

mbevc1 (approved):

PR #1059, opened 2021-12-16
feat(argocd-applicationset): Add ingress for webhook

c0deaddict:
Implements #1014

Verified working on a setup with ArgoCD with ApplicationSet controller.

Note on DCO:

If the DCO action in the integration test fails, one or more of your commits are not signed off. Please click on the *Details* link next to the DCO action for instructions on how to resolve this.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#versioning)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/tree/master/community#contributing-to-argo).
* [x] My build is green ([troubleshooting builds](https://argoproj.github.io/argo-cd/developer-guide/ci/)).

Changes are automatically published when merged to `master`. They are not published on branches.

PR #1070, opened 2021-12-24
feat(argocd-applicationset): Update to app version v0.3.0

enys:
- Bump to 0.3.0 and add webhook port

Ci only passes in the same namespace where argocd is running.

Note on DCO:

Checklist:


enys:
Looks like it's failing like : https://github.com/argoproj/argo-helm/pull/1059. will have to have a look.

enys:
Need to add webhook port to app option in deployment

enys:
ct install --namespace argocd locally works fine, not clear why the run in GH actions fails

enys:
CT local testing, without argocd installed produces the same error. Crash with no logs.

.github/configs/ct-install.yaml
```diff
@@ -16,3 +16,5 @@ validate-maintainers: true
 validate-yaml: true
 exclude-deprecated: true
 excluded-charts: []
+# debug applicationset
+namespace: argocd
```
enys, line 20: this allows ci to pass because application set needs to be in same namespace than argo

mkilchhofer:
As we already discussed in the regression issue in the upstream repo over there: https://github.com/argoproj-labs/applicationset/issues/445, I don't feel confident to workaround the regression.
Maybe we should exclude the appset chart from being installed for now and see if a bugfix release arrives.

WDYT?

enys:
Totally agree, I'm not comfortable withe the possible side effects.
Will revert all the 'CI' changes and comment out the tests for now, tomorrow morning.
Does that work for you ?

mkilchhofer:
Yup lets exclude the chart again. (Was already excluded in the past: https://github.com/argoproj/argo-helm/pull/977/files) ;-)

.github/workflows/lint-and-test.yml
```diff
@@ -56,6 +56,13 @@ jobs:
         run: |
           kubectl apply -f charts/argo-cd/crds
 
+      # - name: Deploy latest ArgoCD
+      #   if: |
+      #     contains(steps.list-changed.outputs.changed_charts, 'argocd-applicationset')
+      #   run: |
+      #     kubectl create namespace argocd
+      #     kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml
+
```
enys, line 65: Leaving this commented for now ? Could be useful in the future ?

charts/argocd-applicationset/values.yaml
```diff
@@ -18,6 +18,9 @@ args:
   metricsAddr: :8080
   # -- The default health check port
   probeBindAddr: :8081
+  # -- The default webhook port
+  # hardcoded for now in the applicationset-controller
+  webhookAddr: :7000
```
enys, line 23: @mkilchhofer I saw you commented the PR implementing the hardcoded value of this. should we leave as is here ?

charts/argocd-applicationset/templates/deployment.yaml
```diff
@@ -56,6 +56,10 @@ spec:
             - name: metrics
               containerPort: {{ (split ":" .Values.args.metricsAddr)._1 }}
               protocol: TCP
+            - name: webhook
+              containerPort: 7000
+              containerPort: {{ (split ":" .Values.args.webhookAddr)._1 }}
```
mkilchhofer, line 61: containerPort defined twice

charts/argocd-applicationset/values.yaml
mkilchhofer: I think we can remove it for now as there is no option to configure it.

.github/workflows/lint-and-test.yml
mkilchhofer: I think we should remove this. Otherwise I'd like to use the Chart variant we use in "our" helm space here :-)

mkilchhofer:
Wohooo, there's activity fixing the crashing appset controller: https://github.com/argoproj-labs/applicationset/pull/450 🥳

charts/argocd-applicationset/values.yaml
enys: ok and done.

.github/workflows/lint-and-test.yml
enys: makes sense.

charts/argocd-applicationset/templates/deployment.yaml
enys: indeed. fixed

enys:
I have updated regarding your comments

mkilchhofer (approved):

enys:
Cool thanks, what's the next step ?

mkilchhofer:
> Cool thanks, what's the next step ?

Release it 🚀

enys:
Thanks for the review !

PR #1094, opened 2022-01-18
feat(argo-cd): Added ability to deploy extra K8s manifests

nxf5025:
Signed-off-by: Nick Fisher <nxf5025@gmail.com>

By adding a secretProviderClass object to the Argo CD helm chart we are able to mount secrets (such as our oidc config) on the fly by leveraging the Secrets Store CSI Driver.

Documentation can be found here: https://github.com/kubernetes-sigs/secrets-store-csi-driver

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.2.2
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.30.1
+version: 3.31.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 keywords:
@@ -21,4 +21,4 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Add configurable resources to copyutil initContainer"
+    - "[Added]: Add support for extra K8s manifests"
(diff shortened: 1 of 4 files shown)


mbevc1:
Thank you for your contribution! Just thinking if adding this here is right place as it created dependency on CSI driver :thinking: . I don't have strong opinion on this, but happy to hear other peoples' thoughts. @davidkarlsen @mkilchhofer @yann-soubeyrand

nxf5025:
@mbevc1 - While it does create a dependency it's completely optional for a user to go the CSI driver route. It's similar logic to the chart supporting an ingress or service monitor where the user has a choice to install an ingress controller/prometheus.

However it's definitely a worthy discussion. For example the kube-prometheus-stack helm chart handled this the same way my PR does - https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/csi-secret.yaml

On the other hand, the grafana helm chart leaves it a bit more open and gives you the ability to deploy extra K8s manifests:
 - https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/extra-manifests.yaml
 - https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L799

I'm fine with either route and am happy to update the PR accordingly, but I definitely think we need some way to deploy the secretProviderClass with the chart.

mkilchhofer:
I don't know this CSI driver yet. I will think about this PR as soon as I read through the docs of this thingy.

But today (without knowing what functionality this resource provide) I'd go the way for generic extra resources as also all bitnami charts support this for example.

nxf5025:
PR updated to leverage extra generic K8s manifest deployments

mbevc1 (approved):
Looks nicer and more generic approach with example using SecretProviderClass :+1: . Thanks!

dwerder:
Hello, we have a feature request for some solution like that for the fluxcd helm chart. Maybe I am a little bit to overcautious, but that basically means that I allow someone to install everything he likes with the chart. From a security point of view that doesn't seem to be such a good idea. How do you guys think about that?

mkilchhofer:
Hmm a helm chart doesn't address security IMHO. It is just a implementation detail on your deployment workflow. If you deploy via Kustomize as an alternative, it would be also easily possible to deploy an arbitrary manifest into the k8s API.

This needs to be addressed in the RBAC for the deploying user.

dwerder:
@mkilchhofer Thanks for the feedback

PR #918, opened 2021-09-02
feat(argocd-image-updater): Add ability to run using k8s API and monitoring

DavidSpek:
This PR adds the ability to run the argocd-image-updater using the Kubernetes API, removing the need to configure a user in Argo CD and generating an API key. It also adds a service monitor for gathering metrics with the prometheus operator, as well as the example Grafana dashboard for those that would like to deploy it.

I will update the README soon to include these changes if they are agreed upon.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#versioning)
* [ ] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/master/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/tree/master/community#contributing-to-argo).
* [ ] My build is green ([troubleshooting builds](https://argoproj.github.io/argo-cd/developer-guide/ci/)).

Changes are automatically published when merged to `master`. They are not published on branches.

PR #1097, opened 2022-01-23
feat(argocd-image-updater): Add hooks for creating secrets

ulfox:
Bump chart version

Note on DCO:

Checklist:

### DIFF
--- a/charts/argocd-image-updater/Chart.yaml
+++ b/charts/argocd-image-updater/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argocd-image-updater
 description: A Helm chart for Argo CD Image Updater, a tool to automatically update the container images of Kubernetes workloads which are managed by Argo CD
 type: application
-version: 0.6.1
+version: 0.7.0
 appVersion: v0.11.2
 home: https://github.com/argoproj-labs/argocd-image-updater
 icon: https://argocd-image-updater.readthedocs.io/en/stable/assets/logo.png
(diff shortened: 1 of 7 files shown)


ulfox:
I was thinking that this could be easier to maintain if it the scripting was moved as part of a container.

- apt install would not be needed
- we can use more advanced scripting (compared to when we write scripts in the job's arguments)

WDYT?

mkilchhofer:
Hmm I don't use this image updater. But reading the upstream documentation you're implementing an edge case, right? This is only needed if ArgoCD runs on a different cluster than the image updater, right?

ulfox:
It can work with the same cluster also. Essentially it auto-creates the secret `argocd-image-updater-secret` that is required by image updater to communicate with ArgoCD. If I am not mistaken that secret part requires manual steps.

With the chart hooks we can auto create this secret. The only requirement is that argocd has an account with apiKey capability.

Is this functionality already implemented in the image updater and I completely missed that out?

mkilchhofer:
As far as I understand is that the upstream team from argoproj-labs implented a method to communicate to ArgoCD via native K8s (through RBAC/CRDs).
Someone tried to implemented this in this PR here: https://github.com/argoproj/argo-helm/pull/918

Xref for upstream: https://github.com/argoproj-labs/argocd-image-updater/pull/144

ulfox:
Hm, makes sense. I was not aware of that. Will close this PR then!

PR #1101, opened 2022-01-26
fix: controller service should be headless when controller.enableStatefulSet is enabled

Mattes83:
As stated in https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#limitations services for statefulsets should be headless

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.2.2
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.31.1
+version: 3.31.2
 home: https://github.com/argoproj/argo-helm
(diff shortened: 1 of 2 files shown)


mkilchhofer:
Hmmm.. We cannot implement this as this element is immutable:
> spec.clusterIPs[0]: Invalid value: []string{"None"}: may not change once set

Also I think we should follow what our upstream colleagues do: https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-application-controller/service.yaml

So I'd suggest to close this one here and you open a PR in the upstream repo?

mkilchhofer (changes requested):
Hmmm.. We cannot implement this as this field is immutable:
> spec.clusterIPs[0]: Invalid value: []string{"None"}: may not change once set

Also I think we should follow what our upstream colleagues do: https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-application-controller/service.yaml

So I'd suggest to close this one here and you open a PR in the upstream repo?

PR #1161, opened 2022-03-07
fix(argo-cd): create service account for repo server by default

eddycharly:
This PR enables the creation of a service account for repo server by default.

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.2.5
 description: A Helm chart for ArgoCD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 3.35.2
+version: 3.35.3
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 keywords:
@@ -21,4 +21,4 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Fixed]: Quote clusterResources value to avoid invalid Secret manifest"
+    - "[Fixed]: Create service account for repo server by default"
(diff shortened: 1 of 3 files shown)


mbevc1:
Thanks, seems this would align this to other charts. Could you please provide changelog infor as well?

eddycharly:
@mbevc1 sure, how shall I provide the changelog ?

mbevc1:
There is an annotation that is used, i.e. https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/Chart.yaml#L22-L24

mbevc1:
Thanks! Mind you need to replace current one as artifacthub is appending it automatically.

eddycharly:
Done, do I need to bump the chart version ?
I guess i will have to regen the docs too.

mbevc1:
Yes, please. Version needs bumping as well. Docs should be fine unless there are any changes related to those.

eddycharly:
@mbevc1 version bumped and docs up to date.

mbevc1 (approved):

mbevc1:
Looks good, just before merging thinking if we should bump minor version as this changes current default :thinking: 
Thoughts @mkilchhofer @davidkarlsen @yann-soubeyrand ?

mkilchhofer (commented):
I wanted to ask for the use case? :-)

The upstream manifest (which are our template) do not create a ServiceAccount for the repo-server. Also the server does not have a mounted service account token by default

eddycharly:
@mkilchhofer the use case is to identify to allow usage of `CiliumNetworkPolicy` with rules based on service accounts.

eddycharly:
IMHO it shouldn't hurt to have a service account. Even if you don't specify one it will fallback to use `default`.
TBH, i don't see any case where it's not advisable to create one.

mkilchhofer:
> @mkilchhofer the use case is to identify to allow usage of `CiliumNetworkPolicy` with rules based on service accounts.

Okay. This is a valable use case, I also override to true as I use helm plugins talking to the kube-api.

But it still is not comply in these two ideas:
- charts README
  > The default installation is intended to be similar to the provided ArgoCD [releases](https://github.com/argoproj/argo-cd/releases).
- least privilege principle (okay to be fair, the default SA normally also habe kube-api access)

Idk, I am okay if we proceed and merge as it is such a small change ;)

eddycharly:
> Idk, I am okay if we proceed and merge as it is such a small change ;)

Let's do that ;)

PR #1167, opened 2022-03-10
fix(argo-cd): ArgoCD application controller - metrics application labels

apj0nes:
Previously, ArgoCD application controller application labels metric was incorrectly templated.
This PR changes `.Values.controller.metrics.applicationLabels.labels` from a map to a list and updates the templating to set the correct arguments as seen [here](https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/#exposing-application-labels-as-prometheus-metrics).

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/templates/argocd-application-controller/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/deployment.yaml
@@ -57,9 +57,9 @@ spec:
         - --loglevel
         - {{ .Values.controller.logLevel }}
         {{- if .Values.controller.metrics.applicationLabels.enabled }}
-        - --metrics-application-labels
         {{- range .Values.controller.metrics.applicationLabels.labels }}
-        - {{- toYaml . | nindent 8 }}
+        - --metrics-application-labels
+        - {{ . }}
         {{- end }}
         {{- end }}
         {{- if or (and .Values.redis.enabled (not $redisHa.enabled)) (and $redisHa.enabled $redisHa.haproxy.enabled) }}
--- a/charts/argo-cd/templates/argocd-application-controller/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/deployment.yaml
@@ -57,8 +57,8 @@ spec:
         - --loglevel
         - {{ .Values.controller.logLevel }}
         {{- if .Values.controller.metrics.applicationLabels.enabled }}
-        {{- range .Values.controller.metrics.applicationLabels.labels }}
         - --metrics-application-labels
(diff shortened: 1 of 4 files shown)


mbevc1 (approved):

apj0nes:
@mkilchhofer can you clarify what change you are suggesting.
Are you saying you would still like the toYaml so it can take a list of maps? e.g.
`{{ . }}` -> `{{ toYaml . }}`
Does it make sense to do this?
```
labels:
 - { foo: bar }
```

charts/argo-cd/templates/argocd-application-controller/deployment.yaml
```diff
@@ -57,9 +57,9 @@ spec:
         - --loglevel
         - {{ .Values.controller.logLevel }}
         {{- if .Values.controller.metrics.applicationLabels.enabled }}
-        - --metrics-application-labels
         {{- range .Values.controller.metrics.applicationLabels.labels }}
-        - {{- toYaml . | nindent 8 }}
+        - --metrics-application-labels
+        - {{ . }}
         {{- end }}
         {{- end }}
```
mkilchhofer, line 64: > @mkilchhofer can you clarify what change you are suggesting. Are you saying you would still like the toYaml so it can take a list of maps? e.g. `{{ . }}` -> `{{ toYaml . }}` Does it make sense to do this?
> 
> ```
> labels:
>  - { foo: bar }
> ```



No, you moved the `--metrics-application-labels` into the range block. Like this every item in the list of strings will be "prefixed" with `--metrics-application-labels` which I think is not necessary.

apj0nes: I was following the docs here -> https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/#exposing-application-labels-as-prometheus-metrics

But if you say it is not necessary, I can change

apj0nes: Moved `--metrics-application-labels` outside of range block as suggested

mkilchhofer:
Oh, you are right. This was a bug. Cobra arg parsing for lists are
```console 
--myflag foo,bar,mitz,mau
```

or 

```console 
--myflag foo --myflag bar --myflag mitz --myflag mau
```

Sorry for that. Would you mind to revert your last commit?

mkilchhofer (approved):
Did the revert for you :-)

LGTM, thanks you very much for this contribution and make me have a look into the code and the cobra library ;-)

Issue #1199, opened 2022-03-28
Helm chart not providing default value for applicationset service targetport


visheyra:
### Describe the bug

At this [location](https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/templates/argocd-applicationset/service.yaml) the service port is defined as follow:

```
# ommited for clarity
spec:
  ports:
  - name: {{ .Values.applicationSet.service.portName }}
    port: {{ .Values.applicationSet.service.port }}
    targetPort: {{ .Values.applicationSet.containerPort }}
```

However `Values.applicationSet.containerPort` has no value set in the [default values.yaml file](https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/values.yaml) meaning that the chart will not be rendered properly using default values only.

### Related helm chart

argo-cd

### To Reproduce

clone the repo as is:
```
git clone https://github.com/argoproj/argo-helm
cd argo-helm/charts/
helm template argo-cd argo-cd -f argo-cd/values.yaml
```

see output below
```
# ommited for clarity
---
# Source: argo-cd/templates/argocd-applicationset/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: argo-cd-argocd-applicationset-controller
  labels:
    helm.sh/chart: argo-cd-4.2.3
    app.kubernetes.io/name: argocd-applicationset-controller
    app.kubernetes.io/instance: argo-cd
    app.kubernetes.io/component: applicationset-controller
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/part-of: argocd
spec:
  ports:
  - name: webhook
    port: 7000
    targetPort:
  selector:
    app.kubernetes.io/name: argocd-applicationset-controller
    app.kubernetes.io/instance: argo-cd
 ---
 # ommited for clarity
```

### Expected behavior

for the containerPort to be populated and the chart to build without additional configuration

PR #1200, opened 2022-03-28
fix(argo-cd): Correct ApplicationSet controller port

visheyra:
This PR propose a fix for the #1199.

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -21,5 +21,4 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Fixed]: Correct ArgoCD notification subscriptions type"
     - "[Fixed]: Correct ArgoCD applicationset controller port"
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.3.2
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 4.2.3
+version: 4.2.4
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 keywords:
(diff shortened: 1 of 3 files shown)


mkilchhofer (changes requested):
Hi @visheyra, thanks for your contribution.

These tasks still are open:
- Adapt Chart.yaml: bump version, and write a changelog line
- Sign-off your commit (git commit -s --amend)

For the change itself, i'd suggest to omit the containerPort from values and "hardcode" it as introduced here: https://github.com/argoproj/argo-helm/commit/e7a27468b235d867929dd921f58c5fa11fb3850c

It seems that @mikeeq missed that while he implemented the 2.3 changes

visheyra:
@mkilchhofer Okay, I think indeed hardcoding it is better ^^. I pushed/force I think this cover all your points now ?

charts/argo-cd/templates/argocd-applicationset/service.yaml
```diff
@@ -15,7 +15,7 @@ spec:
   ports:
   - name: {{ .Values.applicationSet.service.portName }}
     port: {{ .Values.applicationSet.service.port }}
-    targetPort: {{ .Values.applicationSet.containerPort }}
+    targetPort: 7000
```
mkilchhofer, line 18:
```suggestion
    targetPort: webhook
```
Let's use the named port - then its less hardcoding ;)

charts/argo-cd/Chart.yaml
```diff
@@ -22,3 +22,4 @@ dependencies:
 annotations:
   artifacthub.io/changes: |
     - "[Fixed]: Correct ArgoCD notification subscriptions type"
```
mkilchhofer, line 24: The old line needs to be replaced/removed

mkilchhofer (changes requested):

charts/argo-cd/Chart.yaml
```diff
@@ -22,3 +22,4 @@ dependencies:
...
+    - "[Fixed]: Correct ArgoCD applicationset controller port"
```
mkilchhofer, line 25:
```suggestion
    - "[Fixed]: Correct ArgoCD applicationset controller port"
```
The old line needs to be replaced/removed

mkilchhofer (approved):
Lgtm now 👍

Issue #1249, opened 2022-04-29
Argo Events improvements


pdrastil:
### Is your feature request related to a problem?

I would like to align Argo Events Helm chart closer to Argo CD structure. Namely user experience isn't great in following areas:

- Redefining image on 3 places when behind corporate proxy
- Redefining image pull secrets on multiple places
- Redefining pod labels on a multiple places
- Redefining host aliases on a multiple places
- Running controllers with restricted / namespaced installation by default that quickly pollutes the namespace with deployments created from CRDs
- Service account creation / configuration is not exactly typical nowadays
- Core RBAC is exposed in values.yaml instead placed into Role / ClusterRole

### Related helm chart

argo-events

### Describe the solution you'd like

Argo Events 1.7.0 will introduce different EventBus and configuration of multiple versions for bus controllers via ConfigMap. What do you think @mkilchhofer about refactoring / cleaning up this chart a little bit and aligning it closer how Argo CD chart looks like.

mkilchhofer:
Yes aligning all charts to follow one structure is great. As I am not one of the CODEOWNERS of that repo I act a bit passive here :-) But lets proceed and push your PR forward 👍

Issue #550, opened 2021-01-20
[Release] Helm charts for deploying ArgoCD CRDs


sbose78:
Given that Helm doesn't "upgrade" CRDs in `/crds`, we might want to consider maintaining two charts per release. 

1) One chart which has the CRDs in `/templates`
2) Another chart that references (1).

Happy to hear what the community thinks.

sbose78:
( sorry, this isn't a bug per se )

hawksight:
I like the sound of this idea @sbose78 and matches exactly the recommendations made [here](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) by helm project.

I am facing this issue as I have one instance or argo for all our clusters, regularly upgrading it. But as the CRDs are evolving.. I can't make use of all the features. currently.

I am interested to see how everyone else is getting around the issue of updating the crds when helm installed. Is everyone else just getting the yaml files and updating?

jppitout:
I think this is a great idea. Right now CRDs can't even be installed at all: 
```
Steps to reproduce the behavior:
  1. $ kubectl apply -f https://github.com/argoproj/argo-cd/blob/master/manifests/crds/application-crd.yaml 
error: error parsing https://github.com/argoproj/argo-cd/blob/master/manifests/crds/application-crd.yaml: error converting YAML to JSON: yaml: line 92: mapping values are not allowed in this context

  2. $ kubectl apply -f https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/crds/crd-project.yaml
error: error parsing https://github.com/argoproj/argo-helm/blob/master/charts/argo-cd/crds/crd-project.yaml: error converting YAML to JSON: yaml: line 92: mapping values are not allowed in this context
```
Tried logging a bug but for some reason not able to.

jppitout:
What is the best way to automate updating CRDs since they aren't part of the Helm chart anymore?

jppitout:
Thanks for the info! 

Our ArgoCD is deployed through a CI/CD pipeline which means manual steps aren't easily handled but any breaking changes should be picked up in our sandbox environment when ArgoCD is deployed there. Based on this use case I think a dependency chart would  be great! Also it would help us to tie specific chart versions to CRD versions. 

Is there a way to check which CRD versions should be paired with each chart version?

jppitout:
bump

jppitout:
[...]

PR #1269, opened 2022-05-08
feat(argo-events): Upgrade argo-events to v1.7.0

pdrastil:
Opened this PR to get initial feedback from maintainers on 2.x version of chart for argo-events with image version 1.7.0. This release is breaking because all parts were unified into single binary - ref [blog post](https://blog.argoproj.io/argo-events-v1-7-theres-a-new-bus-in-town-21949e915771) and new configuration options were introduced. I will be testing this more in upcoming days because I have a business need to use this so PR might change a little bit.

Note:
Some webhook resources are not templated on purpose because controller expects hardcoded values found [here](https://github.com/argoproj/argo-events/blob/41936c63b52be2964174d759b681d3abd3eb1855/webhook/cmd/start.go#L45)

Changes:
- Similar structure to argo-cd chart
- Upgraded image to v1.7.0 that consolidated 3 controllers into 1
- Cluster-wide install is now a default option 
- Removed RBAC that allowed to modify CRDs by default
- Added global options to override the images and shared properties
- Added prometheus support
- Added PDBs to support HA deployments

Relates to:
- https://github.com/argoproj/argo-helm/issues/1249

Checklist:


pdrastil:
Any chance I can get your attention @jbehling @VaibhavPage for this PR please?

braghettos:
+1 waiting for this PR

mkilchhofer:
As we introduce a breaking change and bump the major version I thought about moving the CRDs into the templates folder to manage them. WDYT?

charts/argo-events/values.yaml
```diff
@@ -1,163 +1,403 @@
...
   # -- [Node selector]
   nodeSelector: {}
-  # -- Labels to be added to event source controller pods
-  podLabels: {}
-  # -- Priority class for the event source controller
-  priorityClassName: ""
   # -- [Tolerations] for use with node taints
   tolerations: []
-  # -- Assign custom [affinity] rules to the event source controller
+  # -- Asign custom [affinity] rules to the deployment
```
mkilchhofer, line 202: Typo: Asign -> Assign

charts/argo-events/values.yaml
```diff
@@ -1,163 +1,403 @@
...
+  extraContainers: []
+
+  # -- Init containers to add to the events controller pods
+  initContainers: []
+
+  serviceAccount:
+    # -- Create a service account for the events controller
+    create: true
+    # -- Service account name
+    name: argo-events
```
mkilchhofer, line 235:
```suggestion
    name: ""
```

What about using the template and set this to empty string?

charts/argo-events/values.yaml
```diff
@@ -1,163 +1,403 @@
...
   # -- [Node selector]
   nodeSelector: {}
-  # -- Labels to be added to event event bus controller pods
-  podLabels: {}
-  # -- Priority class for the event bus controller
-  priorityClassName: ""
   # -- [Tolerations] for use with node taints
   tolerations: []
-  # -- Assign custom [affinity] rules to the event bus controller
+  # -- Asign custom [affinity] rules to the deployment
```
mkilchhofer, line 372: Typo

charts/argo-events/values.yaml
```diff
@@ -1,163 +1,403 @@
...
+  #    memory: 512Mi
+  #  requests:
+  #    cpu: 250m
+  #    memory: 256Mi
+
+  serviceAccount:
+    # -- Create a service account for the admission webhook
+    create: true
+    # -- Service account name
+    name: argo-events-webhook
```
mkilchhofer, line 399:
```suggestion
    name: ""
```

charts/argo-events/values.yaml
```diff
@@ -1,163 +1,403 @@
...
-  # -- Repository to use for the event source image
-  eventsourceImage: argoproj/argo-events
-  # -- Annotations to be added to event source controller pods
+
+  # -- Additional labels to add to all resources
+  additionalLabels: {}
+    # app: argo-cd
+
+  # -- Toggle and define securityContext. See [values.yaml]
+  securityContext:
```
mkilchhofer, line 34:
```suggestion
  securityContext: {}
```

mkilchhofer (changes requested):

charts/argo-events/values.yaml
```diff
@@ -1,163 +1,403 @@
...
+
+  ## Events controller metrics configuration
+  metrics:
+    # -- Deploy metrics service
+    enabled: false
+    applicationLabels:
+      # -- Enables additional labels in argocd_app_labels metric
+      enabled: false
+      # -- Additional labels
+      labels: []
```
mkilchhofer, line 249: Drop `applicationLabels` as these are Argo CD app controller specific (`--metrics-application-labels`)

pdrastil:
> As we introduce a breaking change and bump the major version I thought about moving the CRDs into the templates folder to manage their lifecycle. WDYT?

Why to move it from crd folder? Argo CD is also using normal Helm way instead of including CRDs into templates.

mkilchhofer:
> > As we introduce a breaking change and bump the major version I thought about moving the CRDs into the templates folder to manage their lifecycle. WDYT?
> 
> Why to move it from crd folder? Argo CD is also using normal Helm way instead of including CRDs into templates.

Yes but the Argo CD chart has the ability to deploy the CRDs and CR's of that types in one chart. Therefore the CRDs needs to be in the `/crds` folder, otherwise it would lead to dependency issues.

But there's an issue open to discuss the further movement of the CRD handling: https://github.com/argoproj/argo-helm/issues/550

mkilchhofer:
Exaclty. And argo-rollouts also uses CRDs in a subdirectory of /templates: https://github.com/argoproj/argo-helm/tree/main/charts/argo-rollouts/templates

pdrastil:
Ok in such case I would use the same pattern from argo-rollouts where user can enable / disable / keep them from chart configuration.

pdrastil:
@mkilchhofer Any additional improvements you would like?

charts/argo-events/values.yaml
```diff
@@ -1,163 +1,406 @@
...
+        # prometheus: kube-prometheus
+      # -- Prometheus ServiceMonitor namespace
+      namespace: "" # "monitoring"
+      # -- Prometheus ServiceMonitor labels
+      additionalLabels: {}
+
+## Argo Events admission webhook
+webhook:
+  # -- Enable admission webhook. Applies only for cluster-wide installation
+  enabled: true
```
mkilchhofer, line 281: Maybe we should disable it by default as it's documented as an (ALPHA) extension:
https://argoproj.github.io/argo-events/validating-admission-webhook/

charts/argo-events/README.md
```diff
@@ -31,81 +31,135 @@ You can install the CRDs manually from `crds` folder.
```
mkilchhofer, line 31: We should add an ## Upgrading section inside the README.

And maybe we place a howto in it how users can adopt existing CRDs:

```bash
for crd in "eventbus.argoproj.io" "eventsources.argoproj.io" "sensors.argoproj.io"; do
  kubectl label --overwrite crd $crd app.kubernetes.io/managed-by=Helm
  kubectl annotate --overwrite crd $crd meta.helm.sh/release-namespace=<YOUR_NAMESPACE>
  kubectl annotate --overwrite crd $crd meta.helm.sh/release-name=<YOUR_HELM_REALEASE>
done
```

mkilchhofer (changes requested):
> @mkilchhofer Any additional improvements you would like?

Added two more :D

charts/argo-events/README.md
pdrastil: I'm not 100% sure about this - as CRDs are global shared resource I haven't added any labels in these. How should it work in case of namespace install?

charts/argo-events/values.yaml
pdrastil: Sure we can keep it opt-in feature.

charts/argo-events/README.md
pdrastil: @mkilchhofer Any update on this?

mkilchhofer: When they live inside the `/templates` directory helm will try to manage them. If these labels/annotations are absent, helm will panic and say that the resource already exists.

These labels and annotations are added on first install normally for all resources living in the /templates directory.

pdrastil: @mkilchhofer Added section with Upgrade information that contains your script.

mkilchhofer (approved):

mkilchhofer (approved):

Issue #1187, opened 2022-03-21
[argo-cd] - bot container cannot find secret in default helm chart values. 


haghabozorgi:
### Describe the bug

default helm values for notifications controller, when the bot is enabled leads to the bot throwing error.
`notifications.name` is defaulted to `name: notifications-controller`
but the bot throws an error since it is seems to be looking for  `argocd-notifications-secret`

`time="2022-03-21T21:49:41Z" level=error msg="Failed to process: secret \"argocd-notifications-secret\" not found" resource=argocd/prometheus-operator`

### Related helm chart

argo-cd

### To Reproduce

using chart  4.2.1

```
notifications:
  bots:
     slack:
       enabled: true
```

### Expected behavior

expect the bot to lookup the secret relative to the name specified in notifications.name given that is the name the secret is created under? Not sure where it is set exactly though as I dont see any env var/command in deployment spec of bot in chart that could be responsible?

### Screenshots

![Screenshot_20220321_145157](https://user-images.githubusercontent.com/1186748/159369599-babf9e69-d16e-4620-9c5c-f69c166b1e0f.png)

### Additional context

Note: Workaround it seems is removing -controller from name like this...This seems to cause the bot pod to not log these errors.

```
notifications:
  name: notifications
  bots:
     slack:
       enabled: true
```

fe-ax:
This is due to the default value is not `argocd-notifications-secret` but `argocd-notifications-controller-secret`

PR #1299, opened 2022-05-27
fix(argo-cd): argo-cd notifications config map & secret names

Voziv:
The notifications controller is hard coded to look for argocd-notifications-secret and argocd-notifications-cm. Unless this changes I think the sane default should be `argocd-notifications-cm` & `argocd-notifications-secret`

fixes: #1187

Checklist:

* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/tree/main/community#contributing-to-argo).

### DIFF
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -254,7 +254,7 @@ Create the name of the notifications controller secret to use
 */}}
 {{- define "argo-cd.notifications.secretName" -}}
 {{- if .Values.notifications.secret.create -}}
-    {{ default (printf "%s-secret" (include "argo-cd.notifications.fullname" .)) .Values.notifications.secret.name }}
+    {{ default "argocd-notifications-secret" .Values.notifications.secret.name }}
 {{- else -}}
     {{ default "argocd-notifications-secret" .Values.notifications.secret.name }}
 {{- end -}}
@@ -265,7 +265,7 @@ Create the name of the configmap to use
 */}}
 {{- define "argo-cd.notifications.configMapName" -}}
 {{- if .Values.notifications.cm.create -}}
-    {{ default (printf "%s-cm" (include "argo-cd.notifications.fullname" .)) .Values.notifications.cm.name }}
+    {{ default "argocd-notifications-cm" .Values.notifications.cm.name }}
 {{- else -}}
     {{ default "argocd-notifications-cm" .Values.notifications.cm.name }}
 {{- end -}}
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -2212,7 +2212,7 @@ notifications:
     annotations: {}
 
     # -- The name of the secret to use.
-    ## If not set and create is true, the default name 'argocd-notifications-secret' is used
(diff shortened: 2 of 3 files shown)


charts/argo-cd/values.yaml
```diff
@@ -2197,7 +2197,7 @@ notifications:
     annotations: {}
 
     # -- The name of the secret to use.
-    ## If not set and create is true, the default name 'argocd-notifications-secret' is used
+    ## If not set the default name 'argocd-notifications-secret' is used
```
morey-tech, line 2200: If we are removing `and create is true` in this comment, should it also be removed in the comment for `cm.name`?

Voziv:
> We only speak about the bot container right?
> 
> All other things should be setup correctly. The name is injected as CLI argument already:

Ah I see. I think the bug may be in the argo-cd notifications code then if it's supposed to be taking the names of the configmap/secret as arguments.

I'm not super well versed in go yet, but I believe it's this line that causes the errors: https://github.com/argoproj/argo-cd/blob/4f155b539b925c2ad732a57fd03a5b5fb8605ae2/cmd/argocd/commands/admin/notifications.go#L33

`settings.GetFactorySettings(argocdService, "argocd-notifications-secret", "argocd-notifications-cm"), func(clientConfig clientcmd.ClientConfig) {`

That said I've noticed that we're still on argocd 2.3.3, I'll try updating our chart version tomorrow morning to see if it's still happening.

Voziv:
Updating to the latest chart I still get the same error

`time="2022-06-02T13:41:33Z" level=error msg="Failed to process: secret \"argocd-notifications-secret\" not found" resource=argocd/linkerd`

Note: `linkerd` is just the name of an app, every single app name is listed with this error in the argocd-notifications-controller-bot pod


However I do agree with what @mkilchhofer is getting at. This is likely a bug in argo-cd so I'll file a report there. 

In the meantime a fix I can apply myself is overriding the config and secret names using `notifications.cm.name` and `notifications.secret.name` values.

mkilchhofer:
I also started a discussion in slack over there: https://cloud-native.slack.com/archives/C020XM04CUW/p1654109494458449

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -254,7 +254,7 @@ Create the name of the notifications controller secret to use
 */}}
 {{- define "argo-cd.notifications.secretName" -}}
 {{- if .Values.notifications.secret.create -}}
```
gadiener, line 256: The if here is probably no longer needed

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -265,7 +265,7 @@ Create the name of the configmap to use
 */}}
 {{- define "argo-cd.notifications.configMapName" -}}
 {{- if .Values.notifications.cm.create -}}
```
gadiener, line 267: Same here, the if is probably no longer needed

gadiener (changes requested):

mkilchhofer:
PR #1340 also fixed things regarding this context here except for the bot

pdrastil:
Resolved via https://github.com/argoproj/argo-helm/pull/1419

PR #1408, opened 2022-08-19
chore(github): Use Makefile instead of scripts

oliverbaehler:
@mkilchhofer what you think?

Checklist:


mkilchhofer:
Since I am not very familiar with Makefiles, I do not push this idea forward.
For me, Makefiles are harder to read ;-)

I'd like to hear on others. @jmeridth , @mbevc1 , @yann-soubeyrand , @yu-croco , @pdrastil WDYT?

jmeridth:
I'm bias towards Makefiles, but I agree with @mkilchhofer's perspective of them being harder to read.  Understanding variable scope/knowing that each line is run in its own shell/etc are "quirks" that unexperienced contributors may have issues with re: Makefiles.

I'd still vote for a Makefile but look forward to other's inputs on this.

If this moves forward @oliverbaehler, the `helm-docs.sh` file is mentioned in the argo-events helm chart [`README.md`](https://cs.github.com/argoproj/argo-helm/blob/5b6cb2678a4558c39100084c7c0e8f5dd58de430/charts/argo-events/README.md#L17) and [`REAMDE.md.gotmpl`](https://cs.github.com/argoproj/argo-helm/blob/5b6cb2678a4558c39100084c7c0e8f5dd58de430/charts/argo-events/README.md.gotmpl#L17) files also. Don't know if it should be but want you to be aware.

mbevc1:
I'm probably also slightly biased towards Makefile, but I don't have strong preference for either in this case as it doesn't seem to add much value :shrug: . If we decide to go for Makefile, looks like it missing dropping to privileges when running `docker` and `-u $(id -u)` is missing?

pdrastil:
I'm also slightly biased towards Makefile which is simple to use from user perspective. I would be careful about formatting of the file because Makefile is tab sensitive if I remember correctly and finding bugs in whitespace is not fun during code review.

yu-croco:
In addition to my poor experience to Makefile, it's happy for me to keep using ShellScript in the aspect of the current ShellScript quantity... 😇

PR #1448, opened 2022-09-12
feat(argo-cd): Allow setting the private key algorithm in the Certificate spec

xaque208:
Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v2.4.12
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.4.4
+version: 5.4.5
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 keywords:
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -19,4 +19,4 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Changed]: Update to app version 2.4.12"
+    - "[Added]: Add privateKey spec handling from CertificatePrivateKey"
--- a/charts/argo-cd/templates/argocd-server/certificate.yaml
+++ b/charts/argo-cd/templates/argocd-server/certificate.yaml
@@ -31,6 +31,10 @@ spec:
     {{- end }}
     kind: {{ .Values.server.certificate.issuer.kind | quote }}
     name: {{ .Values.server.certificate.issuer.name | quote }}
+  privateKey:
+    {{- with .Values.server.certificate.privateKey.algorithm }}
+      algorithm: {{ . | quote }}
+    {{- end }}
   {{- with .Values.server.certificate.renewBefore }}
   renewBefore: {{ . | quote }}
   {{- end }}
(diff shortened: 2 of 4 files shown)


charts/argo-cd/values.yaml
```diff
@@ -1007,6 +1007,9 @@ server:
     duration: ""
     # -- How long before the currently issued certificate's expiry cert-manager should renew the certificate. Value must be in units accepted by Go time.ParseDuration
     renewBefore: ""
+    # -- The private key algorithm to use in the Certificate spec.
+    privateKey:
+      algorithm: ""
```
pdrastil, line 1012: How about exposing whole [CertificatePrivateKey](https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificatePrivateKey) spec like below with sensible defaults?

```yaml
# Private key of the certificate
privateKey:
  # -- Rotation policy of private key when certificate is re-issued. Either: `Never` or `Always`
  rotationPolicy: Never
  # -- The private key cryptography standards (PKCS) encoding for private key. Either: `PCKS1` or `PKCS8`
  encoding: PKCS1
  # -- Algorithm used to generate certificate private key. One of: `RSA`, `Ed25519` or `ECDSA`
  algorithm: RSA
  # -- Key bit size of the private key. If algorithm is set to `Ed25519`, size is ignored.
  size: 2048
```

charts/argo-cd/templates/argocd-server/certificate.yaml
```diff
@@ -31,6 +31,10 @@ spec:
     {{- end }}
     kind: {{ .Values.server.certificate.issuer.kind | quote }}
     name: {{ .Values.server.certificate.issuer.name | quote }}
+  privateKey:
```
pdrastil, line 34: Please check for the whole specification of the privateKey below and use following construct:

```yaml
{{- with .Values.server.certificate.privateKey }}
privateKey:
  {{- toYaml . | nindent 4 }}
{{- end }}
```

pdrastil (changes requested):
Hi,
thanks for contribution. Please take a look on suggested changes.

charts/argo-cd/Chart.yaml
```diff
@@ -20,3 +20,4 @@ dependencies:
 annotations:
   artifacthub.io/changes: |
     - "[Fixed]: Additional initContainers are handled correctly in the dex deployment"
```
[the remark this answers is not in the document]
xaque208: Correct.  It was there before I made the change.  Is there another way I should note the changes?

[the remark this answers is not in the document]
xaque208: I've scrubbed the old entries during the rebase.

charts/argo-cd/values.yaml
xaque208: This has been updated with your recommendations.  Good idea.

charts/argo-cd/templates/argocd-server/certificate.yaml
xaque208: I've updated to wrap all of the `privateKey` as suggested.

charts/argo-cd/Chart.yaml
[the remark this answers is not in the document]
pdrastil: Thanks - the changes is the release log for the given version.

pdrastil (approved):
LGTM

pdrastil:
@xaque208 Thanks for the changes. Please update the docs via `scripts/helm-docs.sh` so the build will pass and this can be merged.

PR #1454, opened 2022-09-15
feat(argo-workflows): Support array of strings in argo-workflows image pull secrets

CiprianAnton:
There are helm charts that expose image pull secrets as an array of strings. Argo Workflows helm exposes them as an array of objects {name: value}.

Add support to define image pull secrets as string, for better interoperability with other helm charts. This way, the same imagePullSecrets can be passed to multiple helms. Backwards compatibility is mantained.

Tested the behavior with running `helm template`. Compared the original and modified templates, and they are as expected.

The effect is shown bellow for the following input:
```
images:
  pullSecrets:
    - name: pullSecretNameAsObject
    - pullSecretNameAsString
```
image

Note on DCO:

Checklist:


mkilchhofer (commented):
Hi @CiprianAnton , thanks for your contribution.
If we are going to accept that change, we need to implement this in all our charts.
In my opinion (codeowner of argo-cd chart) this is a kind of special use case and will introduce unneccessary complexity. But lets listen for other opinions here 😄

yu-croco (dismissed):
Thank you for your contribution! LGTM

[updated]
I thought it's better to gather other maintainers' opinions, so let me revert the approve. 🙏

PR #1483, opened 2022-09-22
fix(argo-workflows): Set only used values on SSO configuration

yu-croco:
I fixed this https://github.com/argoproj/argo-helm/pull/1473/files#r976930509 . 🙇 

---

Signed-off-by: yu-croco <yu.croco@gmail.com>

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
@@ -119,7 +119,38 @@ data:
     workflowDefaults:
 {{ toYaml .Values.controller.workflowDefaults | indent 6 }}{{- end }}
     {{- with .Values.server.sso }}
-    sso: {{- toYaml . | nindent 6 }}
+    sso:
+      issuer: {{ .issuer }}
+      clientId:
+        name: {{ .clientId.name }}
+        key: {{ .clientId.key }}
+      clientSecret:
+        name: {{ .clientSecret.name }}
+        key: {{ .clientSecret.key }}
+      redirectUrl: {{ .redirectUrl }}
+      {{- if and (.rbac) (.rbac.enabled) }}
+      rbac:
+        enabled: {{ .rbac.enabled }}
+      {{- end }}
+      {{- if .scopes }}
+      scopes: {{ toYaml .scopes | nindent 8 }}
+      {{- end }}
+      {{- if .issuerAlias }}
+      issuerAlias: {{ .issuerAlias }}
+      {{- end }}
+      {{- if and (.sessionExpiry) (.sessionExpiry.duration) }}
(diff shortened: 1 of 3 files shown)


charts/argo-workflows/values.yaml
```diff
@@ -482,6 +482,11 @@ server:
     # redirectUrl: https://argo/oauth2/callback
     # rbac:
     #   enabled: true
+    ## When present, restricts secrets the server can read to a given list.
+    ## You can use it to restrict the server to only be able to access the
+    ## service account token secrets that are associated with service accounts
+    ## used for authorization.
+    #   secretWhitelist: []
```
yu-croco, line 489: I'm sorry, it is still used. 🙇 
ref: https://github.com/argoproj/argo-helm/pull/1473/files#r976930509

charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
```diff
@@ -119,7 +119,38 @@ data:
...
+      {{- end }}
+      {{- if .customGroupClaimName }}
+      customGroupClaimName: {{ .customGroupClaimName }}
+      {{- end }}
+      {{- if .UserInfoPath }}
+      UserInfoPath: {{ .UserInfoPath }}
+      {{- end }}
+      {{- if .insecureSkipVerify }}
+      insecureSkipVerify: {{ .insecureSkipVerify }}
+      {{- end }}
```
yu-croco, line 153: I set only valid values.
ref: ref: https://github.com/argoproj/argo-workflows/blob/v3.4.0/config/sso.go#L10-L24

vladlosev:
I wonder, what is the goal of this change? It does not enable any new features but on the downside, the there is an increased maintenance burden - the chart now has to keep up with  any changes in the SSOConfig struct.

yu-croco:
Hi @vladlosev , thank you for checking PR!
This is because argo-workflow-configmap does not accept unsupported values from v3.4.0, like below
ref: https://argoproj.github.io/argo-workflows/upgrading/#feat-fail-on-invalid-config-8295

> The workflow controller will error on start-up if incorrectly configured, rather than silently ignoring mis-configuration.
> Failed to register watch for controller config map: error unmarshaling JSON: while decoding JSON: json: unknown field \"args\"


We set values like the below on `workflow-controller-config-map.yaml`, but there are unsupported values like `secretWhitelist` for configmap.
For keeping backward compatibility, I set only valid values on `workflow-controller-config-map.yaml` .

```yaml
{{- with .Values.server.sso }}
sso: {{- toYaml . | nindent 6 }}
...
```

I'm wondering if there are better ways... 🤔

vladlosev (approved):
That's a good reason. Thanks for explaining!

Issue #1567, opened 2022-10-21
ApplicationSet CRD update for Gitlab Pull request generator


nbjohnson:
### Is your feature request related to a problem?

In Argocd v2.5.0 Gitlab was added to the Pull request generator. Currently the CRD does not have support for Gitlab in the matrix pull request generator, and it looks like none of the other pull request generator locations either.

### Related helm chart

argo-cd

### Describe the solution you'd like

For the matrix Gitlab pull request generator https://github.com/argoproj/argo-cd/blob/v2.5.0-rc3/manifests/crds/applicationset-crd.yaml#L2399 needs to be added at https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/crds/crd-applicationset.yaml#L2406

Looks like there are two other locations that need to have the Gitlab pull request generator added as well:

- https://github.com/argoproj/argo-cd/blob/v2.5.0-rc3/manifests/crds/applicationset-crd.yaml#L4629 -> https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/crds/crd-applicationset.yaml#L4558
- https://github.com/argoproj/argo-cd/blob/v2.5.0-rc3/manifests/crds/applicationset-crd.yaml#L5724 -> https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/crds/crd-applicationset.yaml#L5576

All instances of `pullRequest` in https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/crds/crd-applicationset.yaml should have Gitlab as an option

pdrastil:
@nbjohnson We do not create Helm charts with release candidate versions right now. This will be present when stable version 2.5.0 is officially released.

mkilchhofer:
Resolved via #1568

Issue #1537, opened 2022-10-10
Use https for dex server


psimms-r7:
### Is your feature request related to a problem?

Should we be using https for the dex server?

As per this issue
https://github.com/argoproj/argo-cd/issues/10841

### Related helm chart

argo-cd

### Describe the solution you'd like

By default use https for the dex server

pdrastil:
This will be relevant in 2.5 and there is already some initial implementation. 

- https://github.com/argoproj/argo-helm/pull/1477
- https://github.com/argoproj/argo-helm/pull/1474 - most of it is here and I will be pulling it out into separate PR

pdrastil:
Update - I've added support for Dex using HTTPS/TLS directly for 2.5.0 release. I will add support for custom Dex certs after that because it's not clear to me if these should be just internally facing certs using K8s endpoints or there is also need for externally facing certs.

mkilchhofer:
Resolved via #1568

PR #1477, opened 2022-09-21
feat(argo-cd): Support custom TLS certificates for Dex

pdrastil:
- Added option to use custom TLS certificate for Dex
- Enforced TLS strict mode when custom TLS is enabled

---

Note on DCO:

If the DCO action in the integration test fails, one or more of your commits are not signed off. Please click on the *Details* link next to the DCO action for instructions on how to resolve this.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Changes are automatically published when merged to `main`. They are not published on branches.

PR #1568, opened 2022-10-22
feat(argo-cd): Upgrade Argo CD to 2.5.0

pdrastil:
Resolves:
- https://github.com/argoproj/argo-helm/issues/1567
- https://github.com/argoproj/argo-helm/issues/1537

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v2.4.15
+appVersion: v2.5.0
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.7.0
+version: 5.8.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -22,6 +22,4 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Configuration sections configs.cm and configs.rbac"
-    - "[Deprecated]: Generic configuration via server.config"
-    - "[Deprecated]: Argo RBAC configuration via server.rbacConfig"
+    - "[Changed]: Upgraded Argo CD to 2.5.0"
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -16,6 +16,17 @@ Create dex name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Create Dex server endpoint
+*/}}
+{{- define "argo-cd.dex.server" -}}
+{{- $insecure := index .Values.configs.params "dexserver.disable.tls" | toString -}}
+{{- $scheme := (eq $insecure "true") | ternary "http" "https" -}}
+{{- $host := include "argo-cd.dex.fullname" . -}}
+{{- $port := int .Values.dex.servicePortHttp -}}
+{{- printf "%s://%s:%d" $scheme $host $port }}
+{{- end }}
+
+{{/*
 Create redis name and version as used by the chart label.
 */}}
 {{- define "argo-cd.redis.fullname" -}}
@@ -185,7 +196,7 @@ repo.server: "{{ include "argo-cd.repoServer.fullname" . }}:{{ .Values.repoServe
 redis.server: {{ . | quote }}
 {{- end }}
 {{- if .Values.dex.enabled }}
-server.dex.server: "http://{{ include "argo-cd.dex.fullname" . }}:{{ .Values.dex.servicePortHttp }}"
+server.dex.server: {{ include "argo-cd.dex.server" . }}
 {{- end }}
 {{- range $component := tuple "controller" "server" "reposerver" }}
 {{ $component }}.log.format: {{ $.Values.global.logging.format | quote }}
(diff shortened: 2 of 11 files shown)


charts/argo-cd/templates/_helpers.tpl
```diff
@@ -40,6 +40,17 @@ Create dex name and version as used by the chart label.
 {{- printf "%s-%s" (include "argo-cd.fullname" .) .Values.dex.name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{/*
+Create Dex server endpoint
+*/}}
+{{- define "argo-cd.dex.server" -}}
```
pdrastil, line 46: Support for new HTTP / HTTPS for dex server. Tested via OIDC provider config on RC3

charts/argo-cd/templates/argocd-repo-server/deployment.yaml
```diff
@@ -183,44 +213,42 @@ spec:
...
           name: helm-working-dir
         - mountPath: /home/argocd/cmp-server/plugins
           name: plugins
+        - mountPath: /tmp
+          name: tmp
         ports:
         - name: repo-server
           containerPort: {{ .Values.repoServer.containerPort }}
           protocol: TCP
-        {{ if .Values.repoServer.metrics.enabled }}
```
pdrastil, line 208: Removed all these guards. Metrics ports are not related to configured metrics Service and can be used via other means like Prometheus service discovery. They are also used by some controllers for liveness / readiness probes.

charts/argo-cd/templates/argocd-repo-server/deployment.yaml
```diff
@@ -183,44 +213,42 @@ spec:
         {{- if .Values.repoServer.volumeMounts }}
           {{- toYaml .Values.repoServer.volumeMounts | nindent 8 }}
         {{- end }}
-        - mountPath: /app/config/gpg/source
-          name: gpg-keys
-        - mountPath: /app/config/gpg/keys
-          name: gpg-keyring
-        {{- if .Values.configs.knownHosts }}
```
pdrastil, line 190: Not an optional volume. All optional volumes are marked as such via `optional: true`

pdrastil:
Synced and tested on RC3. Still waiting for confirmation if new cmd-params for applicationset will be cherry picked to 2.5.0

charts/argo-cd/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v2.4.15
+appVersion: v2.5.0-rc3
```
pdrastil, line 2: To be changed to 2.5.0 when released

karlschriek:
Hey, thanks a lot for getting this up so early. We are pretty anxious to start using 2.5.0 because of the ability to roll out Applications to namespaces other than `argocd`. I am giving this chart a spin today to see if I can get it to work correctly. 

I did notice however that the CRD rollouts in the chart are still `Namespaced`. Shouldn't they have a `Cluster` scope for the above feature to work correctly?

pdrastil:
@karlschriek Hi - you are welcome. CRDs are always global resource in the cluster. As far as I know inside CRD schema can be namespaced (to allow creation of resources inside each namespace) or global (which means that they behave similar like ClusterRoles). This chart just installs the CRDs as part of it installation, namespaced schema ensures that it can be used anywhere.

karlschriek:
Right! Sorry yeah my bad, Cluster CRDs are meant for CRs that don't get
rolled out to a namespace at all but are rolled out as a cluster-wide
resource, has nothing to do with where those types of CRs can be deploy.

Anyway, have rolled out and it looks like it is working. Wil post back here
if I come across any issues related to the chart specifically.

charts/argo-cd/Chart.yaml
jmeridth: it has been done :smile: https://github.com/argoproj/argo-cd/releases/tag/v2.5.0

pdrastil: @jmeridth Builiding

jmeridth (approved):
This LGTM but would like another approval before merge

hairmare (approved):
I tried this on a recent kind v1.24.0 and didn't run into any issues using the out-of-the-box values.

This is just just a drive-by LGTM since i'm not a maintainer though.

mkilchhofer (commented):
Additional to the comment, we should cleanup the default securityContexts as upstream already has default (hardened) security context defined.
And release 2.5 now adds 
```yaml
          seccompProfile:
            type: RuntimeDefault
```
Also the TLS stuff for dex is missing. But I assume you want to do this outside the 2.5 PR and test it again via:
- https://github.com/argoproj/argo-helm/pull/1477

But I did not find a blocker. So I am okay releasing this one and fix the

pdrastil:
@mkilchhofer - See following

Ad security contexts - already have local branch with this because it needs more work on all components.
Dex TLS - The protocol now uses HTTPS by default, certs are self-signed. Custom certs will follow here https://github.com/argoproj/argo-helm/pull/1477
Probes - I'll check in next iteration - they were still present in RC3

bobertrublik:
Thank you for your work! The argocd-server serviceaccount still needs more permissions when apps are deployed to a different namespace (as per https://github.com/argoproj/argo-cd/pull/9755#issuecomment-1212915993).
I'm however not sure what the best practices are here. Can we grant the permissions to the ClusterRole?

pdrastil:
I believe it should be ClusterRole as you are using different namespace. I will take a look what needs to be patched and add it as another fix.

Issue #1560, opened 2022-10-17
[argocd-redis-ha] seccompProfile is forbidden on k8s version 1.21


ayoub-elbachiri:
### Describe the bug

k8s version: 1.21
argocd helm chart version: 5.6.0

After deploying the latest chart of argocd, redis-ha-server statefulset and redis-ha-proxy are not deploying cause of the seccompProfile injected by default from the argocd chart,

I tried to edit the statefulset manually and delete the seccompProfile from the specs, the pods poped up

### Related helm chart

argo-cd

### Helm chart version

5.6.0

### To Reproduce

1. deploy argocd helm chart on k8s

### Expected behavior

  kubectl describe sts
  Warning  FailedCreate  16s (x13 over 36s)  statefulset-controller  create Pod argocd-redis-ha-server-0 in StatefulSet argocd-redis-ha-server failed error: pods "argocd-redis-ha-server-0" is forbidden: PodSecurityPolicy: unable to admit pod: [pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/config-init]: Forbidden: seccomp may not be set pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/redis]: Forbidden: seccomp may not be set pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/sentinel]: Forbidden: seccomp may not be set pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/split-brain-fix]: Forbidden: seccomp may not be set]

pdrastil:
Hi @ayoub-elbachiri - this is actually part of redis-ha subchart [here](https://github.com/DandyDeveloper/charts/blob/30611121e114fea4eebf1202021fbcc6bd1ee1ef/charts/redis-ha/values.yaml#L169) that is used by argo-cd.

You can try to unset the value like this:

```yaml
redis-ha:
  containerSecurityContext:
    seccompProfile: null
```

ayoub-elbachiri:
> 


> Hi @ayoub-elbachiri - this is actually part of redis-ha subchart [here](https://github.com/DandyDeveloper/charts/blob/30611121e114fea4eebf1202021fbcc6bd1ee1ef/charts/redis-ha/values.yaml#L169) that is used by argo-cd.
> 
> You can try to unset the value like this:
> 
> ```yaml
> redis-ha:
>   containerSecurityContext:
>     seccompProfile: null
> ```

[...]

Issue #1494, opened 2022-09-27
Allow for securityContext to be configured on argocd-extensions container in server Pod


congiv-VG:
### Related helm chart

argo-cd

### Describe the solution you'd like

It would be great if the container level securityContext could be configured in the same way as the other workloads in the chart. Something like:

```yaml
server:
  extensions:
    securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop: [ "ALL" ]
```

pdrastil:
Will be done for 6.x release where I'm trying to cleanup & consolidate all things

pdrastil:
Resolved via https://github.com/argoproj/argo-helm/pull/1579

PR #1579, opened 2022-10-26
feat(argo-cd): Set container security contexts

pdrastil:
Resolves: 
- https://github.com/argoproj/argo-helm/issues/1494
- https://github.com/argoproj/argo-helm/issues/1560

Container security context have been placed into same place across all components
Minimum supported version is 1.22 based on https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#supported-versions

---

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -1,8 +1,9 @@
 apiVersion: v2
 appVersion: v2.5.0
+kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.9.1
+version: 5.10.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -22,4 +23,8 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Removed]: Liveness probe for application controller"
+    - "[Security]: Use recommended container security contexts by default"
+    - "[Added]: Container security context for server UI extensions sidecar"
+    - "[Fixed]: Redis metrics sidecar now uses correct configuration option"
+    - "[Removed]: ApplicationSet securityContext in favor of global.securityContext"
+    - "[Removed]: Notification securityContext in favor of global.securityContext"
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -558,13 +558,16 @@ controller:
   podLabels: {}
 
   # -- Application controller container-level security context
+  # @default -- See [values.yaml]
   containerSecurityContext:
-    {}
-    # capabilities:
-    #   drop:
-    #     - all
-    # readOnlyRootFilesystem: true
-    # runAsNonRoot: true
+    runAsNonRoot: true
+    readOnlyRootFilesystem: true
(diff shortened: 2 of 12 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -1,5 +1,6 @@
 apiVersion: v2
 appVersion: v2.5.0
+kubeVersion: ">=1.22.0"
```
pdrastil, line 3: Added info about minimum version of K8s cluster

mkilchhofer: Hmmm, I am unsure wether we should add this constraint or not. Redis-ha is an optional feature

pdrastil: Yes but `seccomp` profile is 1.22 feature.

charts/argo-cd/values.yaml
```diff
@@ -1032,6 +1040,22 @@ redis:
   # -- Labels to be added to the Redis server pods
   podLabels: {}
 
+  # -- Redis pod level security context
+  # @default -- See [values.yaml]
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 999
+    seccompProfile:
```
pdrastil, line 1048: This is available from kubeVersion >= 1.22

charts/argo-cd/Chart.yaml
mubarak-j: This will exclude many users like me who don't use redid-ha feature and run k8s 1.21.

mkilchhofer: I am unsure what path we should follow. Upstream 2.5 manifests (kustomize) also set seccomp and therefore ensures that customers run k8s >= 1.22

Maybe we should follow that also. Otherwise our charts supports a scenario which is not supported from the project itself.

charts/argo-cd/Chart.yaml
```diff
@@ -22,4 +23,8 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Fixed]: Type conversion for ConfigMap values"
+    - "[Security]: Use recommended container security contexts by default"
+    - "[Added]: Container security context for server UI extensions sidecar"
+    - "[Fixed]: Redis metrics sidecar now uses correct configuration option"
+    - "[Removed]: ApplicationSet securityContext in favor of global.securityContext"
```
mkilchhofer, line 29: Why do we want to remove the ability to customize pod level SecurityContext per component? Maybe some addons require special privileges, idk?

charts/argo-cd/Chart.yaml
pdrastil: @mubarak-j Not really - Azure and GKE already allow 1.22 as a minimum. For AWS the EOL is February 2023.

charts/argo-cd/values.yaml
```diff
@@ -814,6 +817,7 @@ dex:
   # @default -- `[]` (defaults to global.imagePullSecrets)
   imagePullSecrets: []
 
+  # Argo CD init image that crates Dex config
```
mkilchhofer, line 820:
```suggestion
  # Argo CD init image that creates Dex config
```

mkilchhofer (approved):

charts/argo-cd/values.yaml
pdrastil: Fixed

mkilchhofer (approved):

Issue #1572, opened 2022-10-24
[argo-workflows] Missing RBAC to list *v1.ServiceAccount


clementlecorre:
### Describe the bug

With a clean install I have RBAC error on my argo src

```
E1024 12:14:44.339841       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.3/tools/cache/reflector.go:167: Failed to watch *v1.ServiceAccount: failed to list *v1.ServiceAccount: serviceaccounts is forbidden: User "system:serviceaccount:my-namespace:argo-server" cannot list resource "serviceaccounts" in API group "" at the cluster scope
```

### Related helm chart

argo-workflows

### Helm chart version

0.20.3

### To Reproduce

install argo-workflows chart

### Expected behavior

n/a

yu-croco:
Hi @clementlecorre , will you please tell me the values.yaml  that you encountered this issue? 🙋

clementlecorre:
hi @yu-croco 
It's work with this fix : https://github.com/argoproj/argo-helm/pull/1586

```
  sso:
    rbac:
      enabled: true
    issuer: https://accounts.google.com
    clientId:
      name: argo-sso-config
      key: client-id
    clientSecret:
      name: argo-sso-config
      key: client-secret
```

yu-croco:
Thank you for your reply, @clementlecorre .
We encountered another error by https://github.com/argoproj/argo-helm/issues/1638, so we revert it on https://github.com/argoproj/argo-helm/pull/1641.
Then, let us discuss how to fix the original issue. 🙋 

ref: https://github.com/argoproj/argo-helm/issues/1638#issuecomment-1313444123

cc: @vladlosev

yu-croco:
Hi @clementlecorre ,
it's just a confirmation, did you also set `server.extraArgs: ["--auth-mode=sso"]` ([here](https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/values.yaml#L421-L427)) as well ?

clementlecorre:
@yu-croco Yes

vladlosev:
The error message aside, does the authentication work?

yu-croco:
I checked with argo-chart version `0.20.8` .
I couldn't get error, so I am wondering what I missed... 😇 

values.yaml

```yaml
server:
  extraArgs:
    - --auth-mode=sso
  sso:
    rbac:
      enabled: true
    issuer: https://accounts.google.com
    clientId:
[...]

PR #1586, opened 2022-10-26
fix(argo-workflows): serviceaccount rbac when sso is enabled

nxf5025:
Fixes: #1572 

Removed the sso.rbac map as it breaks the sso integration. Also if the sso block is used we should include serviceaccount in our clusterrole instead of being managed by another setting.

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.4.2
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.20.4
+version: 0.20.5
 icon: https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 sources:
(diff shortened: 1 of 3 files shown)


vladlosev:
This change grants access to service accounts even when it's not needed. Not every organization wants to give every app access unnecessary access. As an alternative, you can simply remove the `rbac` element  when using `sso` to generate the config map: https://github.com/argoproj/argo-helm/blob/aa669028b899eef65415f9dc730815b4bef9b7a6/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml#L134.

vladlosev (approved):
LGTM

yu-croco (approved):

whereismyjetpack:
I'm not certain why, but I'm pretty sure this change has the inverse effect in my environment, when there isn't an rbac stanza in the configmap, all users have default rbac, which appears to be the sa that argo is using, i had to manually add ```rbac.enabled``` in the configmap to get the old behavior

agilgur5:
For any future readers, this was reverted in #1641 ([comment](https://github.com/argoproj/argo-helm/pull/1641#discussion_r1021363307))

Issue #1587, opened 2022-10-26
ArgoCD upgrade fails


powellchristoph:
### Describe the bug

Upgrade to latest release fails.
```
helm upgrade argocd argo/argo-cd -n argocd -f secrets://values.yaml --version 5.8.2

Error: UPGRADE FAILED: template: argo-cd/templates/argocd-server/clusterrole.yaml:30:9: executing "argo-cd/templates/argocd-server/clusterrole.yaml" at <eq (index (coalesce .Values.server.config .Values.configs.cm) "exec.enabled") "true">: error calling eq: incompatible types for comparison
```

It appears to not be properly referencing older values. https://github.com/argoproj/argo-helm/pull/1528

I fixed it by moving the relevant `Values.server.config` sections to `Values.configs`
```yaml
server:
  config:
    url: https://argocd.example.net
    oidc.config: <insert config>
  rbacConfig:
    policy.default: role:''
    scopes: "[email,groups]"
    policy.csv: |
      g, example-admin, role:admin
```
to
```yaml
configs:
  cm:
    url: https://argocd.example.net
    oidc.config: <insert config>
  rbac:
    policy.default: role:''
    scopes: "[email,groups]"
    policy.csv: |
      g, example-admin, role:admin 
```

### Related helm chart

argo-cd

### Helm chart version

5.8.2

### To Reproduce

1. Have existing helm install prior to 5.7.0 with the relevant `server.config` values set.
2. Execute a helm upgrade

### Expected behavior

Upgraded Argocd.

pdrastil:
Hi @powellchristoph, I've tested with this values on 5.8.2 but I was able to reproduce only with `exec.enabled: true` that is not present in your sample. Notice that `true` is boolean and ConfigMap is using strings.

```
configs:
  cm:
    url: https://argocd.example.net
    oidc.config: ""
    exec.enabled: true
  rbac:
    policy.default: role:''
    scopes: "[email,groups]"
    policy.csv: |
      g, example-admin, role:admin
```

powellchristoph:
@pdrastil I understand. That said, I did not not have `exec.enabled` included in my `values.yaml`.

pdrastil:
Really not sure what is the issue here... only difference I see is the use of `secrets://`

```sh
❯ helm install argocd argo/argo-cd -n argocd -f extra.yaml --version 5.4.3
NAME: argocd
LAST DEPLOYED: Wed Oct 26 21:10:33 2022
[...]

PR #1594, opened 2022-10-29
fix(argo-cd): Type conversion for ConfigMaps values

pdrastil:
Resolves: https://github.com/argoproj/argo-helm/issues/1587

PR improves type conversion when user passes incorrect value types via `--set` instead of `--set-string` for config.cm and config.params sections.

---

Note on DCO:

Checklist:


charts/argo-cd/values.yaml
```diff
@@ -110,21 +110,21 @@ configs:
 
     # -- Enable logs RBAC enforcement
     ## Ref: https://argo-cd.readthedocs.io/en/latest/operator-manual/upgrading/2.3-2.4/#enable-logs-rbac-enforcement
-    server.rbac.log.enforce.enable: "false"
+    server.rbac.log.enforce.enable: false
 
     # -- Enable exec feature in Argo UI
     ## Ref: https://argo-cd.readthedocs.io/en/latest/operator-manual/rbac/#exec-resource
-    exec.enabled: "false"
+    exec.enabled: false
```
pdrastil, line 117: Intentionally changed to type expected by humans.

jmeridth (approved):

mkilchhofer:
Why we do not face this problem in the past? Its according to documentation that the value needs to be type string.

pdrastil:
@mkilchhofer I don't know - only change move from server to configs without any changes in the content of configuration section. As people are migrating some of them pass `--set` for these values instead of `--set-string` which breaks them. This is just my attempt to make it fool-proof.

Issue #1592, opened 2022-10-27
Add argo-helm argo-cd implementation of namespace-install.yaml


racker7:
### Is your feature request related to a problem?

no

### Related helm chart

argo-cd

### Describe the solution you'd like

argo-helm/argo-cd uses this to create full implementation of ArgoCD:
https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml

I would like to be able to create additional ArgoCD instances on same cluster using this with helm:
https://github.com/argoproj/argo-cd/blob/master/manifests/namespace-install.yaml

The main differentiators of this alternate install method are skipping setup of 3 CRDs and 2 Cluster Roles.  This is dependent on having a full install paired alongside it.

The goal is to have multiple ArgoCD instances to separate admin usage from application teams, for security and performance isolation.

Thank you for your consideration.

### Describe alternatives you've considered

Alternative is using non-helm method to install the namespace-install.  This means we don't have all the objects managed by helm, and its less consistent/supportable.

racker7:
Update: I see the latest version of argo-helm has this option: crds.install
Perhaps this accomplishes the goal.  I'll check with the team about upgrading to latest.

pdrastil:
For cluster roles thare are these two switches:

```yam
controller:
  clusterAdminAccess:
    enabled: false

server:
  clusterAdminAccess:
    enabled: false
```

pdrastil:
Mentioned PR above will change the config on all components to single key.

```yaml
global:
  rbac:
    clusterRoles: true
```

racker7:
Thank you very much.  I'll give this a try, hopefully in the next week or so.

racker7:
hi @pdrastil   can I ask another question please?
I am not sure what roles this creates exactly.
Do I want to enable this on the primary admin argo instance or the secondary application argo instance, or both?
global.rbac.aggregatedRoles / createAggregateRoles

racker7:
ok i figured out that you enable aggregate roles in the secondary application argo instance and not the main primary admin argo instance.

the result is you get CRDs and ClusterRoles in the main instance (app-controller,server)
in the secondary instances you do not get CRDs, you get ClusterRoles that are 'aggregated' (admin/edit/view)

the CRDs appear to be accessible to the secondary instance from initial testing.

ty for your help.

racker7:
update.  we decided one argocd per cluster is simpler to support as the multiple is not well documented.  ty for all your help.

PR #1597, opened 2022-10-31
chore(argo-cd): Consolidate cluster-wide RBAC

pdrastil:
Argo CD 2.5.0 introduced feature for managing applications in other namespaces and users are starting to look for a ways to disable cluster roles to have namespaced installation. This PR consolidates cluster-wide RBAC under single configuration key.

Closes: https://github.com/argoproj/argo-helm/issues/1592

---

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -362,7 +362,8 @@ NAME: my-release
 | crds.annotations | object | `{}` | Annotations to be added to all CRDs |
 | crds.install | bool | `true` | Install and upgrade CRDs |
 | crds.keep | bool | `true` | Keep CRDs on chart uninstall |
-| createAggregateRoles | bool | `false` | Create clusterroles that extend existing clusterroles to interact with argo-cd crds |
+| createAggregateRoles | bool | `false` | Create aggregated roles that extend existing cluster roles to interact with argo-cd resources |
+| createClusterRoles | bool | `true` | Create cluster roles for cluster-wide installation. |
 | extraObjects | list | `[]` | Array of extra K8s manifests to deploy |
 | fullnameOverride | string | `""` | String to fully override `"argo-cd.fullname"` |
 | kubeVersionOverride | string | `""` | Override the Kubernetes version, which is used to evaluate certain manifests |
@@ -451,7 +452,6 @@ NAME: my-release
 |-----|------|---------|-------------|
 | controller.affinity | object | `{}` | Assign custom [affinity] rules to the deployment |
 | controller.args | object | `{}` | DEPRECATED - Application controller commandline flags |
-| controller.clusterAdminAccess.enabled | bool | `true` | Enable RBAC for local cluster deployments |
 | controller.clusterRoleRules.enabled | bool | `false` | Enable custom rules for the application controller's ClusterRole resource |
 | controller.clusterRoleRules.rules | list | `[]` | List of custom rules for the application controller's ClusterRole resource |
 | controller.containerPort | int | `8082` | Application controller listening port |
@@ -529,7 +529,6 @@ NAME: my-release
 | repoServer.certificateSecret.enabled | bool | `false` | Create argocd-repo-server-tls secret |
 | repoServer.certificateSecret.key | string | `""` | Certificate private key |
 | repoServer.certificateSecret.labels | object | `{}` | Labels to be added to argocd-repo-server-tls secret |
-| repoServer.clusterAdminAccess.enabled | bool | `false` | Enable RBAC for local cluster deployments |
 | repoServer.clusterRoleRules.enabled | bool | `false` | Enable custom rules for the Repo server's Cluster Role resource |
 | repoServer.clusterRoleRules.rules | list | `[]` | List of custom rules for the Repo server's Cluster Role resource |
 | repoServer.containerPort | int | `8081` | Configures the repo server port |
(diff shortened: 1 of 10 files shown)


charts/argo-cd/README.md
```diff
@@ -345,7 +345,6 @@ NAME: my-release
 | crds.annotations | object | `{}` | Annotations to be added to all CRDs |
 | crds.install | bool | `true` | Install and upgrade CRDs |
 | crds.keep | bool | `true` | Keep CRDs on chart uninstall |
-| createAggregateRoles | bool | `false` | Create clusterroles that extend existing clusterroles to interact with argo-cd crds |
```
mkilchhofer, line 348: I would not move that option to follow what our other charts implement (events and rollouts).

Also Bitnami charts offers this exact option

mkilchhofer (changes requested):

charts/argo-cd/README.md
pdrastil: Bitnami have this, however placing it here would make sense if you run umbrella with other pieces such as events, rollouts. If adjusted there you can have 1 switch for all the charts.

jmeridth: @mkilchhofer are you good with @pdrastil's response or still against the change? I assume latter.

pdrastil:
@mkilchhofer @jmeridth Changed as requested.

charts/argo-cd/values.yaml
```diff
@@ -23,9 +23,12 @@ apiVersionOverrides:
   # -- String to override apiVersion of pod disruption budgets rendered by this helm chart
   pdb: "" # policy/v1
 
-# -- Create clusterroles that extend existing clusterroles to interact with argo-cd crds
+# -- Create agggregated roles that extend existing cluster roles to interact with argo-cd resources
```
mkilchhofer, line 26: typo :-)

agggregated -> aggregated

pdrastil: Arrrrrgh - Fixed :D

charts/argo-cd/templates/argocd-repo-server/clusterrole.yaml
```diff
@@ -1,23 +1,24 @@
-{{- if and .Values.repoServer.serviceAccount.create .Values.repoServer.clusterAdminAccess.enabled }}
```
mkilchhofer, line 1: Strange, only reposerver had this check of `serviceAccount.create`

charts/argo-cd/templates/argocd-repo-server/clusterrolebinding.yaml
```diff
@@ -1,16 +1,17 @@
-{{- if and .Values.repoServer.serviceAccount.create .Values.repoServer.clusterAdminAccess.enabled }}
```
mkilchhofer, line 1: Same here

mkilchhofer (approved):
Just a small interesting finding, but I am okay with it

charts/argo-cd/templates/NOTES.txt
```diff
@@ -52,6 +52,18 @@ DEPRECATED option configs.gpgKeys - Use config.gpg.keys
 {{- if .Values.configs.gpgKeysAnnotations }}
 DEPRECATED option configs.gpgKeysAnnotations - Use config.gpg.annotations
 {{- end }}
+{{- if hasKey .Values "createAggregateRoles" }}
+DEPRECATED option createAggregateRoles - Use global.rbac.aggregatedRoles
```
faizsiddiqui, line 56: Hey @pdrastil 

I am seeing this deprecation warning when upgrading my chart but the change is not done in the default values.yaml, just checking if this change is a mistake or there is future plan to move to global setting?

pdrastil: Hi @faizsiddiqui - you can ignore this. It was originally placed into global with other cluster role rbac. This is leftover that shouldn't be there. Thanks for a find. I'll update the docs / notes.

tyrken:
Heh @pdrastil @mkilchhofer the default for `repoServer.clusterAdminAccess.enabled` was `false` as it's only needed if you have unusual/demanding cm-plugins that require full admin-level K8s access *in themselves* rather than via their output to argocd, see #919 and https://github.com/argoproj/argo-helm/pull/919#pullrequestreview-756577440 .  We don't and it's exposing god-mode to a few more externally sourced plugins than I'd prefer.

Can you please revert this or enable non-deprecated individual control of the repoServer cluster role/binding in future.

pdrastil:
Hi @tyrken - this was done because of introduction of using Application resource in any namespace with 2.5.x release (https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/). Option now reflects upstream manifests for cluster-install kustomization found here (https://github.com/argoproj/argo-cd/blob/master/manifests/cluster-install/kustomization.yaml ) where Argo needs to have access to any namespace to make this work. When option `createClusterRoles: false` is used, Argo will go to namespace installation. You can always provide extra RBAC via `extraObjects` for service accounts that needs it but IMHO mixing ClusterRoles with namespace install seems like a code smell to me.

tyrken:
Thanks for the quick response @pdrastil .  Unfortunately I can't work out where the `kubectl apply -f examples/k8s-rbac/argocd-server-applications/` is meant to be run from in the docs in <https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/#adapt-kubernetes-rbac>, but when I run:

```
kustomize build https://github.com/argoproj/argo-cd/manifests/cluster-install
```

...I get a bunch of yaml that only contains two ClusterRole/Binding sets, one for `argocd-application-controller` and one for `argocd-server`.  I.e. nothing for `argocd-repo-server`, so I think that isn't needed in this helm-chart.

pdrastil:
Hi @tyrken - you are right that upstream do not have ClusterRole on repo-server. This was added for some Helm plugin by https://github.com/argoproj/argo-helm/pull/919 . I'm not 100% sure if this should be really needed. For the examples referenced in doc I have a same problem and will ask @crenshaw-dev where this really is or if it's implemented.

pdrastil:
I agree that this should be improved / fixed. Either we should remove the role as there are other means how to provide this or make it empty without wildcard rules and let the user provide extra RBAC from the chart if it's really needed.

charts/argo-cd/templates/NOTES.txt
pdrastil: Resolved via https://github.com/argoproj/argo-helm/pull/1830

pdrastil:
@tyrken - FYI found the files mentioned in docs in this PR - https://github.com/argoproj/argo-cd/pull/12341

Issue #1578, opened 2022-10-26
Make revisionHistoryLimit configurable


maartengo:
### Is your feature request related to a problem?

revisionHistoryLimit is currently hardcoded in the chart to always be `5`. I don't really care about the revision history since I use Argo to revert versions where needed. At the moment it only clutters my Application view. Although this is less of a problem with the 'grouping' feature, it would still be nice to have a clearer overview by default.

### Related helm chart

argo-cd

### Describe the solution you'd like

`revisionHistoryLimit` as a setting in the `values.yaml` of the charts.

### Additional context

I use Argo CD to rollout argo-cd. I don't use the other argo-* charts, but I suppose those could also benefit from the same change.

PR #1599, opened 2022-10-31
feat(argo-cd): Add revisionHistoryLimit

pdrastil:
Closes: https://github.com/argoproj/argo-helm/issues/1578

---

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -46,6 +46,9 @@ global:
   additionalLabels: {}
     # app: argo-cd
 
+  # -- Number of old deployment ReplicaSets to retain. The rest will be garbage collected.
+  revisionHistoryLimit: 3
+
   # Default image used by all components
   image:
     # -- If defined, a repository applied to all Argo CD deployments
@@ -2838,10 +2841,6 @@ notifications:
       # -- Enable slack bot
       ## You have to set secret.notifiers.slack.signingSecret
       enabled: false
-
-      # -- The deployment strategy to use to replace existing pods with new ones
-      updateStrategy:
-        type: Recreate
 
       ## Slack bot Pod Disruption Budget
       ## Ref: https://kubernetes.io/docs/tasks/run-application/configure-pdb/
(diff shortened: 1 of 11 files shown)


charts/argo-cd/values.yaml
```diff
@@ -2839,10 +2842,6 @@ notifications:
       ## You have to set secret.notifiers.slack.signingSecret
       enabled: false
 
-      # -- The deployment strategy to use to replace existing pods with new ones
```
pdrastil, line 2842: Removed as SlackBot must be always recreated first.

charts/argo-cd/values.yaml
```diff
@@ -46,6 +46,9 @@ global:
   additionalLabels: {}
     # app: argo-cd
 
+  # -- Number of old deployment ReplicaSets to retain. The rest will be garbage collected.
+  revisionHistoryLimit: 3
```
mkilchhofer, line 50: Lets set it to 5 again so we can also update the STS of the controller.

pdrastil: STS does not support this - it's there from historic reasons because it was Deployment 1 year ago.

pdrastil: Also if you try to remove history limit from STS you get following:

```sh
The StatefulSet "argocd-application-controller" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

Only way to upgrade is to run these and I wanted to keep this for 6.x release when we remove old things.

```sh
kubectl delete argocd-application-controller
helm upgrade argocd
```

mkilchhofer: Ah you are right 👍

mkilchhofer (approved):
Lgtm

charts/argo-cd/values.yaml
mbevc1: Any reason why we're not matching 5 here as a default?

pdrastil: I've reduced this to 3 to make UI less cluttered when Argo CD is watching itself. Users can adjust amount of replicas so if you want more you can adjust it.

mbevc1: I'm happy with this, but I'd try to align to previous or upstream defaults if we can.

PR #1648, opened 2022-11-16
feat(argo-cd): Add Volume Pesistence support  for argocd-repo-server

### LINKED ISSUE #438: Argo CD: PersistentVolumeClaims for volumes

I'm trying to mount remote storage to the argocd repo-server. I can provide **volumeMounts** list as well as **volumes** list, but I can't create PVC for those volumes using helm chart.

volumes.yaml could contain a definition of StorageClass name for PVC. When provided, the volume mount would be created on top of this PVC.

Alternatively, documentation could be extended with information about using PVC and manual creation of PVC.

**Additional context**

I can see, that similar feature is already implemented in redis-ha: https://github.com/argoproj/argo-helm/blob/d7da8e863f30a6975b89a1bba34855d83bed59e3/charts/argo-cd/charts/redis-ha/templates/redis-ha-statefulset.yaml#L285

Also, this request is only about repo-server, but similar could be requested for Redis and other components.

If this looks like something useful, I can provide PR.

arielly-parussulo:
- allow argocd-repo-server to be deployed as a StatefulSet.
- allow the creation of Persistent Volumes for argocd-repo-server.

Checklist:


charts/argo-cd/templates/argocd-repo-server/deployment.yaml
```diff
@@ -1,10 +1,17 @@
 apiVersion: apps/v1
+{{- if .Values.repoServer.enableStatefulSet }}
+kind: StatefulSet
```
pdrastil, line 3: I'm little bit worried about this implementation because repo server is stateless, not stateful and typically runs with HPA (we run 30+ replicas). This feature will break the HPA that targets Deployment + will create large amount of PVCs.

pdrastil:
Hi @arielly-parussulo thanks for contribution. 

I personally believe that persistence should be more complex feature. For reason mentioned above the repo server would ideally use 1 shared PVC with ReadWriteMultiple mode so all repositories can share already downloaded resources instead of trying to create lots of PVCs that will eventually converge to the same content. I believe 1 disk per replica would work ok with NFS backend and only a few replicas.

I can imagine that this should have at least following:
1) global.persistence with defaults for new PVCs
2) component level overrides (mapping to volumes)
3) option to choose between PVC (created, existing), in-memory emptyDir and ephemeral emptyDir
4) must be compatible with current HPAs

I think this might be good base but feature should be extended.

arielly-parussulo:
> Hi @arielly-parussulo thanks for contribution.
> 
> I personally believe that persistence should be more complex feature. For reason mentioned above the repo server would ideally use 1 shared PVC with ReadWriteMultiple mode so all repositories can share already downloaded resources instead of trying to create lots of PVCs that will eventually converge to the same content. I believe 1 disk per replica would work ok with NFS backend and only a few replicas.
> 
> I can imagine that this should have at least following:
> 
> 1. global.persistence with defaults for new PVCs
> 2. component level overrides (mapping to volumes)
> 3. option to choose between PVC (created, existing), in-memory emptyDir and ephemeral emptyDir
> 4. must be compatible with current HPAs
> 
> I think this might be good base but feature should be extended.

Cool! I think I can add some of these features in this PR. I've added the StatefulSet feature because I saw some people mentioning that in [this issue](https://github.com/argoproj/argo-helm/issues/438) and we ended up using it in my company as we have less replicas and a monorepo that ends up causing a DiskPressure issue in our pods. So I still think that it could be an option in the Helm Chart.
But I agree with you about the other features. I will try to improve this PR to add more persistency features for argocd-repo-server.

Gianluca755:
Sorry for asking @pdrastil, is it possible to have feature even if it's still in beta? @arielly-parussulo haven't pushed in 3 weeks. I can contribute if needed. Thanks

pierluigilenoci:
Any news about this?

zswanson:
> For reason mentioned above the repo server would ideally use 1 shared PVC with ReadWriteMultiple mode 

Just to note, ReadWriteMany isn't well supported by cloud vendors when pods are on different nodes. Neither GKE nor EKS allow ReadWriteMany when you're using PD or EBS as the volume type.

pierluigilenoci:
@pdrastil using `ReadWriteMany` necessarily imposes the use of some [NFS](https://it.wikipedia.org/wiki/Network_File_System) (for example, [EFS from AWS](https://aws.amazon.com/efs/)) with a significant reduction in performance (for example, in Azure, [SMB](https://learn.microsoft.com/en-us/azure/storage/files/files-nfs-protocol) is used, which has embarrassing performance) as well as essential limitations in cases of clusters that operate multi-AZ.
One day, it will be a concrete option, but today it is only a good theoretical idea that collides with the limits of the various cloud providers.

clement94310:
hello
we also need a pvc on our repo server does someone know if this PR will be merge soon ?

bschaeffer:
Whats the deal with this feature?

charts/argo-cd/templates/argocd-repo-server/deployment.yaml
bschaeffer: Just... don't enable statefulset mode?

yann-soubeyrand (commented):
Sorry for the noise, I did not find a way to remove myself from requested reviewers other than by doing a review :worried:

Daverkex:
> Hi @arielly-parussulo thanks for contribution.
> 
> I personally believe that persistence should be more complex feature. For reason mentioned above the repo server would ideally use 1 shared PVC with ReadWriteMultiple mode so all repositories can share already downloaded resources instead of trying to create lots of PVCs that will eventually converge to the same content. I believe 1 disk per replica would work ok with NFS backend and only a few replicas.
> 
> I can imagine that this should have at least following:
> 
>     1. global.persistence with defaults for new PVCs
> 
>     2. component level overrides (mapping to volumes)
> 
>     3. option to choose between PVC (created, existing), in-memory emptyDir and ephemeral emptyDir
> 
>     4. must be compatible with current HPAs
> 
> 
> I think this might be good base but feature should be extended.

That is a very interesting point. But in my honest opinion, we are overengineering here. I mean, currently this is the current behavior. In a deploy, every pod copies “randomly” the repo needed. This feature is an improvement to mitigate the **copy per pod**. Every time the pod is migrated to another node, the process starts again. With a PVC, this part is solved.
My only concern is how the repo server deletes old or unused clones to avoid an infinite disk in an infinite time.
Step 2 could be the improvement you suggest: a shared system for all the pods to not only mitigate the copy per pod but also limit the **total copy combined**. Several options here: an S3 backend, a ReadWriteMany...

This PR could be step one, then a more sophisticated approach for step two.

jmeridth:
Thanks @arielly-parussulo for the contribution, and thanks to everyone who kept this thread alive. Closing this one out with an explanation, since the landscape has changed since 2022:

**Persistent volumes for the repo-server are now supported by the chart.** #2410 added `repoServer.existingVolumes`, which lets you replace any of the default `emptyDir` volumes (`tmp`, `helmWorkingDir`, `plugins`, `varFiles`, `gpgKeyring`) with a PVC:

```yaml
repoServer:
  existingVolumes:
    tmp:
      persistentVolumeClaim:
        claimName: argocd-repo-server-tmp
    helmWorkingDir:
      persistentVolumeClaim:
        claimName: argocd-repo-server-workdir
```

You can create the PVCs via `extraObjects` in the same values file. This covers the single-replica and shared-RWX use cases that motivated this PR.

**What this PR adds beyond that is StatefulSet mode with `volumeClaimTemplates` (one PVC per replica).** That is an architectural decision this chart won't make ahead of upstream, since we mirror the upstream manifests. It's tracked in argoproj/argo-cd#7927, which is still open. If it lands there, we'll gladly take a fresh PR here to match.

Also worth noting: argoproj/argo-cd#22712 (in v2.14.12+) removed the project name from the repo cache key, so identical repos are no longer re-cloned per project. That significantly reduces the disk pressure that drove much of the demand in this thread.

Closing as superseded/stale (the branch has conflicts and hasn't been updated since 2022). Please 👍 the upstream issue if StatefulSet support matters to you.

PR #1691, opened 2022-12-09
feat(argocd-image-updater): prevent generation of label value of invalid length

steenzout:
Note on DCO:

Checklist:


charts/argocd-image-updater/templates/_helpers.tpl
```diff
@@ -38,7 +38,7 @@ Common labels
 helm.sh/chart: {{ include "argocd-image-updater.chart" . }}
 {{ include "argocd-image-updater.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | trunc 63 | quote }}
```
steenzout, line 41: for some reason this chart doesn't allow `image.tag` overrides. if it did, I have to admit, this would be more useful.

example where this was useful: #1693

if you don't want this feel free to close the PR.

charts/argocd-image-updater/templates/_helpers.tpl
```diff
@@ -38,7 +38,7 @@ Common labels
 helm.sh/chart: {{ include "argocd-image-updater.chart" . }}
 {{ include "argocd-image-updater.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
```
steenzout, line 40: does it make sense to only set this label when `.Chart.AppVersion` is set?

if I override the `image.tag` instead? it shouldn't be set? or should it be set to the `image.tag` value (without SHA digest)?

steenzout: simpler question: `argo-workflows` and `argo-events` actually always set this label. shouldn't this chart do the same?

jmeridth: Same here. Consistency is good. I say add it. @mkilchhofer do you agree?

mbevc1: :+1:

mkilchhofer: 👍

charts/argocd-image-updater/Chart.yaml
```diff
@@ -15,4 +15,4 @@ maintainers:
     url: https://argoproj.github.io/
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Add support for additional initContainers and additional volume/volumeMounts"
+    - "[Changed]: avoid app.kubernetes.io/version kubernetes label from exceeding maximum length (63)
```
mkilchhofer, line 18:
```suggestion
    - "[Changed]: avoid app.kubernetes.io/version kubernetes label from exceeding maximum length (63)"
```

jmeridth (approved):

Issue #1724, opened 2022-12-22
Getting err: unsupported version "0.22.1" when deploying an `EventBus`


rakhbari:
### Describe the bug

Right after deploying the `argo-events` chart, when you attempt to deploy an `EventBus`, no matter the name or namespace, you'll get the following error in the argo-events controller logs:
```
{"level":"error","ts":1665255012.769629,"logger":"argo-events.eventbus-controller","caller":"eventbus/controller.go:56","msg":"reconcile error","namespace":"default","eventbus":"default","error":"failed to get nats streaming version, err: unsupported version \"0.22.1\", supported versions: \"latest\""
```
I'm not the first one to run into this issue as it's been reported a few times in `argo-events` channel on CNCF Slack. This is just one of the threads:
https://cloud-native.slack.com/archives/C01TNKD6KL6/p1665255315412619

### Related helm chart

argo-events

### Helm chart version

2.0.8

### To Reproduce

1. Deploy the `argo-events` chart as per [the instructions](https://argoproj.github.io/argo-events/installation/#using-helm-chart)
2. Install the `default` native `EventBus` [in argo-events examples](https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/eventbus/native.yaml)
3. Tail the `argo-events-controller-manager` pod logs and you'll see the above error.

### Expected behavior

Installation of the `default` native `EventBus` [in argo-events examples](https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/eventbus/native.yaml) works fine with no errors in the `argo-events-controller-manager` pod logs.

### Additional context

This has also been reported as [an issue in the main `argo-events` repo](https://github.com/argoproj/argo-events/issues/2261), but the issue has been closed citing the fact that it's the Helm chart that's missing a crucial configuration.

The problem is that the chart's default `values.yaml` is missing the actual NATS version used in the [`argo-events` controller config](https://github.com/argoproj/argo-events/blob/master/manifests/base/controller-manager/controller-config.yaml#L10-L12).

The fix is quite simple: Add the expected NATS version to the `configs.nats.versions` in the chart's `values.yaml`.

pdrastil:
Resolved via https://github.com/argoproj/argo-helm/pull/1735

PR #1465, opened 2022-09-20
chore(argo-events) upgrade argo-events to 1.7.2

jmeridth:
[Release notes](https://github.com/argoproj/argo-events/releases/tag/v1.7.2)

[This is the only change](https://github.com/argoproj/argo-events/pull/2093) that I thought might need to be added to values.yaml but didn't see where it'd map in our K8S templates.

Signed-off-by: jmeridth <jmeridth@gmail.com>

Note on DCO:

If the DCO action in the integration test fails, one or more of your commits are not signed off. Please click on the *Details* link next to the DCO action for instructions on how to resolve this.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [ ] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Changes are automatically published when merged to `main`. They are not published on branches.

PR #1725, opened 2022-12-22
fix(argo-events): Added missing NATS version to `values.yaml`

rakhbari:
Fixes issue #1724 

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.5.5
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.16.9
+version: 5.16.10
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -18,9 +18,9 @@ maintainers:
     url: https://argoproj.github.io/
 dependencies:
   - name: redis-ha
-    version: 4.22.3
+    version: 4.22.4
     repository: https://dandydeveloper.github.io/charts/
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Support relabelings and metricRelabelings to Notification's ServiceMonitor"
+    - "[Changed]: Update redis-ha to v4.22.4"
--- a/charts/argo-cd/Chart.lock
+++ b/charts/argo-cd/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: redis-ha
   repository: https://dandydeveloper.github.io/charts/
-  version: 4.22.3
-digest: sha256:ef6269e4e073dad10c230ccfb069fc013608111c895c5e7568450bb3967cf195
-generated: "2022-11-03T12:04:33.673857+09:00"
+  version: 4.22.4
+digest: sha256:5df60910862b364ebfb82cba2b2f0951c39ad36446647fb3f501bdeadc92fbd7
+generated: "2022-12-26T22:58:11.561184+09:00"


rakhbari:
I'm a little baffled as to why the helm-docs check is failing. I looked at the `chart-test` output and the `helm-docs.sh` script is exiting with a non-zero exit code. But when I run this script locally, it runs fine (exit code 0). So I'll need one of you helm-docs gurus to let me know what (if anything) I might have missed, perhaps in `README.md.gotmpl`? Or perhaps in `values.yaml` add a `@default` comment for `configs.nats.versions`?

egze:
@rakhbari I checkout your branch and ran `scripts/helm-docs.sh`, was fine.

To make the linter pass, you can add the following to `charts/argo-events/README.md`

```
| configs.nats.versions | list | `[{"metricsExporterImage":"natsio/prometheus-nats-exporter:latest","natsStreamingImage":"nats-streaming:latest","version":"latest"},{"metricsExporterImage":"natsio/prometheus-nats-exporter:0.8.0","natsStreamingImage":"nats-streaming:0.22.1","version":"0.22.1"}]` | Supported versions of NATS event bus |
```

Replace whatever is the line with `| configs.nats.versions | ...` with the above.

charts/argo-events/Chart.yaml
```diff
@@ -16,3 +16,4 @@ maintainers:
 annotations:
   artifacthub.io/changes: |
     - "[Added]: Ability to specify port for webhook deployment"
+    - "[Fixed]: Missing NATS version in values.yaml"
```
egze, line 19: Looking at other pull requests and [docs here](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog), it seems you need to always provide just the latest change.

```suggestion
    - "[Fixed]: Missing NATS version in values.yaml"
```

egze:
Only thing missing now is the [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md)

charts/argo-events/Chart.yaml
rakhbari: @egze Thanks for the fix to the linter in `argo-events/README.md`.
About the suggestion you made about `argo-events/Chart.yaml`, I accepted your suggestion and used `Sign off and commit` right here in the PR UI, but now the DCO check fails. I thought the `Sign off and commit` would satisfy DCO requirements, but apparently it doesn't. So what do I need to do now? Re-commit locally with `git -s -m "..."` and re-push?

egze: Yes, you can try it.

pdrastil:
Hi @rakhbari, thank you for contribution. The version is intentionally omitted in defaults as the value is used by EventBus resources. If someone bumps this (and it will happen as it happened in the past) expect you will get broken during the upgrade when someone overwrites this. You can see discussion about this in PR https://github.com/argoproj/argo-helm/pull/1465. It's expected that user will provide and pin his versions to avoid this.

rakhbari:
@pdrastil Thanks for the additional context into this issue. It's understandable that you're worried about a broken `EventBus` deployment when the `argoproj/argo-events` dev team bumps their default version. But the point is that this Helm chart, left as is, already results in a broken `EventBus` deployment because the actual version used by the `argo-events` team is not in the list of supported versions. Wouldn't it be better to have a *working* Helm chart for now, and if the `argo-events` dev team bumps their default version, well, the Helm chart simply adds that new default version to its list of supported versions at that point.

The RCA time it took to chase down this issue was a significant sink of my time and I was just trying to avoid anyone else running into the same thing. As I mentioned in the referenced issue #1724, several people have ran into this same issue and even more time & productivity lost, which would be avoided if this default version would simply be added to the Helm chart values.yaml.

But alas, I'll yield to the consensus and decision of the community. Let me know and I can just close this PR unresolved.

pdrastil:
@rakhbari After some thought I'm ok to merge this as it might help other people. Even then I would recommend users to pin their versions.

charts/argo-cd/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: v2.5.5
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.16.7
+version: 5.16.9
```
pdrastil, line 6: @rakhbari Can you please rebase this? Argo CD changes should not be part of this PR.

rakhbari: @pdrastil: The changes to `argo-cd` and `argo-workflows` charts came in when I was trying to resolved the DCO `sign-off` issue, which instructed me to do `git rebase HEAD~8 --signoff`. But now those commits have been pushed to my fork and therefore part of this PR. I'm pretty sure a `rebase` isn't going to solve it. I think I'm going to have to do `git checkout upstream/master --` on my fork for both `argo-cd/` and `argo-workflows/` folders, then re-commit to remove them from this PR. Do you have a better set of steps to remove those commits from this PR?

egze: Sorry to jump in like that :) I think this is a good plan.

P.S. I'm also waiting for the PR 🚀

pdrastil: What I do in such case is to sync my main with upstream repo and then rebase the PR branch on top.

PR #1751, opened 2023-01-06
fix(argo-cd): Fully disable Dex when setting `dex.enabled: false`

cten:
Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.5.5
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.16.14
+version: 5.16.15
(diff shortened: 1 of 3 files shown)


pdrastil:
These and all other configuration values are very clearly marked as `optional: true` and thus not mandatory. If the value is not provided (which isn't if dex .enabled is false) it's not mounted / used by running a container. IMHO this PR just brings unnecessary conditional behaviour that do not have any effect and it will be harder to do syncs with upstream that adds new values. Thoughts @mkilchhofer @jmeridth

cten:
I understand for the ENVs, I can remove that logic, but for the volumeMounts there is no `optional` option I believe.

cten:
Sorry, I didn't know about the optional setting on volumes/volumeMounts. I will delete this PR and associated Issue.
Thanks

pdrastil:
Hi, no problem. It's relatively new and was introduced with upstream in 2.5.x release for K8s 1.22. This feature simplifies templating on multiple places as it's no longer needed to use if / else for older versions of K8s.

PR #1756, opened 2023-01-10
feat(argo-workflows): Only enable PVC RBAC when persistence is enabled

OneCricketeer:
Seems like PVC would only be needed for the database, which is under the same condition.

Signed-off-by: Jordan Moore <1930631+OneCricketeer@users.noreply.github.com>

Note on DCO:

Checklist:


charts/argo-workflows/templates/controller/workflow-controller-cluster-roles.yaml
```diff
@@ -31,6 +31,7 @@ rules:
   - get
   - watch
   - list
+{{- if .Values.controller.persistence }}
 - apiGroups:
```
OneCricketeer, line 35: Or this section could be moved down into the conditional? 😅

jmeridth:
@OneCricketeer my hesitation to move the PVC related permission fro the workflow controller under the persistence boolean is that upstream argo-workflows minimal example has them [here](https://github.com/argoproj/argo-workflows/blob/8562f95c34412a607000cb707ca7c9e61d4f9a4f/manifests/quick-start-minimal.yaml#L953) but also has quickstart examples of [mysql](https://github.com/argoproj/argo-workflows/blob/master/manifests/quick-start-mysql.yaml) and [postgresql](https://github.com/argoproj/argo-workflows/blob/master/manifests/quick-start-postgres.yaml) specifically.  I'm not 100% sure those PVCs and perms are strictly related to persistence and may still need to be exposed.  May want to ask that specific question in #argo-workflows CNCF slack.  Or maybe someone else here knows the answer.

You need an version increment and explanation of change added to the argo-workflows Chart.yaml for this PR also please (first checkbox in the description above)

OneCricketeer:
> version increment

About that... Seems like this would be a backwards-incompatible change for anyone expecting to use PVC out of the box? So, minor bump instead of patch?

jmeridth:
@OneCricketeer I don't think putting the PVC perms behind the persistence flag is a good idea.  The `.Values.controller.persistence` flag is if people want to use the database to offload large workflows that might exceed the 1MB limit of EtcD.  Explained [here](https://argoproj.github.io/argo-workflows/offloading-large-workflows/).  People may still want to use PVCs in their workflows like [here](https://argoproj.github.io/argo-workflows/walk-through/volumes/).

Am I missing something that you might be seeing that I'm not?

@yu-croco wdyt?

OneCricketeer:
Ah. From docs 

> Volumes are a very useful way to move large amounts of data from one step in a workflow to another. 

Makes sense.Will close. 

Though, this does require the service account to be created, as it appears as tripped people up in the comments

PR #1754, opened 2023-01-09
feat(argo-workflows): Add helper function to determine image value, minus tag

jmeridth:
We have a situation where we want to exclude the registry so that images can be pulled from in-network mirrors. Currently when you exclude the registry, you get stuck with a prefixing forward slash.

Had to keep the tags in the manifest files because helm template requires a colon in the image value.  Otherwise I'd move it into the helper function.

image entries changed: server, controller, executor

<details>
  <summary>Example when registry is default `quay.io` like current:</summary>

```
✗ helm template --release-name argo-workflows -f values.yaml --namespace argo . --debug --include-crds | grep "image" -C2                                                14:12:08
install.go:192: [debug] Original chart version: ""
install.go:209: [debug] CHART PATH: /Users/jasonmeridth/code/argo-helm/charts/argo-workflows

feat(argo-workflows): Add helper function to determine image value, minus tag
      containers:
        - name: controller
          image: "quay.io/argoproj/workflow-controller:v3.4.4"
          imagePullPolicy: Always
          command: [ "workflow-controller" ]
          args:
          - "--configmap"
          - "argo-workflows-workflow-controller-configmap"
          - "--executor-image"
          - "quay.io/argoproj/argoexec:v3.4.4"
          - "--loglevel"
--
      containers:
        - name: argo-server
          image: "quay.io/argoproj/argocli:v3.4.4"
          imagePullPolicy: Always
          securityContext:
            allowPrivilegeEscalation: false
```

</details>

<details>
  <summary>Example when registry is empty:</summary>

```
✗ helm template --release-name argo-workflows -f values.yaml --namespace argo . --debug --include-crds | grep "image" -C2                                                14:12:08
install.go:192: [debug] Original chart version: ""
install.go:209: [debug] CHART PATH: /Users/jasonmeridth/code/argo-helm/charts/argo-workflows

feat(argo-workflows): Add helper function to determine image value, minus tag
      containers:
        - name: controller
          image: "argoproj/workflow-controller:v3.4.4"
          imagePullPolicy: Always
          command: [ "workflow-controller" ]
          args:
          - "--configmap"
          - "argo-workflows-workflow-controller-configmap"
          - "--executor-image"
          - "argoproj/argoexec:v3.4.4"
          - "--loglevel"
--
      containers:
        - name: argo-server
          image: "argoproj/argocli:v3.4.4"
          imagePullPolicy: Always
          securityContext:
            allowPrivilegeEscalation: false
```

</details>

Note on DCO:

[...]

PR #1760, opened 2023-01-12
feat(argo-workflows): simplify image values so registry is included

jmeridth:
In #1754 I added helper function to fix issue if image registry is left empty in values.yaml so a prefixing forward slash is left.

I noticed the argo-cd chart values.yaml doesn't have a separate key for image registry.  This simplifies the including or excluding an image registry.

- [x] remove helper function from previous PR
- [x] update README
  - [x] remove [controller|executor|server].image.registry
  - [x] update [controller|executor|server].image.repository to include `quay.io` registry prefix by default
  - [x] remove mention of registry key in update from old argo chart at bottom
- [x] update image.repository to include default quay.io registry value
- [x] simplify image value usage

Signed-off-by: jmeridth <jmeridth@gmail.com>

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.4.4
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.22.8
+version: 0.23.0
 icon: https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 sources:
@@ -13,4 +13,4 @@ maintainers:
     url: https://argoproj.github.io/
 annotations:
   artifacthub.io/changes: |
-    - "[Added]: Helm helper function to allow image registry to be absent"
+    - "[Removed]: Removed image registry key from values.yaml for controller, executor and server"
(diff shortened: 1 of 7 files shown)


mkilchhofer:
Imho i am a bug fan of having a separate key for the registry as here in switzerland every large organization mirrors the images via Nexus or Artifactory. In my opinion we should sync that feature over to argo-cd

jmeridth:
@mkilchhofer we do the same with artifactory mirrors and don't use an explicit registry.  For your case do you use a local/internal registry FQDN? By removing it from the image url you get the exact same effect, right?  Having to have a helper function to do this seems overkill IMO.

mkilchhofer:
Hmm I reviewed the initial refactoring of the `argo` chart where the contributor @chgl refactored the chart: https://github.com/argoproj/argo-helm/pull/668/files?diff=unified&w=0#diff-e3f58f596f6078d4e3e486b7e8d28f929a29ebfb428f12fca00cb34241f4598bR33

I am a fan since the most charts (ie. bitnami charts) supports this also.

jmeridth:
@mkilchhofer please double check me, but it seems by [here](https://github.com/bitnami/charts/blob/3a09423ef04992f8f305a7c91b2e3ca8142534f3/bitnami/common/templates/_images.tpl#L20) this helper function in bitmani suffers the same fate as our helm chart, in that, if you do not provide a registry value it ends up looking like this `/blah/blah` wish a prefixing forward slash.  The reason I am a big fan of the single image string is that you can add/remove the registry as needed.  I get this removes the ability to do a global registry.  I'm torn tbh.  @yu-croco @pdrastil y'all care to chime in? 

I may also be interpreting that helper function incorrectly also.

mkilchhofer:
Yup, bitnami charts can also not "unset" the registry as it would lead to `/path/toimage:1.2.3`.

What is the use case to empty the registry? Preconfigured nodes?

jmeridth:
Mirror nodes in the cluster. I'm good closing this PR. Our helper function that got merged gives us what we need. 

The inconsistency between the charts bugs me. But we have a bit of that currently. 😄

PR #1774, opened 2023-01-18
chore(argo-rollouts): Upgrade argo-rollouts to 1.4.0

jmeridth:
Fixes #1763

Signed-off-by: jmeridth <jmeridth@gmail.com>

Note on DCO:

If the DCO action in the integration test fails, one or more of your commits are not signed off. Please click on the *Details* link next to the DCO action for instructions on how to resolve this.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Changes are automatically published when merged to `main`. They are not published on branches.

PR #1677, opened 2022-12-01
fix(argo-rollouts): Add traefikservices resource in role

JuanRamino:
Signed-off-by: juan ramino <juan.ramino@gmail.com>

Note on DCO:

If the DCO action in the integration test fails, one or more of your commits are not signed off. Please click on the *Details* link next to the DCO action for instructions on how to resolve this.

Checklist:

* [ ] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [ ] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [ ] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [ ] Any new values are backwards compatible and/or have sensible default.
* [ ] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Changes are automatically published when merged to `main`. They are not published on branches.

PR #1777, opened 2023-01-19
feat(argo-rollouts): Added flags to toggle provider-specific RBAC

illrill:
Added flags to be able to toggle provider-specific RBAC on or off.

Signed-off-by: Richard Johansson <richard.jimmy.johansson@gmail.com>

Note on DCO:

Checklist:


illrill:
Added a motivation for the added change to the PR description.

jmeridth:
I'm adding a bunch of the missing RBAC rules in #1774.

Nice work. With the provider RBAC rules offered upstream with no toggling, are we worried about leaving them in?  Does lean towards principle of least privilege but you're defaulting them to true anyways.  What do you think?

illrill:
> With the provider RBAC rules offered upstream with no toggling, are we worried about leaving them in?

The problem is that in order for a user to be allowed to create a Role with certain permissions, that **user must himself possess those very same permissions**. If he doesn't, Kubernetes won't allow him to create the Role. I assume the upstream manifests don't suffer that much from this problem since you can use Kustomize or similar to just cut out the stuff you don't want from the manifest.

As an example, in our clusters we don't run Istio or SMI (nor any of the other "providers"), so naturally we haven't granted our users any permissions over that type of resources. When one of our users tried to install this chart in his namespace (with `clusterInstall: false` so the Role comes with it) he was denied to do so, because his **user** didn't possess any of the permissions that the Role defines for Istio and SMI resources. So what we had to do as a workaround was to grant our **users** this:

```
- apiGroups:
  - networking.istio.io
  resources:
  - virtualservices
  verbs:
  - watch
  - get
  - update
  - list
- apiGroups:
  - split.smi-spec.io
  resources:
  - trafficsplits
  verbs:
  - create
  - watch
  - get
  - update
  - patch
```

> but you're defaulting them to true anyways

I figured it made sense to toggle them `true` by default to avoid a breaking/major change.

> What do you think?

Way ahead of me there with #1774, didn't see that one. With #1774 however, we would be further expanding upon this problem, as I would now need to grant my **users** even more permissions for Traefik/Apisix/AppMesh related resources - resources that don't even exist in my cluster and thus I would rather not grant "ghost" permissions to just to enable my users to install Argo Rollouts 😃 

What do you think?

illrill:
If we decide to proceed with the toggles, how should we go about it? Close this PR and move the toggles into yours? We also have #1677 that circles around the same area.

jmeridth:
@illrill my PR has merged.  please rebase and we can get yours in to toggle the provider parts.  Thank you for this work.

jmeridth (approved):
LGTM 👍

illrill:
@jmeridth Should be all done now 👍

charts/argo-rollouts/templates/controller/clusterrole.yaml
```diff
@@ -229,6 +242,9 @@ rules:
   - watch
   - get
   - update
+{{- end }}
+{{- if .Values.providerRBAC.providers.apisix }}
+# Traefik access needed when using the Apisix provider
```
jmeridth, line 247: Should this say `Traefik` here?

illrill: Thanks, corrected now.

PR #1841, opened 2023-02-14
fix(argo-cd): Add quotes on Ingress Host to allow wildcards

nicogl:
This change fixes a bug that prevents wildcards from being used in Ingress' `host`.
Currently there are no quotes surrounding the host, therefore it cannot begin with a wildcard, which should be possible as per [Ingress specification](https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards).
Note that without this quotes, hostnames starting with wildcard make the Helm Chart installation fail as follows:

```
Error: YAML parse error on argo-cd/templates/argocd-server/ingress.yaml: error converting YAML to JSON: yaml: line 26: did not find expected alphabetic or numeric character
helm.go:84: [debug] error converting YAML to JSON: yaml: line 26: did not find expected alphabetic or numeric character
YAML parse error on argo-cd/templates/argocd-server/ingress.yaml
```

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.6.1
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.20.4
+version: 5.20.5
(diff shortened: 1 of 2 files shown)


jmeridth:
Need to update Chart.yaml

mbevc1 (approved):

yu-croco (approved):
LGTM!

Issue #1761, opened 2023-01-13
[argo-cd] Add ability to specify backend service in ingress for argocd-server


jdomag:
### Is your feature request related to a problem?

Template for argo-cd/argocd-server/ingress.yaml  is build in a way that:
1) Each path specified in `server.ingress.path` has the same type of pathType. I would like to specify pathType for each path on the list
2) It's not possible to specify backend service name and port for default path. My use case:
We use AWS ALB as the entry point of our k8s cluster. We use istio ingress gateway in k8s to receive all the ingress traffic and distribute it across the different services. This setup requires to create an ingress object with a given name e.g. `argocd.mydomain.com` and point it to the istio ingress gateway service. Current helm chart doesn't allow to overwrite the backend service name or port - it always uses argocd-server service.  

### Related helm chart

argo-cd

### Describe the solution you'd like

1) pathType value is under array of server.ingress.paths instead of server.ingress 
2) Add following fields to values.yaml:
`server.ingress.paths[].backend.service.name:`  - if not specified argocd-server service is used 
`server.ingress.paths[].backend.service.port:` - Can be string or int. If not specified use argocd-server service as default

jdomag:
If this request sounds reasonable I can provide PR.

ankitcharolia:
@jdomag  : we are using argocd with istio ingress gateway. We are using ingress gateway for receiving all the ingress traffic (we do't use Ingress Nginx controller). It would be awesome if official helm chart supports creating Ingress gateway and virtualService with some values.yaml parameter.

pdrastil:
@ankitcharolia For this please use `extraObjects`, otherwise this will lead to managing custom resources for Istio, Linkerd, Traefik, Gloo, Kong, etc. to provide additional ingress controller integration.

jdomag:
@ankitcharolia I agree with @pdrastil, for VirtualService objects, `extraObjects: []` should be used. However for ingress, it would be nice to reuse ingress.yaml, however it's not possible with the current structure hence I proposed enhancement. Moving `prefixType` into the `paths: []`, sounds like a breaking change though as people would need to adjust the values.yaml they currently use.

pdrastil:
[...]

PR #1853, opened 2023-02-21
feat(argo-cd): added ability to install server ingress in different n…

jdomag:
This PR adds the ability to control a namespace in which ingress of argo-cd server is installed. This helps to use `.server.ingress.extraPaths` to reference the backend service that resides in the different namespace than argo-cd. It's a common use case for Istio Ingress Gateway combined with AWS ALB.

This PR solves: 
https://github.com/argoproj/argo-helm/issues/1761

---
Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.6.2
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.22.1
+version: 5.23.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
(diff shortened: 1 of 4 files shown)


pdrastil:
Repeating answer here as well:

@jdomag For Istio I would stay away from native Ingress as the design of this service mesh forces you to rethink your cluster networking completely and use their CRDs for everything (rather provide VirtualService via extraObjects). You can always target Istio gateway in different namespace via:

```yaml
kind: VirtualService
gateways
- <namespace>/<gateway-name>
```

Also few caveats with Istio - sidecar injection is not exactly great fit for Argo CD as it breaks mutual TLS between components. Ingress HTTP -> HTTPS redirection will not work on native Ingress if you plan to run with `server.insecure` flag and terminate TLS on ingress controller.

pdrastil:
Btw if you deploy the Ingress into different namespace then the selector for service wouldn't work because Service is still in original namespace.

mkilchhofer:
I'll close this, as Kubernetes itself does not support having Ingress in one namespace and Service in another one. Feel free to continue discussion in #1761

PR #1861, opened 2023-02-24
feat(argo-cd): Adding "appProtocol: HTTPS" to HTTPS service port of argocd-server.

hau21um:
We use Gateway API in GKE, where you can not specify [appProtocol](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) using [annotaion](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#def_inf_hc) as with ingress. By default load balancer (LB) health checks will be HTTP (even if the port is 443) and only way to force them to be HTTPS are by using this appProtocol field. And we do not want any unencrypted traffic between LB and our pods.
Also this change is only about setting it on service port 443 called https, so not touching service port 80 called http.

---
Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.4.6
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.23.0
+version: 0.23.1
 icon: https://raw.githubusercontent.com/argoproj/argo-workflows/master/docs/assets/argo.png
 home: https://github.com/argoproj/argo-helm
 sources:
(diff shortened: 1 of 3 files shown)


mkilchhofer (changes requested):
Hi @hau21um  thanks for your contribution.
Can you add a description to the PR why we need this?

Upstream manifests don't define this and also some users run the server without TLS ans terminate the TLS session on the ingress, therefore it cannot be hardcoded without a conditional.

hau21um:
Reasoning behind this. We use Gateway API in GKE, where you can not specify [appProtocol](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) using [annotaion](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#def_inf_hc) as with ingress. By default load balancer (LB) health checks will be HTTP  (even the port is 443) and only way to force them to be HTTPs are by using this appProtocol field. And we do not want any unencrypted traffic between LB and our pods.
Also this change is only about setting it on service port 443 called https, so not touching service port 80 called http.

mkilchhofer (changes requested):

hau21um:
Sorry, seems I've closed this one by mistake and created new one as  #1955 with suggested conditional templating. Thank you all!.

PR #1864, opened 2023-02-26
fix(argo-cd): Clarify syntax in values.yaml

TheRealNoob:
the previous format was incorrectly indented and thus misleading.  Hopefully this aligns it with the other examples I see scattered around the file.

Also I should note that I think having a selector on this rule is confusing.  Rules can't have selectors.

Note on DCO:

Checklist:


TheRealNoob:
I can bump the chart version if needed, however no functional changes were made so I lean more towards not bumping it.

mbevc1:
Following contribution guidelines please bump the chart version and update the changelog as well.

TheRealNoob:
> Following contribution guidelines please bump the chart version and update the changelog as well.

Done.

jmeridth:
Need to update the artifact annotations at bottom Chart.yaml also please.

TheRealNoob:
> Need to update the artifact annotations at bottom Chart.yaml also please.

Done.

charts/argo-cd/README.md
```diff
@@ -105,6 +105,10 @@ For full list of changes please check ArtifactHub [changelog].
 
 Highlighted versions provide information about additional steps that should be performed by user when upgrading to newer version.
 
+### 5.23.3
```
jmeridth, line 108: This is not needed.

jmeridth (changes requested):

charts/argo-cd/README.md
TheRealNoob: That's what I thought as well when I was making the change but @mbevc1 said to update the changelog.  Did he mean something else?

TheRealNoob: Removed.

charts/argo-cd/values.yaml
```diff
@@ -714,7 +714,15 @@ controller:
     rules:
       # -- Deploy a PrometheusRule for the application controller
       enabled: false
-      # -- PrometheusRule.Spec for the application controller
+      # -- PrometheusRule namespace
+      namespace: "" # "monitoring"
+    
+      selector: {}
+      # prometheus: kube-prometheus
```
jmeridth, line 721: Does this need to be indented one more to match the others in the file?

jmeridth:
> Also I should note that I think having a selector on this rule is confusing. Rules can't have selectors.

I believe you are correct.  ServiceMonitor's have selectors, like ours [here](https://github.com/argoproj/argo-helm/blob/3a6617ce8f76284776486007cf3279501068cc83/charts/argo-cd/templates/argocd-application-controller/servicemonitor.yaml#L11-L13).  The selector was added to the PrometheusRule with #178 and especially [this commit](https://github.com/argoproj/argo-helm/pull/178/commits/539efe3a12304b87f28a745ff937a65c9e37a238).  I think it can be removed from the PromethesuRule if you'd like to remove it from the template [here](https://github.com/argoproj/argo-helm/blob/3a6617ce8f76284776486007cf3279501068cc83/charts/argo-cd/templates/argocd-application-controller/prometheusrule.yaml#L11-L13) and from your values.yaml change.  Wdyt?  This might require a deprecation notice in README.md. /cc @mkilchhofer wdyt?

charts/argo-cd/values.yaml
TheRealNoob: most of the helm charts that I've looked at (public and private) treat the `#` comment symbol as invisible.  Expecting the consumer to replace the symbol with a space.  Doing it the other way around can lead to confusion in some scenarios.  An example:
```yaml
rootDict:
  key1: value1
# key2: value2
```

With this in mind I see that a lot of the helm chart does not treat the comment as invisible, however there are some exceptions.  [Comment followed by two spaces](https://github.com/argoproj/argo-helm/blob/3a6617ce8f76284776486007cf3279501068cc83/charts/argo-cd/values.yaml#L85) and [Invisible comment](https://github.com/argoproj/argo-helm/blob/3a6617ce8f76284776486007cf3279501068cc83/charts/argo-cd/values.yaml#L91).  This is by no means comprehensive I'm just cherry picking.  The way that I'm used to reading values files tells me that [this ](https://github.com/argoproj/argo-helm/blob/3a6617ce8f76284776486007cf3279501068cc83/charts/argo-cd/values.yaml#L702)is meant to be indented by 4 spaces, however I know that's not the case.  You tell me which way you'd like it formatted and I'll do that.

TheRealNoob:
I agree `.Values.controller.metrics.rules.selector` should be removed, however I don't think this is the MR to do it in.  I think that change should probably bump the chart version from `5.23.2` to `5.24.0` just because it's making a minor breaking change.  I apologize if you already know the rest of this comment, I just thought I'd explain my thought process now while I have the time to do it.

Helm versioning follows the SemVer scheme which defines the naming convention as `MAJOR.MINOR.PATCH` which says you increment the:
* MAJOR version when you make incompatible API changes
* MINOR version when you add functionality in a backwards compatible manner
* PATCH version when you make backwards compatible bug fixes
[source](https://semver.org/)

Removing the selector is a breaking change, however I really don't feel it justifies a major number.  So I look at the above more as guidelines than rules :).  I think this current MR that I'm doing here definitely falls under the `patch` category.

mkilchhofer:
When I first looked at this, when I started to contribute and help maintain here, I was also confused.

The idea behind this "selector" is for the prometheus-operator running on the cluster. You can configure on your `kind: Prometheus` what ServiceMonitors/PrometheusRules/etc should be picked up. Eg if you have multiple Prometheus instances running like one instance for the core cluster and others for Applications or Tenants (when users do soft tenancy on kubernetes).

Eg. when you use the kube-prometheus-stack chart, the selector by default is `release: <ReleaseName`. Some users (the one sonly using one Prom instance per cluster) disable this feature on Prometheus operator completely by setting `serviceMonitorSelectorNilUsesHelmValues: false`

Does that help?

charts/argo-cd/values.yaml
TheRealNoob: let me know which way you'd like it formatted

jmeridth: Whether it's right or wrong, my vote is for consistency with the rest of this project's current choice. Wdyt?

TheRealNoob: alright indented to match the others.

TheRealNoob:
@jmeridth ready for review

jmeridth (approved):

mbevc1:
You need to update to docs as well and that will pass the tests.

TheRealNoob:
> You need to update to docs as well and that will pass the tests.

@mbevc1 Please point me to what specifically needs to be updated.  I added a node to the changelog as you instructed me before and @jmeridth told me to remove it.  I agree with him that that changelog (at present) only contains info about changes which could impact users of this helm chart.

jmeridth:
@TheRealNoob run `./scripts/helm-docs.sh` locally to have it update README.md for you. Should fix the issue.

mbevc1:
@TheRealNoob what @jmeridth said :smile:

mkilchhofer (approved):

jmeridth (approved):

Issue #1866, opened 2023-02-27
Web-Based Terminal Not Supported by Chart when using Namespaced Argo


ugoogalizer:
### Describe the bug

According to the documentation on the [Web Based Terminal]( https://argo-cd.readthedocs.io/en/stable/operator-manual/web_based_terminal/), the second step to enable to terminal is: 

> Patch the argocd-server Role (if using namespaced Argo) or ClusterRole (if using clustered Argo) to allow argocd-server to exec into pods
> ```
> - apiGroups:
>   - ""
>   resources:
>   - pods/exec
>   verbs:
>   - create
> ```

While this appears implemented and supported for this [Helm Chart's ClusterRole](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-server/clusterrole.yaml#L31), the equivalent has not been implemented for this [Helm Chart's Role](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-server/role.yaml).

I would like to enable this in my environment, however I believe this limitation is blocking me.  Attempts to patch the Role manually, simply get reverted immediately as my argo-cd is setup to deploy itself and it detects the deviation from the chart. 

Is this deliberate, or if not, can it be added please?

### Related helm chart

argo-cd

### Helm chart version

5.23.2

### To Reproduce

Install argo-cd with the argo-cd helm chart and have it manage itself.  Attempt to follow  [Web Based Terminal]( https://argo-cd.readthedocs.io/en/stable/operator-manual/web_based_terminal/) instructions, but be unable to complete step 2 as the helm chart does not support it.  Unable to patch because it doesn't align with the helm chart either.

### Expected behavior

Able to enable Web Based Terminal in argo-cd using an namespaced Argo and this helm chart.

### Screenshots

Without the fix, k8s denies users from being able to open a terminal from the argo-cd web-ui.
![image](https://user-images.githubusercontent.com/1974156/221491831-b0e2ebcf-7092-4ec0-8ded-2295632f76e9.png)

ugoogalizer:
I have attemped to create a pull request to fix this under #1867. 

This is my first pull request, so sorry if I've missed something!

PR #1867, opened 2023-02-27
fix(argo-cd): Added pod exec permission to argo-server Role when exec.enabled is True.

ugoogalizer:
As per issue #1866 , attempting to fix this by adding the relevant code to grant the pod/exec create permission to the argo-server Role when exec.enabled is True.

The argocd-server Role requires the pods/exec create permission in order to be able to start the web based terminal as per: https://argo-cd.readthedocs.io/en/stable/operator-manual/web_based_terminal/

This brings the Role in line with the ClusterRole change already made

Note on DCO:

Checklist:

- [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)

### DIFF
--- a/charts/argo-cd/templates/argocd-server/role.yaml
+++ b/charts/argo-cd/templates/argocd-server/role.yaml
@@ -42,3 +42,11 @@ rules:
   verbs:
   - create
   - list
+{{- if eq (toString (index (coalesce .Values.server.config .Values.configs.cm) "exec.enabled")) "true" }}
+- apiGroups:
+  - ""
+resources:
+  - pods/exec
+verbs:
+  - create
+{{- end }}
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -23,5 +23,6 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - kind: fixed
-      description: Fixed typos in values.yaml and README.md
+    - kind: added
+      description: Added pod exec permission to argo-server Role when exec.enabled is True.
+      
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.6.2
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.23.2
+version: 5.23.3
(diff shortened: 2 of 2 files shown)


charts/argo-cd/templates/argocd-server/role.yaml
```diff
@@ -42,3 +42,11 @@ rules:
   verbs:
   - create
   - list
+{{- if eq (toString (index (coalesce .Values.server.config .Values.configs.cm) "exec.enabled")) "true" }}
```
mkilchhofer, line 45: Do we need an AND condition here to the namespaced / server.clusterAdminAccess option @jmeridth @pdrastil?

mkilchhofer (commented):
Hi @ugoogalizer, Thanks for bringing that up and filing a PR. Almost perfect for the first hit 🙏 , but I wanted to ask other maintainers here if we need to check for the "namespaced mode"?

@pdrastil @jmeridth @mbevc1  WDYT?

charts/argo-cd/templates/argocd-server/role.yaml
pdrastil: @mkilchhofer I don't think we need any additional condition for this PR.

The question is if we actually need these checks for Role and ClusterRole as the Argo UI also requires proper permission in Argo RBAC or AppProject. Without these permissions terminal is not shown in the UI.

pdrastil (approved):

mbevc1:
LGTM if we can get testing passing and resolve conflicts.

mbevc1 (approved):

charts/argo-cd/templates/argocd-server/role.yaml
mkilchhofer: Agree. Let's merge it :)

mkilchhofer (approved):

ugoogalizer:
Awesome! Thanks all!

ugoogalizer:
@mkilchhofer  @mbevc1 @pdrastil 

I'm so very sorry, I'm still learning this and my pull request appears to have broken the chart as I've got the yaml spacing wrong somehow. Assuming the best way to fix is to submit another PR request? 

`- apiGroups:
  - ""
resources:
  - pods/exec
verbs:
  - create`

should have been: 

- apiGroups:
  - ""
  resources:
  - pods/exec
  verbs:
  - create

pdrastil:
@ugoogalizer Hi - don't worry this can happen and no one caught this as it's whitespace.

I've opened https://github.com/argoproj/argo-helm/pull/1871 with a fix.

PR #1874, opened 2023-03-01
fix(argocd-apps): Add revisionHistoryLimit option to apps

sergialonsaco:
Note on DCO:

Checklist:

### DIFF
--- a/charts/argocd-apps/ci/applications-values.yaml
+++ b/charts/argocd-apps/ci/applications-values.yaml
@@ -20,6 +20,7 @@ applications:
     automated:
       prune: false
       selfHeal: false
+  revisionHistoryLimit: 5
   ignoreDifferences:
   - group: apps
     kind: Deployment
--- a/charts/argocd-apps/ci/applications-values.yaml
+++ b/charts/argocd-apps/ci/applications-values.yaml
@@ -20,7 +20,7 @@ applications:
     automated:
       prune: false
       selfHeal: false
-  revisionHistoryLimit: 5
+  revisionHistoryLimit: null
   ignoreDifferences:
   - group: apps
     kind: Deployment
--- a/charts/argocd-apps/values.yaml
+++ b/charts/argocd-apps/values.yaml
@@ -30,7 +30,7 @@ applications: []
 #     automated:
 #       prune: false
 #       selfHeal: false
-#   revisionHistoryLimit: 5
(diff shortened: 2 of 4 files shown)


jmeridth (approved):

charts/argocd-apps/ci/applications-values.yaml
```diff
@@ -20,6 +20,7 @@ applications:
     automated:
       prune: false
       selfHeal: false
+  revisionHistoryLimit: 5
```
mkilchhofer, line 23: Maybe `null` or `""` would be a better default:

```
kubectl explain applications.spec
KIND:     Application
VERSION:  argoproj.io/v1alpha1

RESOURCE: spec <Object>

DESCRIPTION:
     ApplicationSpec represents desired application state. Contains link to
     repository with application definition and additional parameters link
     definition revision.

FIELDS:
   (..)

   revisionHistoryLimit <integer>
     RevisionHistoryLimit limits the number of items kept in the application's
     revision history, which is used for informational purposes as well as for
     rollbacks to previous versions. This should only be changed in exceptional
     circumstances. Setting to zero will store no history. This will reduce
     storage used. Increasing will increase the space used to store the history,
     so we do not recommend increasing it. Default is 10.

   (..)
```

> so **we do not recommend increasing it. Default is 10**.

sergialonsaco:
```suggestion
  revisionHistoryLimit: null
```
Is this a valid value for revisionHistoryLimit if \<integer\> is expected?

mkilchhofer: `null` into `with .revisionHistoryLimit` would not render that block

sergialonsaco: Great, thanks! Committing the change

charts/argocd-apps/values.yaml
```diff
@@ -30,6 +30,7 @@ applications: []
 #     automated:
 #       prune: false
 #       selfHeal: false
+#   revisionHistoryLimit: 5
```
sergialonsaco, line 33:
```suggestion
#   revisionHistoryLimit: null
```
Updating the values example to null

mkilchhofer (approved):
lgtm

jmeridth (approved):

Issue #1883, opened 2023-03-03
Update entrypoints after argo pr 12707


gczuczy:
### Is your feature request related to a problem?

The entrypoint change in https://github.com/argoproj/argo-cd/pull/12707 is another step for making a distroless image possible.
Currently the argocd helm chart does not make it possible to remove the entrypoint.sh from the manifests for using tini directly instead.

Once the PR is merged, `entrypoint.sh` can be removed from the manifests, and all services started directly with `args`. Please see the first comment in the PR for details and kubernetes docs on the topic.

### Related helm chart

argo-cd

### Describe the solution you'd like

Removing the dependency on any kind of shell, so there's one less blocker for a distroless image.

### Describe alternatives you've considered

I'm not sure there are a lot of alternatives, tini is pretty much accepted as the de-facto standard for container entrypoints.

yu-croco:
This issue seems resolved.

PR #1898, opened 2023-03-14
feat(argo-c): Entrypoint can be configured now

gczuczy:
This PR prepares the argo-cd chart for https://github.com/argoproj/argo-helm/issues/1883

Two new global values has been introduced:
`global.entrypoint.useImplicit`: defaults to false, if true, then containers are not explicitly declaring an entrypoint with `command`.
`global.entrypoint.entrypoint`: Currently defaults to `entrypoint.sh`, in case `useImplicit` is false, the entrypoint specified here will be used in the command.

Aftger the argo2.7 release, `useImplicit` can be set to true, and for safety reasons entrypoint should be set to the tini using an absolute path.

Having this little mechanic accessable beforehand provides no incompatible changes, and also allows the argocd helm chart users to run their deployments with directly tini or any other way (currently it's not configurable).

Note on DCO:

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.6.4
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.26.0
+version: 5.26.1
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -23,5 +23,5 @@ dependencies:
     condition: redis-ha.enabled
 annotations:
   artifacthub.io/changes: |
-    - kind: Added
-      description: ConfigMap for Config Management Plugins using sidecar approach
+    - kind: changed
+      description: entrypoint usage can be configured
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -394,6 +394,8 @@ NAME: my-release
 | global.affinity.nodeAffinity.type | string | `"hard"` | Default node affinity rules. Either: `none`, `soft` or `hard` |
 | global.affinity.podAntiAffinity | string | `"soft"` | Default pod anti-affinity rules. Either: `none`, `soft` or `hard` |
 | global.deploymentAnnotations | object | `{}` | Annotations for the all deployed Deployments |
+| global.entrypoint.entrypoint | string | `"entrypoint.sh"` | The entrypoint to use for the containers. |
+| global.entrypoint.useImplicit | bool | `false` | Implicitly use the docker image's entrypoint. This requires the image to have ENTRYPOINT set properly |
 | global.hostAliases | list | `[]` | Mapping between IP and hostnames that will be injected as entries in the pod's hosts files |
 | global.image.imagePullPolicy | string | `"IfNotPresent"` | If defined, a imagePullPolicy applied to all Argo CD deployments |
 | global.image.repository | string | `"quay.io/argoproj/argocd"` | If defined, a repository applied to all Argo CD deployments |
--- a/charts/argo-cd/templates/argocd-applicationset/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-applicationset/deployment.yaml
@@ -51,8 +51,11 @@ spec:
         - name: {{ .Values.applicationSet.name }}
           image: {{ default .Values.global.image.repository .Values.applicationSet.image.repository }}:{{ default (include "argo-cd.defaultTag" .) .Values.applicationSet.image.tag }}
           imagePullPolicy: {{ default .Values.global.image.imagePullPolicy .Values.applicationSet.image.imagePullPolicy }}
+          {{- if not .Values.global.entrypoint.useImplicit }}
           command:
-            - entrypoint.sh
+            - {{ .Values.global.entrypoint.entrypoint | quote }}
+          {{- end }}
+          args:
             - argocd-applicationset-controller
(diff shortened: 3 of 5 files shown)


yu-croco:
Thank you for your contribution! 

> Aftger the argo2.7 release, useImplicit can be set to true, and for safety reasons entrypoint should be set to the tini using an absolute path.
> Having this little mechanic accessable beforehand provides no incompatible changes, and also allows the argocd helm chart users to run their deployments with directly tini or any other way (currently it's not configurable).

IMO,
For safety, it's better to make it as draft until Argo CD v2.7 is release or write down on values.yaml that users can configure these attributes after Argo CD v2.7 or later ...? 🤔

gczuczy:
> Thank you for your contribution!
> 
> > Aftger the argo2.7 release, useImplicit can be set to true, and for safety reasons entrypoint should be set to the tini using an absolute path.
> > Having this little mechanic accessable beforehand provides no incompatible changes, and also allows the argocd helm chart users to run their deployments with directly tini or any other way (currently it's not configurable).
> 
> IMO, For safety, it's better to make it as draft until Argo CD v2.7 is released or write down on values.yaml explicitly that users can configure these attributes after Argo CD v2.7 or later ...? 🤔

I wouldn't say so. For an example our internal images are already trying to the `ENTRYPOINT` defined in the container, and the blocker for this is the lack of support from the helm chart. Even if someone wants to override the entrypoint at this time, it's impossible without modifying chart, there's simply no support for it currently.

With this PR, this support for overriding the entrypoint is introduced, so people who want to do this (with using this chart) will be able to.

Regarding the argo-cd image's entrypoint chage, it only means that after the 2.7 release, the defaults in the values will have to be updated to reflect the changes.

With the current changes this PR is producing effectively the very same chart. It's the same `entrypoint.sh` at the Deployments' `spec.template.spec.containers.*.command`:

```
(...)
# Source: argo-cd/templates/argocd-applicationset/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: release-name-argocd-applicationset-controller
(...)
    spec:
      serviceAccountName: argocd-applicationset-controller
      containers:
        - name: applicationset-controller
          image: quay.io/argoproj/argocd:v2.6.4
          imagePullPolicy: IfNotPresent
          command:
            - "entrypoint.sh"
          args:
            - argocd-applicationset-controller
            - --metrics-addr=:8080
            - --probe-addr=:8081
            - --webhook-addr=:7000
```

And
```
# Source: argo-cd/templates/argocd-repo-server/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: release-name-argocd-repo-server
(...)
    spec:
      serviceAccountName: release-name-argocd-repo-server
      containers:
      - name: repo-server
        image: quay.io/argoproj/argocd:v2.6.4
        imagePullPolicy: IfNotPresent
        command:
          - "entrypoint.sh"
        args:
        - argocd-repo-server
        - --port=8081
        - --metrics-port=8084
```
Effectively, it's no changes with the defaults in `values.yaml`.

yu-croco:
@gczuczy 
Thank you for explaining.

> For an example our internal images are already trying to the ENTRYPOINT defined in the container, and the blocker for this is the lack of support from the helm chart. Even if someone wants to override the entrypoint at this time, it's impossible without modifying chart, there's simply no support for it currently.

Yes, I am aware that this PR does not change current configurations, as far as users don't override values.yaml.

> With this PR, this support for overriding the entrypoint is introduced, so people who want to do this (with using this chart) will be able to.
>
> Regarding the argo-cd image's entrypoint chage, it only means that after the 2.7 release, the defaults in the values will have to be updated to reflect the changes.

Once this PR is released, users can configure the values even though Argo CD is still v2.6.x.
So I just wonder if it would cause any confusion/wrong use case to users, though it would be a small possibility.


The changes themselves are LGTM, so let me ask other maintainers to review. 🙋

gczuczy:
> @gczuczy Thank you for explaining.
> 
> > For an example our internal images are already trying to the ENTRYPOINT defined in the container, and the blocker for this is the lack of support from the helm chart. Even if someone wants to override the entrypoint at this time, it's impossible without modifying chart, there's simply no support for it currently.
> 
> Yes, I am aware that this PR does not change current configurations, as far as users don't override values.yaml.
> 
> > With this PR, this support for overriding the entrypoint is introduced, so people who want to do this (with using this chart) will be able to.
> > Regarding the argo-cd image's entrypoint chage, it only means that after the 2.7 release, the defaults in the values will have to be updated to reflect the changes.
> 
> Once this PR is released, users can configure the values even though Argo CD is still v2.6.x. So I just wonder if it would cause any confusion/wrong use case to users, though it would be a small possibility.
> 
> The changes themselves are LGTM, so let me ask other maintainers to review. 🙋

Thanks. In the current state of ArgoCD I would say that this feature is most useful for users who are baking their own image.

There's one thing I'm contemplating about. Other components are using the service directly as an entrypoint, and yet I'm not sure about it, but my gut instinct is telling me that those might also benefit from being able to use an init (such as tini) entry point. Which is especially useful if there are any side processes inside the container, but otherwise still useful. Here's a writeup on it from tini's side:
https://github.com/krallin/tini/issues/8

I don't exactly know which other components might spawn processes, but if any of the others do, it would be beneficial to include those as well.

gczuczy:
@davidkarlsen , @mr-sour, @yann-soubeyrand, @mbevc1, @mkilchhofer, @yu-croco, @jmeridth, @pdrastil May I ask when will you have some time to check this, so we can merge it before argo 2.7?

mbevc1:
Nothing specifically comes to mind that would immediately break and should be okay.

yu-croco (approved):

mbevc1 (approved):

gczuczy:
@mbevc1, @yu-croco Thank you.

PR #1937, opened 2023-03-30
fix(argo-cd): add namespace field for namespace scoped resources

jakuboskera:
This PR adds namespace field for namespace scoped resources.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.6.7
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.27.4
+version: 5.27.5
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
--- a/charts/argo-cd/templates/argocd-notifications/servicemonitor.yaml
+++ b/charts/argo-cd/templates/argocd-notifications/servicemonitor.yaml
@@ -3,6 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ template "argo-cd.notifications.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
   {{- if .Values.notifications.metrics.serviceMonitor.namespace }}
   namespace: {{ .Values.notifications.metrics.serviceMonitor.namespace }}
   {{- end }}
--- a/charts/argo-cd/templates/argocd-application-controller/metrics.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/metrics.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "argo-cd.controller.fullname" . }}-metrics
+  namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.controller.name "name" "metrics") | nindent 4 }}
     {{- with .Values.controller.metrics.service.labels }}
--- a/charts/argo-cd/templates/argocd-application-controller/networkpolicy.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/networkpolicy.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
   name: {{ template "argo-cd.controller.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
 spec:
   ingress:
   - from:
--- a/charts/argo-cd/templates/argocd-application-controller/pdb.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/pdb.yaml
@@ -3,6 +3,7 @@ apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: {{ include "argo-cd.controller.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
     {{- with .Values.controller.pdb.labels }}
--- a/charts/argo-cd/templates/argocd-application-controller/role.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/role.yaml
@@ -2,6 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ template "argo-cd.controller.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
 rules:
--- a/charts/argo-cd/templates/argocd-application-controller/rolebinding.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/rolebinding.yaml
@@ -2,6 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: {{ template "argo-cd.controller.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
 roleRef:
--- a/charts/argo-cd/templates/argocd-application-controller/serviceaccount.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/serviceaccount.yaml
@@ -4,6 +4,7 @@ kind: ServiceAccount
 automountServiceAccountToken: {{ .Values.controller.serviceAccount.automountServiceAccountToken }}
 metadata:
   name: {{ template "argo-cd.controllerServiceAccountName" . }}
+  namespace: {{ .Release.Namespace | quote }}
 {{- if .Values.controller.serviceAccount.annotations }}
   annotations:
   {{- range $key, $value := .Values.controller.serviceAccount.annotations }}
--- a/charts/argo-cd/templates/argocd-application-controller/statefulset.yaml
+++ b/charts/argo-cd/templates/argocd-application-controller/statefulset.yaml
@@ -8,6 +8,7 @@ metadata:
     {{- end }}
   {{- end }}
   name: {{ template "argo-cd.controller.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
 spec:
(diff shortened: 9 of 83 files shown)


mkilchhofer:
I saw such PRs in the past, but I really don't understand, why this is needed. Helm automatically adds this upon install for all namespaced resources. As this is a core feature of helm itself, I prefer to not add this.

But maybe you can elaborate your use case?

jakuboskera:
> I saw such PRs in the past, but I really don't understand, why this is needed. Helm automatically adds this upon install for all namespaced resources. As this is a core feature of helm itself, I prefer to not add this.
> 
> But maybe you can elaborate your use case?

I am trying to use a Terraform for bootstrap the infrastructure. One of the component of bootstrap is to install argocd. As I want to do just initial installation of argocd via terraform and then manage argocd via argocd, I want to apply just plain text manifests to a cluster and not to make helm release (because later then it will be managed via argocd and argocd doesn't make helm releases rather it makes `helm template` and then `kubectl apply -f ...`. One of the prereqs of applying plain text manifests using [Kubernetes Terraform provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest) is that manifests have specified namespace field, otherwise apply fails.

See error from Terraform apply

```bash
╷
│ Error: Namespace required
│ 
│   with kubernetes_manifest.argocd["charts/argo-cd/templates/dex/deployment.yaml"],
│   on main.tf line 34, in resource "kubernetes_manifest" "argocd":
│   34: resource "kubernetes_manifest" "argocd" {
│ 
│ Resources of type 'apps/v1, Kind=Deployment' require a namespace
```

For me it is a best practise to specify the namespace field to every Kubernetes manifest even if it is Helm template. Look for example at [argo-cd Helm chart from bitnami](https://github.com/bitnami/charts/tree/main/bitnami/argo-cd), where is namespace field specified (also in other bitnami's charts).

Basically it doesn't cost anything but it can bring some benefits in other use cases (e.g. mentioned use case with Terraform).

mbevc1:
Could you use TF K8s provider to create the namespace first, before applying your manifest?

pdrastil:
This works fine:

```terraform
resource "kubernetes_namespace" "argocd" {
  metadata {
    name = "argocd"
  }
  lifecycle {
    ignore_changes = [
      metadata.0.annotations["cattle.io/status"],
      metadata.0.annotations["field.cattle.io/projectId"],
      metadata.0.annotations["lifecycle.cattle.io/create.namespace-auth"],
      metadata.0.annotations["management.cattle.io/no-default-sa-token"],
      metadata.0.labels["field.cattle.io/projectId"],
    ]
  }
}

resource "helm_release" "argocd" {
  name       = "argocd"
  repository = "https://argoproj.github.io/argo-helm"
  chart      = "argo-cd"
  version    = "5.27.1"
  namespace  = kubernetes_namespace.argocd.metadata.0.name

  max_history = local.helm_max_history

  values = [
    file("${path.module}/helm-values/argocd.yaml")
  ]
}
```

pdrastil (changes requested):
Thanks for the effort but I'm not ok with hardcoding namespace into the manifests.

mkilchhofer:
@jakuboskera can you share your TF code with us? I am sure that helm offers the possibility to inject the namespace.

luigizhou:
Hi everyone, I would also be interested to have this in the Chart.

I don't agree on this being "hardcoding" anything into the manifest as it would still rely on the "Release.Namespace" value that is filled by Helm.

Right now I feel there is a inconsistency in how Helm behaves between doing "helm install" and "helm template" where in the latter case the target Namespace is not computed due to lack of support of this feature from the Helm Community https://github.com/helm/helm/issues/3553

When using helm purely as a template engine, not having the **namespace** field filled in is troublesome as I need to rely on either the kubectl command or any other external tooling to fill in the additional information.

jmeridth:
@mkilchhofer I'm in favor of adding based on your find in your last message.

jakuboskera:
> > Right now I feel there is a inconsistency in how Helm behaves between doing "helm install" and "helm template" where in the latter case the target Namespace is not computed due to lack of support of this feature from the Helm Community [helm/helm#3553](https://github.com/helm/helm/issues/3553)
> > When using helm purely as a template engine, not having the **namespace** field filled in is troublesome as I need to rely on either the kubectl command or any other external tooling to fill in the additional information.
> 
> Ah indeed `helm template` doesn't add the namespace 😮:
> 
> ```shell
> $ helm template argo/argo-cd  --namespace foobar | grep -B 5  "^  namespace:"
>   kind: ClusterRole
>   name: release-name-argocd-application-controller
> subjects:
> - kind: ServiceAccount
>   name: argocd-application-controller
>   namespace: foobar
> --
>   kind: ClusterRole
>   name: release-name-argocd-repo-server
> subjects:
> - kind: ServiceAccount
>   name: release-name-argocd-repo-server
>   namespace: foobar
> --
>   kind: ClusterRole
>   name: release-name-argocd-server
> subjects:
> - kind: ServiceAccount
>   name: argocd-server
>   namespace: foobar
> --
>   kind: Role
>   name: release-name-argocd-application-controller
> subjects:
> - kind: ServiceAccount
>   name: argocd-application-controller
>   namespace: foobar
> --
>   kind: Role
>   name: release-name-argocd-repo-server
> subjects:
> - kind: ServiceAccount
>   name: release-name-argocd-repo-server
>   namespace: foobar
> --
>   kind: Role
>   name: release-name-argocd-server
> subjects:
> - kind: ServiceAccount
>   name: argocd-server
>   namespace: foobar
> --
>   kind: Role
>   name: release-name-argocd-dex-server
> subjects:
> - kind: ServiceAccount
>   name: argocd-dex-server
>   namespace: foobar
> ```
> 
> -> Only RoleBindings are found :(.
> 
> When using helm's skeleton created via:
> 
> ```shell
> $ helm create foo-chart
> Creating foo-chart
> 
> $ head -n 10  templates/deployment.yaml
> apiVersion: apps/v1
> kind: Deployment
> metadata:
>   name: {{ include "foo-chart.fullname" . }}
>   labels:
>     {{- include "foo-chart.labels" . | nindent 4 }}
> spec:
>   {{- if not .Values.autoscaling.enabled }}
>   replicas: {{ .Values.replicaCount }}
>   {{- end }}
> ```
> 
> Namespace is also missing. WDYT @pdrastil / @mbevc1 / @jmeridth / @yu-croco ? In the meantime I'd be okay now to add the Namespace to all namespaced resources.

Thanks @mkilchhofer, that's exactly use case why I fired up this PR. I am using data resource [helm_template](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/data-sources/template) from helm Terraform provider and then apply plain rendered YAML manifests into K8s using kubectl Terraform provider. And due to missing namespace the apply is failing.

Motivation is to apply plain text manifests into K8s instead of `helm release`, because later then the resources are managed via ArgoCD and ArgoCD is not creating helm release.

mkilchhofer:
I'd add it without a conditional as multiple charts in the world supports it:
- almost all bitnami charts, eg
  - postgres: https://github.com/bitnami/charts/blob/main/bitnami/mariadb/templates/primary/statefulset.yaml#L5
- cert-manager chart: https://github.com/cert-manager/cert-manager/blob/v1.11.1/deploy/charts/cert-manager/templates/cainjector-deployment.yaml#L6
- prometheus charts
- [kyverno](https://github.com/kyverno/kyverno/)
- [OPA gatekeeper](https://github.com/open-policy-agent/gatekeeper)

mbevc1:
I think that's okay, but perhaps default should be unset/default 🤔

jakuboskera:
@mbevc1 @mkilchhofer correctly mentioned the the most famous Helm Charts have this set by default. I do not see any disadvantages to not used as it is best practise.

leewoobin789:
@jakuboskera yes looks like. I also have the same issue with the usage of Kustomize, as it seems to execute also the template command.

pdrastil:
After reading the above discussion I'm no longer against this PR because it makes sense when running Helm as template engine used by some other tool that does not allow to define namespace / context explicitly. Thank you for explaining in depth the use-case and pointing out current Helm flaw.

jmeridth (approved):

yu-croco:
FYI to maintainers, I will make same kind of PR to argo-workflows. 🙋

charts/argo-cd/templates/argocd-notifications/servicemonitor.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ template "argo-cd.notifications.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
   {{- if .Values.notifications.metrics.serviceMonitor.namespace }}
   namespace: {{ .Values.notifications.metrics.serviceMonitor.namespace }}
   {{- end }}
```
yu-croco, line 9: I think this should be like below? 🙋 

```diff
- namespace: {{ .Release.Namespace | quote }}
  {{- if .Values.notifications.metrics.serviceMonitor.namespace }}
  namespace: {{ .Values.notifications.metrics.serviceMonitor.namespace }}
+ {{- else }}
+ namespace: {{ .Release.Namespace | quote }}
  {{- end }}
```

yu-croco (changes requested):

charts/argo-cd/templates/argocd-notifications/servicemonitor.yaml
jakuboskera: Yes that's true, thanks for mention that. I've fixed it in https://github.com/argoproj/argo-helm/pull/1937/commits/501d920e168d1b6362e58f0488da20af45a0352a.

charts/argo-cd/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: v2.6.7
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.28.2
+version: 5.28.3
```
mkilchhofer, line 6: Please bump minor chart version and update the changelog line on the bottom of this file

mkilchhofer (changes requested):

charts/argo-cd/templates/argocd-notifications/servicemonitor.yaml
mkilchhofer: What about dropping the if/else and use the default func for that?:

```
namespace: {{ default .Release.Namespace .Values.notifications.metrics.serviceMonitor.namespace | quote }}
```

charts/argo-cd/Chart.yaml
jakuboskera: Chart version updated to `5.28.3` https://github.com/argoproj/argo-helm/pull/1937/commits/96b627ee989d4ed68e6d72bfe4a012af50e14d29 as latest version of ArgoCD Helm Chart in `main` is `5.28.2`. Changelog updated  https://github.com/argoproj/argo-helm/pull/1937/commits/14621b4c713da7fb70235037716b284e783e4358.

charts/argo-cd/templates/argocd-notifications/servicemonitor.yaml
jakuboskera: I am ok with the proposed solution. Should I implemented?

mkilchhofer: Would appreciate it, yes.
Also please make sure that you update all servicemonitor.yaml for all components, I think you missed some of them

charts/argo-cd/Chart.yaml
mkilchhofer: I wrote minor, not patch ;)
So as main is 5.28.2, you should bump to 5.29.0

jakuboskera: Fixed https://github.com/argoproj/argo-helm/pull/1937/commits/bf2e21b9e201491c6c463abe2af443839d944049.

mkilchhofer (approved):
Lgtm 💯

yu-croco (approved):

mbevc1 (approved):

farcaller:
I'm using helm template to install argocd and this resulted in diffs like this:
image

Is it really supposed to be quote-escaped?

Issue #2071, opened 2023-05-22
securitycontext can not be overridden


hsmade:
### Describe the bug

Due to a (very long standing) [bug](https://github.com/helm/helm/issues/9027) in Helm, one can't remove a value  in a map in the values file.
Now because the values.yaml for argo-cd has this:
https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L701
```yaml
  containerSecurityContext:
    runAsNonRoot: true
    readOnlyRootFilesystem: true
    allowPrivilegeEscalation: false
    seccompProfile:
      type: RuntimeDefault
    capabilities:
      drop:
      - ALL
```

and the template [argocd-server/deployment.yaml](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-server/deployment.yaml#L337) has:
```yaml
        securityContext:
          {{- toYaml .Values.server.containerSecurityContext | nindent 10 }}
```
It is impossible to override the security context.
This used to have an `enabled` property that was checked on in the template. Would it be possible to bring that back?

### Related helm chart

argo-cd

### Helm chart version

5.34.1

### To Reproduce

What ever value you specify (`{}` or `null`), where (on `containerSecurityContext` or anything below), it's not possible to remove the defaults.

### Expected behavior

Being able to disable the seccompProfile

mkilchhofer:
It is possible to remove fields of the securityContext, although it isn't possible to remove the complete block with one setting:

```bash
$ cat issue-2071-values.yaml
server:
  containerSecurityContext:
    allowPrivilegeEscalation: null
    capabilities: null
    readOnlyRootFilesystem: null
    runAsNonRoot: null
    seccompProfile: null
```

```bash
$ helm template argo/argo-cd --version 5.34.3  --values issue-2071-values.yaml -s templates/argocd-server/deployment.yaml
---
# Source: argo-cd/templates/argocd-server/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: release-name-argocd-server
# ..
        securityContext:
          {}
```

It's a bit inconvenient, but it works :-)

hsmade:
So I thought as well, but even that didn't work. It does work when I change the charts to use `with`, though. As done in the PR

hsmade:
It indeed still doesn't work with subcharts. This is due to the [bug](https://github.com/helm/helm/issues/9027) with Helm.

PR #2072, opened 2023-05-23
fix(argo-cd): Allow to disable containerSecurityContext

hsmade:
Use `with` so (parts of) the securityContext can be `null`ed out. Fixes https://github.com/argoproj/argo-helm/issues/2071

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.7.2
 kubeVersion: ">=1.22.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.34.3
+version: 5.34.4
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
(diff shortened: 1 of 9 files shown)


hsmade:
wait, this won't work. Because it will try to serialise the `enabled` as well

hsmade:
> 



> Just use `with` instead of `if`. Then for the `toYaml` parameter, you can simply reference `.` and you're good to go :)

Not sure how this will work (not that experienced with Helm per se). But `with` would just change the scope right, it won't allow us to not use the dict values if `enabled` is set to `false`? Because that's what I'm trying to do.

I'm trying to work around a long standing bug in Helm, where it doesn't allow you to override a map.
Right now, it's not possible to disable the security context at all (not even by setting it to `null`).

mkilchhofer (approved):

jmeridth (approved):

hsmade:
(note that while this makes the code more consistent (with the global.securityContext), it doesn't solve the issue linked. When using this chart as a subchart, it still fails to override, due to the (longstanding) bug in Helm

SomniVertix:
@hsmade  - Is there any workaround or link to the helm bug?

mkilchhofer:
Helm refs:
- https://github.com/helm/helm/issues/9136
- https://github.com/helm/helm/issues/9027

PR #2087, opened 2023-05-31
chore(argo-cd): Upgrade supported Kubernetes version to 1.23.0 due to Amazon EKS EoL

yu-croco:
*I will be ready this PR around June 6th.

https://endoflife.date/amazon-eks

image

---

Checklist:


tico24 (approved):

charts/argo-cd/README.md.gotmpl
```diff
@@ -104,6 +104,12 @@ For full list of changes please check ArtifactHub [changelog].
 
 Highlighted versions provide information about additional steps that should be performed by user when upgrading to newer version.
 
+### 5.35.0
+This version supports Kubernetes version from `>=1.22.0-0` to `>=1.23.0-0`. Though the supported version of Kubernetes is v1.24 or later, we align with Amazon EKS because there are a lot of EKS users.
```
mbevc1, line 108: Is this a typo and should it be `>=1.23.0-0` instead of both. Reads a bit confusing now :thinking:

tico24: Apologies for missing this on the review. My brain filled in the words instead of reading them :)

yu-croco: I fixed at https://github.com/argoproj/argo-helm/pull/2087/commits/2b2e351ae1fa33d5f2ea8e0be7a6675ad9fe3a1f 🙋

charts/argo-cd/README.md
```diff
@@ -105,6 +105,11 @@ For full list of changes please check ArtifactHub [changelog].
 
 Highlighted versions provide information about additional steps that should be performed by user when upgrading to newer version.
 
+### 5.35.0
+This version supports Kubernetes version `>=1.23.0-0`. Though the supported version of Kubernetes is v1.24 or later, we align with Amazon EKS because there are a lot of EKS users.
```
mbevc1, line 109: How about something like this? 
```suggestion
This version supports Kubernetes version `>=1.23.0-0`. The current supported version of Kubernetes is v1.24 or later and we align with Amazon EKS calendar, because many of AWS users and conservative approach.
```

charts/argo-cd/README.md.gotmpl
```diff
@@ -104,6 +104,12 @@ For full list of changes please check ArtifactHub [changelog].
 
 Highlighted versions provide information about additional steps that should be performed by user when upgrading to newer version.
 
+### 5.35.0
+This version supports Kubernetes version `>=1.23.0-0`. Though the supported version of Kubernetes is v1.24 or later, we align with Amazon EKS because there are a lot of EKS users.
```
mbevc1, line 108: How about something like this? 
```suggestion
This version supports Kubernetes version `>=1.23.0-0`. The current supported version of Kubernetes is v1.24 or later and we align with Amazon EKS calendar, because many of AWS users and conservative approach.
```

mbevc1:
Thanks, looks better. I've left a potential improvement.  But it's not a blocker.

jmeridth (approved):

mbevc1 (approved):

mbevc1:
Thanks :+1:

agaudreault-jive:
This is kind of annoying to have that configuration in the chart. IMO, the `kubeVersion` defined in this chart should not be coupled to EKS release cycle. Not all users of this chart use EKS. The value should represent the minimum value of the cluster on which it can be installed on, like specified in the Helm doc. If it works on 1.22 and all resources are supported, why make it unusable? 

Because of this change, users deploying argo on older version of Kubernetes will need to fork the chart.

mkilchhofer:
Do you know the support matrix of Argo CD?
While the newest version mostly works also with very old kubernetes versions, it is simply not tested by the upstream project.
The upstream project supports only the 3 recent minor versions of kubernetes.

Refs:
- https://github.com/argoproj/argo-cd/issues/12276
- https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#supported-versions

yu-croco:
In addition to @mkilchhofer 's opinion, the `kubeVersion` itself covers more than Kubernetes's EoL. So I think this change was conservative enough.
https://endoflife.date/kubernetes
image

agaudreault-jive:
>The upstream project supports only the 3 recent minor versions of kubernetes.

While the container image may or may not work with some versions of Kubernetes, this is not the same as the compatibility of a Helm chart. Both do not need to be coupled. The chart can be used with an older version of ArgoCD by overriding the tag. 

The `kubeVersion` field in a chart is a hard limit and should only be set if the chart resources aren't compatible with older Kubernetes versions or container image. AFAIK this chart works fine with older Kubernetes versions, assuming that the image tag is changed to use a suitable ArgoCD image version. 

My suggestion would be to remove `kubeVersion` and update the README with a compatibility table instead of setting a hard limit. 

One change that might require to document such compatibility is the tini/entrypoint change in 2.8 if the chart is not made backward compatible with 2.7 images.

PR #2109, opened 2023-06-15
fix(argo-workflows): Remove hardcoded minio in values.yaml

ariel-snyk:
MinIO isn't the default s3 alternative and its license is limiting ( AGPL-3.0 license) Remove hardcoded 'minio'

Checklist:


charts/argo-workflows/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: v3.4.8
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.29.2
+version: 0.29.3
```
tico24, line 6: This will be a breaking change for current minio users, no? If I'm reading this correctly (which there's every chance I'm not!), they will need to rename their secrets?

charts/argo-workflows/Chart.yaml
```diff
@@ -18,3 +18,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: added
       description: Add support for UI columns configuration
```
yu-croco, line 20: please remove old logs. 🙏

charts/argo-workflows/Chart.yaml
yu-croco: I think @tico24 is right, this PR will affect some users.
I think it should be `0.30.0`.

vladlosev: It's better to specify the secret+key name in `values.yaml`, with template expansion support, with the default being the current value.

ariel-snyk, line 6:
```suggestion
version: 0.30.0
```

vladlosev: That's the only way it makes sense to structure this PR, IMHO.

ariel-snyk: @vladlosev please review, I've exposed another value for secretName with the default being the current value and used template to parse it, let me know if it fits

jmeridth (approved):

charts/argo-workflows/values.yaml
```diff
@@ -706,15 +706,16 @@ artifactRepository:
...
     accessKeySecret:
-      name: "{{ .Release.Name }}-minio"
+      name: "{{ tpl .Values.artifactRepository.s3.secretName . }}"
       key: accesskey
     secretKeySecret:
-      name: "{{ .Release.Name }}-minio"
+      name: "{{ tpl .Values.artifactRepository.s3.secretName . }}"
       key: secretkey
-    # insecure will disable TLS. Primarily used for minio installs not configured with TLS
+    # insecure will disable TLS. Used for s3 comptiable alternatives not configured with TLS
```
vladlosev, line 718:
```suggestion
    # insecure will disable TLS. Used for s3 compatible alternatives not configured with TLS
```

charts/argo-workflows/values.yaml
```diff
@@ -706,15 +706,16 @@ artifactRepository:
   # -- Store artifact in a S3-compliant object store
   # @default -- See [values.yaml]
   s3:
+    secretName: "{{ .Release.Name }}-minio"
     # Note the `key` attribute is not the actual secret, it's the PATH to
     # the contents in the associated secret, as defined by the `name` attribute.
     accessKeySecret:
-      name: "{{ .Release.Name }}-minio"
```
vladlosev, line 712: I am confused now. This is supposed to be a default, and "{{ .Release.Name }}-minio"` is a perfectly good one. Why change it to a template expansion here, especially a recursive one? It's already expanded at the [call site](https://github.com/argoproj/argo-helm/blob/7959fee689ec3a72df5c1d966044c3d93e8f414a/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml#L72).

Same for `secretKeySecret`.

charts/argo-workflows/values.yaml
```diff
@@ -706,15 +706,16 @@ artifactRepository:
...
+      name: "{{ tpl .Values.artifactRepository.s3.secretName . }}"
       key: accesskey
     secretKeySecret:
-      name: "{{ .Release.Name }}-minio"
```
vladlosev, line 715: Same here.

vladlosev (changes requested):

PR #2154, opened 2023-07-08
feat(argo-cd): make ServiceMonitor deployment conditional on CRD existence

alex-souslik-hs:
Checklist:


yu-croco:
Hi  @alex-souslik-hs , thank you for your PR.
Can you please separate this PR into resources(e.g. argo-cd, argo-workflows)?
We make PR by each resource.
Ref: https://github.com/argoproj/argo-helm/pulls?q=is%3Apr+is%3Amereged

Also can you please give us the explanation of motivation for this change?
I am not familiar with it but does `monitoring.coreos.com/v1` have compatibility to specific k8s version?

alex-souslik-hs:
> Hi @alex-souslik-hs , thank you for your PR. Can you please separate this PR into resources(e.g. argo-cd, argo-workflows)? We make PR by each resource, instead of bundling them. Ref: https://github.com/argoproj/argo-helm/pulls?q=is%3Apr+is%3Amereged

It's off course is possible but if you can make an exception for this particular case that'd be great. If not I'll work on splitting.
> Also can you please give us the explanation of motivation for this change? I am not familiar with it but does `monitoring.coreos.com/v1` have compatibility to specific k8s version (I didn't get info for this, so I want the detail) ?

`monitoring.coreos.com/v1` is not a part of the k8s API but comes from the Prometheus Operator. So for example if I deploy a new cluster and I want to install ArgoCD first and Prometheus after I need two value files for ArgoCD.

yu-croco:
@alex-souslik-hs 

>@yu-croco Off course I can but if you can make an exception for this particular case that'd be great. If not I'll work on splitting.

Please separate PR. 

>monitoring.coreos.com/v1 is not a part of the k8s API but comes from the Prometheus Operator. So for example if I deploy a new cluster and I want to install ArgoCD first and Prometheus after I need two value files for ArgoCD.

Thank you for explaining!

alex-souslik-hs:
@yu-croco cleaned up PR to be ArgoCD only, this is the one that was important to me anyway.

charts/argo-cd/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: v2.7.7
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.39.0
+version: 5.39.1
```
mkilchhofer, line 6: Please bump the minor version, as this is not a real bugfix
```suggestion
version: 5.40.0
```

mkilchhofer (changes requested):
Almost LGTM

alex-souslik-hs:
@mkilchhofer done.

yu-croco (approved):
LGTM

pdrastil (approved):
LGTM

mkilchhofer (approved):

batazor:
@alex-souslik-hs Will this work correctly in case prometheus-operator is installed via argocd?  

Now I am installing in this sequence prometheus-operator-crd -> argocd -> prometheus-operator

Won't we have cases that argocd will be installed without installing monitoring components?

alex-souslik-hs:
> @alex-souslik-hs Will this work correctly in case prometheus-operator is installed via argocd?
> 
> Now I am installing in this sequence prometheus-operator-crd -> argocd -> prometheus-operator
> 
> Won't we have cases that argocd will be installed without installing monitoring components?

@batazor 
If you pre-install the CRDs you'll have the same behavior as before, the idea is to allow you to install ArgoCD first and deploy everything else with it without modifying its values.

@yuha0
It's important to note the optional dependency `redis-ha` chart works the same way, so it's possible you already had this behavior on some of your `ServiceMonitors`.

FalconerTC:
I believe this change has caused my ArgoCD service monitors to disappear. I manage ArgoCD with ArgoCD and after this upgrade I am no loner seeing any ArgoCD Service Monitors (I am however seeing ServiceMonitor for argo rollouts, but I see that's because this capabilities check isn't set there). Is there something in particular I should do to mock `Capabilities.APIVersions.Has "monitoring.coreos.com/v1`  when managing argocd with argocd?

mkilchhofer:
Yup some tools like Kustomize and Helmfile do not pass the available APIversions to helm. In your mentioned issue above, there is also a comment which explains how to pass APIversions:

> Example:
> 
> ```yaml
> namespace: foo
> 
> helmCharts:
> - name: bar
>   releaseName: bar
> 
>   #...
>   apiVersions:
>     - apiregistration.k8s.io/v1
>   #...
> ```

PR #2166, opened 2023-07-16
feat(argo-cd): allow configurable hpa metrics

### LINKED ISSUE #2152: argocd HPAs should support user provided metrics

ArgoCD helm chart is "hard coded" to use only pod Resource metrics for CPU and memory based on usage vs Requests. Users may have other metrics such as custom prometheus metric aggregations, that they'd rather use. The helm templates for the HPA resources should support user customization of the metric types that are permitted by HPA v2

### Related helm chart

argo-cd

User supplied values templating of the HPA spec.metrics fields

Currently we are disabling the chart autoscaling parameters and then using the values.extraObjects to define our own HPAs. This works, but due to issues with 'helm templating within helm templating' its hard to make these fully interopt with the rest of the chart so its a very custom solution to each use case.

### Additional context

_No response_

zswanson:
Fixes #2152 
Accepts a user defined list of HPA metrics for the server and repoServer HPA when autoscaling/HA is enabled. There is no change to current behavior, which defaults to use the Resource metrics for CPU and memory. Because the metrics is block is highly configurable I've opted to just take the entire block as user provided yaml - the user must provide the full and correct syntax.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.7.7
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.39.0
+version: 5.40.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: added
       description: Allow configuring Dex's init image resources separately
+    - kind: changed
+      description: Allow configurable metrics in server and repoServer HPAs
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -576,6 +576,7 @@ NAME: my-release
 | repoServer.autoscaling.behavior | object | `{}` | Configures the scaling behavior of the target in both Up and Down directions. This is only available on HPA apiVersion `autoscaling/v2beta2` and newer |
 | repoServer.autoscaling.enabled | bool | `false` | Enable Horizontal Pod Autoscaler ([HPA]) for the repo server |
 | repoServer.autoscaling.maxReplicas | int | `5` | Maximum number of replicas for the repo server [HPA] |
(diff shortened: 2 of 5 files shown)


charts/argo-cd/README.md
```diff
@@ -576,6 +576,7 @@ NAME: my-release
 | repoServer.autoscaling.behavior | object | `{}` | Configures the scaling behavior of the target in both Up and Down directions. This is only available on HPA apiVersion `autoscaling/v2beta2` and newer |
 | repoServer.autoscaling.enabled | bool | `false` | Enable Horizontal Pod Autoscaler ([HPA]) for the repo server |
 | repoServer.autoscaling.maxReplicas | int | `5` | Maximum number of replicas for the repo server [HPA] |
+| repoServer.autoscaling.metrics | list | `[]` | Configures custom HPA metrics for the Argo CD server Ref: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/ |
```
mkilchhofer, line 579:
```suggestion
| repoServer.autoscaling.metrics | list | `[]` | Configures custom HPA metrics for the Argo CD repo server Ref: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/ |
```

charts/argo-cd/Chart.yaml
```diff
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: added
       description: Allow configuring Dex's init image resources separately
```
mkilchhofer, line 30: Drop the old change. Only your change needs to stay here.

mkilchhofer (changes requested):
Almost LGTM ;-)

charts/argo-cd/README.md
zswanson: Oh, I need to change that in the Values file too then, since this was auto generated.

charts/argo-cd/Chart.yaml
zswanson: Fixed

mkilchhofer (approved):
🚀

tico24 (approved):

mbevc1 (approved):

zswanson:
Thanks for the quick review!

Issue #2171, opened 2023-07-19
[argo-cd] SAML login broken when using cert-manager


ShimiTaNaka:
### Describe the bug

When creating a Certrificate resource for the ingress, and configuring okta (for example)
SAML get broken.
In the SAML configuration, the dexURL (according to the docs) is https://ui.argocd.yourorganization.net/api/dex/callback ([ref](https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/okta/))
The URL actually points to the external endpoint of ArgoCD server, which is secured with SSL on the Ingress.
This creates an issue since the `argocd-dex-server-tls` mount to argocd-server is using a static secret name.
If we create a Certificate, we need to also modify the secretName for dex server accordingly

### Related helm chart

argo-cd

### Helm chart version

5.31.0

### To Reproduce

1. Install ArgoCD with certificate enable with private issuer
2. Configure Okta SAML using https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/okta/
3. Login

### Expected behavior

Login should be successful without x.509 SSL error

mkilchhofer:
Can you explain where do you see this x.509 erorr?
please send a screenshot or log output

PR #2172, opened 2023-07-19
fix(argo-cd): Use ingress crt for dex when enabled

ShimiTaNaka:
Fixes: https://github.com/argoproj/argo-helm/issues/2171

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.7.7
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.41.0
+version: 5.41.1
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: Make ServiceMonitor deployment conditional on CRD existence
+    - kind: fixed
+      description: Fix SAML auth when using private CA with Ingress
--- a/charts/argo-cd/templates/argocd-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-server/deployment.yaml
@@ -303,6 +303,11 @@ spec:
           name: styles
         - mountPath: /tmp
           name: tmp
+        {{- if .Values.server.certificate.enabled -}}
+        - mountPath: /etc/ssl/certs/ingress-ca-certificates.crt
+          name: ingress-ca-certificate
+          subPath: ca.crt
+        {{- end }}
         {{- if .Values.server.extensions.enabled }}
         - mountPath: /tmp/extensions
           name: extensions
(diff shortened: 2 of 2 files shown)


tico24:
It's failing at exactly the point you've changed, so I'd argue it's related to your change :)

templates/argocd-server/deployment.yaml: unable to parse YAML: error converting YAML to JSON: yaml: line 336: mapping values are not allowed in this context

ShimiTaNaka:
ok, so it's line 336 after rendering, because the chart itself change is in line 426 :)
Checking

mkilchhofer:
I am not sure if this is needed. Dex is controlled by a binary which is copied into the dex container. I thought that this utility does generate some TLS config between argo-cd-server and dex. And dex is exposed via a reverse proxy component inside argo-cd-server

ShimiTaNaka:
It actually is
specifically when setting up OKTA.
The change doesn't require anything on the dex itself actually and dex isn't even configured with SSL https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L994C12-L994C12
Instead, we configure ingress for argocd-server.
During the login, argocd-server makes an API endpoint to:
https://ui.argocd.yourorganization.net/api/dex/.well-known/openid-configuration
and then there is an SSL issue.
I can't reproduce it on a working environment, seems that the CA is cached, but it's shown in the server logs:
`Initializing OIDC provider (issuer: https://ui.argocd.yourorganization.net/api/dex)"`
(obviously with our private URL) and there was an SSL error, it was fixed once we changed the secret name in question

ShimiTaNaka:
Important to mention, this does not create a certificate for DEX, rather, if we create an ingress to ArgoCD *server*, with a certificate, map it.

ShimiTaNaka:
I think I've found a better solution
I'll check it out first

ShimiTaNaka:
Current implementation inspired by https://github.com/argoproj/argo-cd/issues/7572#issuecomment-1057376181

I also hope it makes the incentive clearer :)

mkilchhofer:
I can still not follow your issue.
For ingress traffic from outside kubernetes to dex (user callbacks), the path chain is:

Ingress -> argocd-server (acts as a proxy) -> dex

For outgoing traffic from dex to the upstream IdP, traffic flows directly:

Dex -> IdP

ShimiTaNaka:
I have tested it
and I know it can be achieved by it, it's just makes things simpler.
The login flow is as follows:

1. User goes to https://argocd.domain.suffix and clicks "Sign in with Okta"
2. Argo-server sends him to ssoUrl (As specified in https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/okta/)
3. User logs in, and gets a redirect to https://argocd.domain.suffix/api/dex/callback
4. User gets authenticated with a session (for https://argocd.domain.suffix site) and gets 303 redirect to https://argocd.domain.suffix
5. argo-server makes API call to https://argocd.domain.suffix/api/dex/.well-known/openid-configuration to validate the session cookie signature

The last part fails because argo-server communicates with the dex via external call all the way with the ingress and not internally, so when the ingress is using self signed CA, the call fails.

As you've mentioned, this can be also achieved by passing the values you've provided, the PR as it is just makes it easier on the user

I will try to reproduce and get back to you on it

PR #2183, opened 2023-07-23
docs(argo-workflows): add an HA section

agilgur5:
## Summary

~Create a `values-ha.yaml` for Argo Workflows~
Add HA docs, example, and test for Argo Workflows

## Details

- ~`values-ha.yaml`~ `ci/ha-values.yaml` is an example and test for how one would set-up HA according to the [High Availability](https://argoproj.github.io/argo-workflows/high-availability/) doc (which I have contributed to)
  - PDB and min 2 replicas set
  
- Add a "High Availability" section to the README, similar to [Argo CD's chart](https://github.com/argoproj/argo-helm/blob/3e35b0c7f7d758d553b17f369cc5940484ef5d89/charts/argo-cd/README.md?plain=1#L17)
  
## Review Notes

This is probably the least environment-specific piece of my Argo Workflows config, so figure I could contribute this and others could add to it 😅 
- ~`values-ha.yaml` is not a Helm convention per se, but is something I've seen before, albeit not _that_ frequently. As such, totally feel free to reject and close this as out of scope!~ **EDIT**: changed as requested [in review](https://github.com/argoproj/argo-helm/pull/2183#discussion_r1271800523)

## Future Work

1. Another piece I could add here that is fairly env-independent is `topologySpreadConstraints` to prefer multi-node and multi-zone placement of replicas
    - or `podAntiAffinity` as the alternative
1. `priorityClassName` with some common default name or just commented out?
    - these definitely vary per distribution / installation / configuration of k8s, which is why I didn't add it in this PR
1. Perhaps more opinionated / controversial, but another best practice would be to add GC per the [Cost Optimization](https://argoproj.github.io/argo-workflows/cost-optimisation/) doc (which I have contributed to as well)

Checklist:


charts/argo-workflows/values-ha.yaml
```diff
@@ -0,0 +1,23 @@
+# Sample values for High Availability configuration, following https://argoproj.github.io/argo-workflows/high-availability/
```
mkilchhofer, line 1: I don't like this approach to have multiple values.yaml inside the chart root. But we can follow the approach from Argo CD for sure:
- Place a section into the Charts README
- move this file into `ci/`. This ensures, that our CI workflow tests with these values

tico24: FYI, I believe it has to end in -values.yaml for the ci to pick it up (eg ha-values.yaml).

agilgur5: Ah did not realize Argo CD had a section on this. I can make those changes.
Would probably prefer to link from the a new README section to the `ci/` files instead of having a [duplicate code block](https://github.com/argoproj/argo-helm/blob/962342fe2acef6022ac5c1a3eb352b336308b3eb/charts/argo-cd/README.md?plain=1#L27) in the docs though.

> FYI, I believe it has to end in -values.yaml for the ci to pick it up (eg ha-values.yaml).

I was wondering how these got picked up, so great note!

agilgur5: Updated as per above

charts/argo-workflows/Chart.yaml
```diff
@@ -16,5 +16,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: added
-      description: Add support for executor args
+    - kind: fixed
+      description: Add HA docs and example
```
agilgur5, line 20: This is now a docs and tests only change, so actually not sure if this necessary to bump anymore?

I used "fixed" because ArtifactHub doesn't support "docs" (["valid supported kinds are added, changed, deprecated, removed, fixed and security."](https://artifacthub.io/docs/topics/annotations/helm/))

agilgur5:
Could this get a re-review? cc @mkilchhofer

mkilchhofer (approved):

jmeridth (approved):

yu-croco:
Hi @agilgur5 , can you please resolve conflict ? 🙋

agilgur5:
`Chart.yaml` conflict resolved

yu-croco (approved):

Issue #2202, opened 2023-08-03
Merging of global and individual component configs is not working properly


jmeridth:
### Describe the bug

Merging of global component config options and individual component config options are not working.

### Related helm chart

argo-cd

### Helm chart version

v5.42.1 (latest as of filing this issue)

### To Reproduce

Update your `values.yaml` in an umbrella chart with the following:

```yaml
## Argo CD Specific Configurations
argo-cd:
  configs:
    params:
      repoServer.log.level: debug
```

results in `argocd-cmd-params-cm.yaml` config map having the following:

```yaml
apiVersion: v1
data:
  ...
  repoServer.log.level: info # default global
  repoServer.log.level: debug #individual component config
  ...
```

bug: `repoServer.log.level` is shown twice

### Expected behavior

Update your `values.yaml` in an umbrella chart with the following:

```yaml
## Argo CD Specific Configurations
argo-cd:
  configs:
    params:
      repoServer.log.level: debug
```

should result in `argocd-cmd-params-cm.yaml` config map having the following:

```yaml
apiVersion: v1
data:
  ...
  repoServer.log.level: debug # individual component config should overwrite global
  ...
```

`repoServer.log.level` should only be there once

### Additional context

Code that needs to be investigated is the [`argo-cd.config.params`](https://github.com/argoproj/argo-helm/blob/962342fe2acef6022ac5c1a3eb352b336308b3eb/charts/argo-cd/templates/_helpers.tpl#L206-L215) helper function.  It uses the [`argo-cd.config.params.presets`](https://github.com/argoproj/argo-helm/blob/962342fe2acef6022ac5c1a3eb352b336308b3eb/charts/argo-cd/templates/_helpers.tpl#L187C13-L187C42) helper function to get global config options and then merges with overwrite.

mkilchhofer:
@pdrastil I slightly remember that you already pointed in this direction, right?

pdrastil:
If I remember correctly I've tried to unify `--set "repoServer.log.level"=debug` and `--set repoServer.log.level=debug` that produced double entries as key looks same but Helm treats it differently. This could be some bug in template where merge is not done with defaults.

PR #2205, opened 2023-08-04
fix(argo-cd): Template configuration keys as a string instead of map

pdrastil:
Resolves:

https://github.com/argoproj/argo-helm/issues/2202

Explanation:

Helm treats text value of `repoServer.log.level` and similar keys as a map that will return a false for simple test `{{ hasKey  $presets "repoServer.log.level" }}`. To force correct behaviour items must be stored in the dict before serialization.

Checklist:

### DIFF
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -162,9 +162,11 @@ Create the name of the notifications service account to use
 Argo Configuration Preset Values (Incluenced by Values configuration)
 */}}
 {{- define "argo-cd.config.cm.presets" -}}
+{{- $presets := dict -}}
 {{- if .Values.configs.styles -}}
-ui.cssurl: "./custom/custom.styles.css"
+{{- $_ := set $presets "ui.cssurl" "./custom/custom.styles.css" -}}
 {{- end -}}
+{{- toYaml $presets }}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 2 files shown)


yu-croco (approved):
LGTM 🚀

jmeridth (approved):

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -183,24 +185,25 @@ Merge Argo Configuration with Preset Configuration
...
+{{- $_ := set $presets "repo.server" (printf "%s:%s" (include "argo-cd.repoServer.fullname" .) (.Values.repoServer.service.port | toString)) -}}
+{{- $_ := set $presets "server.repo.server.strict.tls" (.Values.repoServer.certificateSecret.enabled | toString ) -}}
+{{- $_ := set $presets "redis.server" (include "argo-cd.redis.server" .) -}}
+{{- if .Values.dex.enabled -}}
+{{- $_ := set $presets "server.dex.server" (include "argo-cd.dex.server" .) -}}
+{{- $_ := set $presets "server.dex.server.strict.tls" .Values.dex.certificateSecret.enabled -}}
+{{- end -}}
+{{- range $component := tuple "applicationsetcontroller" "controller" "server" "reposerver" -}}
+{{- $_ := set $presets (printf "%s.log.format" $component) $.Values.global.logging.format -}}
+{{- $_ := set $presets (printf "%s.log.format" $component) $.Values.global.logging.format -}}
```
milpog, line 201: @pdrastil I think you've introduced a regression by removing possibility to set `logging.level` globally. Lines 200 and 201 are the same but `201` should look like

```yaml
{{- $_ := set $presets (printf "%s.log.level" $component) $.Values.global.logging.level -}}
```

pdrastil: @milpog - You are right, thanks for letting me know. Fix is on the way.

milpog: Thank you 🤗

Issue #1768, opened 2023-01-17
Distribute Charts in OCI format


sabre1041:
### Is your feature request related to a problem?

No

### Related helm chart

argo-cd, argo-events, argo-rollouts, argo-workflows, argocd-image-updater, argocd-apps

### Describe the solution you'd like

Helm supports multiple distribution methods, traditional Helm repositories (http based) and the newer [OCI Registry](https://helm.sh/docs/topics/registries/).

Provide support for distributing charts as OCI artifacts within an OCI registry

pdrastil:
Maybe this https://github.com/appany/helm-oci-chart-releaser can be used to support this

geowalrus4gh:
we are moving to OCI registries for both images and helm charts. So it would be super easy if argo charts are published to registries in OCI format along with the respective images.

PR #2209, opened 2023-08-07
feat(github): add Helm OCI support

dirien:
This PR add support to create OCI helm charts. No new feature is added to the existing Helm charts.

This PR continues #1768

Checklist:

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "charts/**"
 
 permissions:
   contents: read
@@ -12,6 +14,7 @@ jobs:
   publish:
     permissions:
       contents: write  # for helm/chart-releaser-action to push chart release and create a release
+      packages: write  # to push OCI chart package to GitHub Registry
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -60,3 +63,20 @@ jobs:
           config: "./.github/configs/cr.yaml"
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Login to GHCR
+        uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push chart to GHCR
+        run: |
+          shopt -s nullglob
+          for pkg in .cr-release-packages/*.tgz; do
+            if [ -z "${pkg:-}" ]; then
+              break
+            fi
+            helm push "${pkg}" oci://ghcr.io/${{ github.repository }}
+          done


dirien:
CC: @sabre1041

.github/workflows/publish.yml
```diff
@@ -60,3 +63,20 @@ jobs:
           config: "./.github/configs/cr.yaml"
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Login to GHCR
+        uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
+        with:
+          registry: ghcr.io
+          username: ${ GITHUB_REPOSITORY_OWNER }
```
sabre1041, line 71: Recommend changing to `github.actor` as it is suggested by the `login-action` [documentation](https://github.com/docker/login-action#github-container-registry)

.github/workflows/publish.yml
```diff
@@ -60,3 +63,20 @@ jobs:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push chart to GHCR
```
sabre1041, line 74: As we are looking to publish the charts as OCI artifacts, I am wondering is there is also an interest in signing the charts using Sigstore and `cosign`. Other aspects of the Argo ecosystem are signing content and it would be good if we were able to reuse a lot of the same paradigms here.

The following is a snippet that illustrates how to not showcase how to publish the charts as OCI artifacts, but also how to sign them using Sigstore

https://github.com/sigstore/helm-charts/blob/main/.github/workflows/release.yml#L73-L83

sabre1041 (changes requested):
@dirien Added a few comments and suggestions

.github/workflows/publish.yml
dirien: I thought about the same but wanted to create a separate PR for  it! For now this PR is to get OCI charts in place

sabre1041: Would we want to start with unsigned charts and then move to signed charts or tackle both at the same time. I defer to the project maintainers and community on the thoughts of the collective group

.github/workflows/publish.yml
dirien: Done! @sabre1041 thanks for spotting this!

.github/workflows/publish.yml
dirien: I am fine with booth! If the maintainers want it added into this PR happy to do so.

dirien:
Thanks @sabre1041 for the comments. Appreciate your feedback! Let's get this done and we finally have OCI charts for Argo! 🥳

.github/workflows/publish.yml
mkilchhofer: I think we can skip cosign for now. We should clarify with  argoproj maintaines what key to use for this.

But can we use the codeblock from cosign mentioned above? Your PR have two issues
- `find` needs to filter for tgz only as we also generate the provenance file for the old non-OCI index
- `/charts` doesn't exists as a repo in github.com/argoproj

dirien: Hey @mkilchhofer,

- Done for filtering only the *.tgz! Thanks for the hint!

- /charts will be created and is part of the name. The finale OCI artefact name would look like this e.g: `oci://ghcr.io/argo-helm/charts/argo-cd`

mkilchhofer: I think the first element after
ghci.io would be `argoproj` then.

I prefer having `ghcr.io/argoproj/argo-helm` for consitency:

```
helm push "${pkg}" oci://ghcr.io/${GITHUB_REPOSITORY}
```

dirien: Sounds good! Changed it.

mkilchhofer (approved):
Lgtm. Thanks for implementing this @dirien !!🥳

sabre1041:
Glad to see this get pushed through. thanks again @dirien

mkilchhofer:
Wohoooo, first chart arrived via OCI ❤️ :
https://github.com/argoproj/argo-helm/pkgs/container/argo-helm%2Fargo-cd

thanks again @dirien

sabre1041:
@dirien @mkilchhofer was thinking about the same. I would wait until we have official documentation on consumption. I would also like to continue the Sigstore discussion and can help lead that if desired on the signing requirements for the charts.

dirien:
@sabre1041 What about meeting next year and discuss the next steps regarding sigstore integration?

sabre1041:
> @sabre1041 What about meeting next year and discuss the next steps regarding sigstore integration?

Certainly up for the discussion. While I do advocate for the use for sigstore, lets have a discussion on methods for the incorporation of additional signing methods

PR #2215, opened 2023-08-10
feat(argo-workflows): add configurable liveness+readiness probes to server

byxorna:
Adds a sane default `livenessProbe` for `server`, and allows both `livenessProbe` and `readinessProbe` to be overridden from `values.yaml`.

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: v3.4.9
+appVersion: v3.4.10
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
@@ -16,5 +16,7 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
+    - kind: added
+      description: Add support for overriding server livenessProbe, readinessProbe
     - kind: added
       description: Add support for executor args
--- a/charts/argo-workflows/README.md
+++ b/charts/argo-workflows/README.md
@@ -277,6 +277,7 @@ Fields to note:
 | server.ingress.pathType | string | `"Prefix"` | Ingress path type. One of `Exact`, `Prefix` or `ImplementationSpecific` |
 | server.ingress.paths | list | `["/"]` | List of ingress paths |
 | server.ingress.tls | list | `[]` | Ingress TLS configuration |
+| server.livenessProbe | object | httpGet on 2746 | Configure liveness [probe] for the server |
 | server.loadBalancerIP | string | `""` | Static IP address to assign to loadBalancer service type `LoadBalancer` |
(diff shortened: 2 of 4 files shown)


charts/argo-workflows/Chart.yaml
```diff
@@ -16,5 +16,7 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
+    - kind: added
+      description: Add support for overriding server livenessProbe, readinessProbe
     - kind: added
       description: Add support for executor args
```
yu-croco, line 22: Please remove the old one. :)

jmeridth:
I'd prefer the defaults in values.yaml instead of hard coded in the template. Wdyt? @yu-croco @mkilchhofer opinions?

yu-croco:
I agree with @jmeridth 🙋 .
It's nice to follow [argo-cd's values.yaml](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L1053-L1079) ?

vladlosev:
My first reaction also was a request to move these definitions into the defaults but then I realized that doing that will make it quite hard for users to override a check in some cases - for example, changing a default HTTP check to an exec-type check.

byxorna:
I opted to leave the defaults in the deployment, because the server's `readinessProbe` had some templating in it (and it felt odder to keep `readinessProbe`'s default in the deployment and define the `livenessProbe` in values.yaml. ([ref](https://github.com/argoproj/argo-helm/pull/2215/files#diff-080817188de6465125c7b1c37527c595614445431d276389c2ad0b4cecacce35R89-R93)). I am happy to change to whatever the preference here is, just let me know. If we are ok with setting a default HTTP scheme for readinessProbe, I can pull them both out into values.yaml to align better with Argo CD's chart.

mkilchhofer:
> I'd prefer the defaults in values.yaml instead of hard coded in the template. Wdyt? @yu-croco @mkilchhofer opinions?

I think if we need more flexibility than having everything in values.yaml (when the use case of changing from `httpGet` to `exec` needs to be possible) we should use something like the bitnami charts:
https://github.com/bitnami/charts/blob/14e96f8e97d849fff282fe8cc2c2d8a74cce5c38/bitnami/ghost/templates/deployment.yaml#L238

vladlosev:
The bitnami approach complicates configuration for users too much, IMHO.

charts/argo-workflows/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: v3.4.9
+appVersion: v3.4.10
```
agilgur5, line 2: 3.4.10 is not released yet: https://github.com/argoproj/argo-workflows/issues/11552.
Will probably be out after Tuesday's contributor meeting

jmeridth: https://github.com/argoproj/argo-helm/pull/2220

PR #2249, opened 2023-09-05
feat(argo-cd): adds toggle for helm-working-dir

ahamilton55:
- Adds a toggle to be able to turn off the helm-working-dir for the repo server deployment. Using a shared helm repo storage directory can cause issues when multiple helm commands are being run in parallel. The repo server also has the ability to rebuild the repos and do updates in the normal flow for checking the status of an application so it won't cause issues if it's disabled.

We've been having issues with collisions happening in the shared helm-working-dir and with the index.yaml file being empty when read by one of the processes. The repo server code is setup to rebuild the repo and do a chart build for each templating run for us so it makes more sense to allow it to utilize the temporary space. I didn't want to remove it for everyone so I made the default to enabled with the option to disable it. Since testing this change in one of our environments we've had no issues with services consistently entering the Unknown state.

Checklist:

### DIFF
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -2177,6 +2177,9 @@ repoServer:
   #  - name: cmp-tmp
   #    emptyDir: {}
 
+  # -- Toggle the usage of a ephemeral Helm working directory
+  useEphemeralHelmWorkingDir: true
+
   # -- Annotations to be added to repo server Deployment
   deploymentAnnotations: {}
 
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.8.3
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.45.5
+version: 5.46.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: added
-      description: Documented scheduling parameters for redis-ha
+      description: added a toggle for the shared Helm working directory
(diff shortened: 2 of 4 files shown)


tico24:
Can you please update the changelog?

ahamilton55:
@tico24 Updated.

tico24 (approved):

charts/argo-cd/values.yaml
```diff
@@ -2152,6 +2152,11 @@ repoServer:
   #  - name: cmp-tmp
   #    emptyDir: {}
 
+  # -- Toggle the usage of a shared Helm working directory
+  helmWorkingDir:
```
pdrastil, line 2156: This is map, not a feature toggle. Are we expecting another set of parameters that will be inside of map or is it just for enabling / disabling the directory? WDYT @mkilchhofer @jmeridth

ahamilton55: I think I originally had this setup as just a single boolean variable but thought it looked cleaner and like other values with and `enabled`. I also thought that perhaps making the path configurable in the future could be something that people might want so allowing it to expand could be helpful.

I can change the comment or make all of this into a single variable.

pdrastil: The reason why I'm asking the rest of the team is that we had this historically in chart roughly 1 year ago for multiple features that were a map with single parameter and nothing else. These sections were eventually moved either to simple global toggles in top of the values.yaml or into cm and params sections if possible. Only one that I survived this cleanup is currently `openshift.enabled` until I can do some regression on OpenShift cluster.

mkilchhofer: I think we should use a top-level `.Values.useEphemeralHelmWorkingDir` (or similar 😉) bool or move it below the `repoServer:` structure.

ahamilton55: My vote would be to keep this inside of the `repoServer` structure since it only corresponds to that service and recreate it as just as a single boolean value. I can update the PR if this is the direction you want to go towards.

mkilchhofer: > My vote would be to keep this inside of the `repoServer` structure since it only correspo

Sure 👍 the only thing is that we don't preffer a subkey `repoServer.bla.enabled` if there is only an enabled/disabled option below the `bla` subkey. Subkey is fine if there are multiple options down there.

ahamilton55: Okay I updated it based on a few of the above comments and turned it into a single boolean under the `repoServer` section.

mkilchhofer (approved):

yu-croco (approved):

PR #2256, opened 2023-09-08
fix(argo-cd): Add applicationsets resource to argocd-server cluster role

JorTurFer:
argocd-server tries to list applicationsets accross all the cluster in v2.8:
```
argocd appset list
WARN[0000] Failed to invoke grpc call. Use flag --grpc-web in grpc calls. To avoid this warning message, use flag --grpc-web.
FATA[0000] rpc error: code = PermissionDenied desc = error listing ApplicationSets with selectors: applicationsets.argoproj.io is forbidden: User "system:serviceaccount:argocd:argocd-server" cannot list resource "applicationsets" in API group "argoproj.io" at the cluster scope
```

That permission is granted within the argo namespace but is missing in the cluster role. This PR adds it to the cluster role

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.8.3
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.45.3
+version: 5.45.4
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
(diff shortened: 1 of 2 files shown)


tico24:
[Unless I'm looking in the wrong place](https://github.com/argoproj/argo-cd/blob/v2.8.3/manifests/cluster-rbac/server/argocd-server-clusterrole.yaml), I don't see this in the official manifests either, so I'm not sure it should go into the helm chart.

JorTurFer:
That's totally true, but why has the server requested me that permission? The message has gone after adding it.
I'll check it there and raise another PR there for this topic. Could we keep this PR open until that? You can close it if you prefer

JorTurFer:
I close this PR because the root cause seems to be in the server and not in helm: https://github.com/argoproj/argo-cd/issues/15429

PR #2299, opened 2023-10-17
feat(argo-cd): Allow chart to be deployed and only ship CRDs

cypher7682:
Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: fixed
-      description: Sync redis / redis-ha readOnlyRootFilesystem=true option from upstream. This was part of Argo CD 2.8.0.
+    - kind: added
+      description: Added capability to deploy only CRDs, for cases where remote clusters need argo CRDs


cypher7682:
When deploying a separate cluster that is going to be managed by a remote argo, CRDs are needed for it to ship anything. This PR allows CRDs to be published without anything else in the chart coming with it.

mkilchhofer:
I don't like this idea to be honest (too noisy on all resources and never seen this approach in the "wild"). If this is something we need to support I would be on the side to outsource the CRDs to a subchart.

What is the issue in pulling in the CRDs from the upstream repo as a Kustomize base like its described in the README.md?

cypher7682:
> When I saw this PR, my first thoughts were:

I had the same thoughts doing this PR... don't worry. It's clunky AF; I am under no illusions 😂  If new charts are made, they'll need to make sure they have the `{{ if }}` otherwise CRD installations will randomly install miscellaneous apps.

A sub chart is definitely a better idea. I don't think the versioning matters all that much, helm wise. So far the versioning doesn't exist, so even starting at 0.0.1 would work. Existing deployments can just reference the sub-chart on 0.0.1. However...

Regarding the double PRs, I don't see a way to avoid that, to be honest; at least not without essentially guessing the version before it's shipped. On the other hand, copy pasting everything into a "slimmed down" chart is also not ideal.

I, too, would be interested in hearing from the maintainers on what their thoughts are on the best way forward.

jmeridth:
As @mkilchhofer has stated, your use case is valid but maintaining it would currently be painful. I'm thinking on proposals here and other ideas.

cypher7682:
Do you want me to move this to an issue, rather than a PR?

jmeridth:
@cypher7682 As @mkilchhofer stated

> With my reflection about complexity this proposal here might look a bit clunky, but it is a KISS solution :D .

I'm good with this but can we do a PR per chart please?  Can you also sign your commits (`git commit -s`)?

jmeridth:
@cypher7682 please wait before you break out the PRs.  We, the maintainers, are discussing options.  Will let you know asap.

PR #2354, opened 2023-11-22
feat(argo-workflows): Make workflow controller ConfigMap optional

qa-florian-wende:
This makes the Argo workflow controller ConfigMap optional and allows its name to be overridden
* I have a use-case where my ConfigMap is created by Terraform but Terraform doesn't manage the helm-chart. In this case, the helm-chart would still create a configMap and pass its name to workflow-controller and server in the args, so the ConfigMap created by Terraform would be overwritten or would not be used.
* In any case, I think it's best practice to have a value to disable the creation of all non-essential components such as configuration.
* By default, a new configMap is created and named as before, so this is backwards compatible.
* If you choose not to create a ConfigMap, the flag "--configMap" with the given or default name is still added to the args of workflow-controller and server (I assume, users will always need a ConfigMap, they might just not want to create it using this helm-chart.

Checklist:


charts/argo-workflows/templates/_helpers.tpl
```diff
@@ -97,6 +97,13 @@ app.kubernetes.io/component: {{ .component }}
 {{- end }}
 {{- end }}
 
+{{/*
+Create the name of the controller configMap
+*/}}
+{{- define "argo-workflows.controller.config-map.name" -}}
+{{- .Values.controller.configMap.name | default (printf "%s-%s" (include "argo-workflows.controller.fullname" .) "configmap") | trunc 63 | trimSuffix "-" -}}
```
yu-croco, line 104: I wonder how Argo Workflows controller refers the ConfigMap if `.Values.controller.configMap.name` is configured. 🤔

yu-croco: Ah this define is included in `charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml` as well.

charts/argo-workflows/templates/controller/workflow-controller-deployment.yaml
```diff
@@ -45,7 +45,7 @@ spec:
           command: [ "workflow-controller" ]
           args:
           - "--configmap"
-          - "{{ template "argo-workflows.controller.fullname" . }}-configmap"
+          - {{ template "argo-workflows.controller.config-map.name" . | quote }}
```
yu-croco, line 48: I wonder we should support for the case, `controller.configMap.create = false` ? 🤔

yu-croco (approved):
LGTM

vladlosev (approved):
LGTM

tico24 (approved):

PR #2390, opened 2023-12-15
fix(argo-cd): Resolve metrics label mismatch between ServiceMonitor and controller

jessebye:
This PR resolves a mismatch in the labelSelector used in the ServiceMonitor vs. the actual labels for the controller.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: Upgrade Argo CD to v2.9.3
+    - kind: changed
+      description: Resolve mismatch between ServiceMonitor labels and controller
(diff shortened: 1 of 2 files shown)


mkilchhofer (changes requested):
This selector is fine in my opinion.
The selector of this ServiceMonitor selects the Service here by the exact same label:
https://github.com/argoproj/argo-helm/blob/77a4196fc31294b7abeb1d4d30df136f2b77a675/charts/argo-cd/templates/argocd-application-controller/metrics.yaml#L8

Can you describe your issue you have? Or is it just for consistency in your eyes?

jessebye:
@mkilchhofer you're completely right! I was trying to track down why our Prometheus instance isn't picking up the `ServiceMonitor` and thought this might be the problem (I was mistakenly looking at the labels on the pod, not the service :facepalm:). But now you've mentioned the Service, I realize I was mistaken.

mkilchhofer:
👍 
You need to have a look on the labels of the ServiceMonitor directly.
Most users use kube-prometheus-stack helm chart.
They need to match the `kind: Prometheus` object inside your cluster.
Use `kubectl get prometheus -o yaml` and have a look for `.spec.serviceMonitorSelector`. Most probably it isn't matching 😉

PR #2400, opened 2023-12-23
feat(argo-cd): Add Probes for redis

Farfaday:
Resolves https://github.com/argoproj/argo-helm/issues/2464

Checklist:


charts/argo-cd/templates/redis/_configs.tpl
```diff
@@ -0,0 +1,29 @@
+{{/* vim: set filetype=mustache: */}}
```
pdrastil, line 1: Why do we need extra tpl file when templates are used used just once inside the ConfigMap and templating contains only redis container port? This can be placed inline with:

```yaml
data:
  redis_liveness.sh: |
    response=$(
      redis-cli \
        -h localhost \
        -p {{ .Values.redis.containerPorts.redis }} \
        ping
    )
    if [ "$response" != "PONG" ] && [ "${response:0:7}" != "LOADING" ] ; then
      echo "$response"
      exit 1
    fi
    echo "response=$response"
```

Farfaday: Hi @pdrastil 
I simply tried to stick to the same logic as in the redis-ha helm chart from DandyDeveloper.
See here: https://github.com/DandyDeveloper/charts/blob/master/charts/redis-ha/templates/_configs.tpl
If you prefer, I can move that to the ConfigMap as you proposed.

mkilchhofer: I'd also prefer to have it directly inside the ConfigMap (for simplicity). I think that the HA helm chart reuses the "script" depending on the mode you install it.

Farfaday: Hi,
I rebased my branch and moved the probes scripts directly into the ConfigMap.

Farfaday: Sorry, I had forgotten to bump the version.
Looks like someone needs to re-trigger the CI jobs :)
Thanks!

mkilchhofer (commented):
Change LGTM.

@yu-croco , @pdrastil and @jmeridth wdyt, shall we make it optionally to be:
- 100% backward compatible
- in line with upstream

?

yu-croco:
I agree with making it optional 🙋

mkilchhofer (approved):

charts/argo-cd/templates/redis/deployment.yaml
```diff
@@ -139,8 +189,12 @@ spec:
         {{- end }}
         {{- end }}
       {{- end }}
-      {{- with .Values.redis.volumes }}
       volumes:
+        - name: health
+          configMap:
+            name: {{ include "argo-cd.redis.fullname" . }}-health-configmap
+            defaultMode: 0755
```
yu-croco, line 196: This ConfigMap is created when `if and .Values.redis.enabled (not $redisHa.enabled)` , so we also need same condition to refer here ...? 👀

Farfaday: The whole deployment.yaml file is enclosed in that if , see [line 2](https://github.com/argoproj/argo-helm/pull/2400/files/8497df2d7795fe948c960b6cc0cb666138a379e8#diff-cc2af74d39b4cb33e7fe369cddbed518c6a636083bdaee5e803a142584b918d1L2)

yu-croco: Oh thank you!

yu-croco (approved):

PR #2403, opened 2023-12-26
feat(argo-events): Evaluating image values with tpl

valleedelisle:
Checklist:

### DIFF
--- a/charts/argo-events/Chart.yaml
+++ b/charts/argo-events/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v1.8.1
 description: A Helm chart for Argo Events, the event-driven workflow automation framework
 name: argo-events
-version: 2.4.1
+version: 2.4.2
 home: https://github.com/argoproj/argo-helm
 icon: https://avatars.githubusercontent.com/u/30269780?s=200&v=4
 keywords:
--- a/charts/argo-events/Chart.yaml
+++ b/charts/argo-events/Chart.yaml
@@ -19,4 +19,7 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
+      description: Evaluating image values as templates
+    - kind: changed
       description: Upgrade Argo Events to v1.8.1
+
--- a/charts/argo-events/templates/argo-events-controller/config.yaml
+++ b/charts/argo-events/templates/argo-events-controller/config.yaml
@@ -13,8 +13,8 @@ data:
         versions:
         {{- range .Values.configs.nats.versions }}
         - version: {{ .version }}
-          natsStreamingImage: {{ .natsStreamingImage }}
-          metricsExporterImage: {{ .metricsExporterImage }}
+          natsStreamingImage: {{ tpl .natsStreamingImage $ }}
+          metricsExporterImage: {{ tpl .metricsExporterImage $ }}
         {{- end }}
       {{- end }}
       {{- if .Values.configs.jetstream.versions }}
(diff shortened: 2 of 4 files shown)


yu-croco:
HI @valleedelisle , thank you for your PR.
Can you please tell us the motivation / use case for your PR because I am not sure what `tpl` solves ?

valleedelisle:
Hello @yu-croco,

One of our requirements is to maintain a downstream image registry because we don't have internet access from some of our lifecycle environments. We want to use the same values files for each of our lifecycle environment and we store the image registry host (among other things) in a values  file for each one of those environment. When we spin a helm chart, we pass that lifecycle's environment values as well as the values for this specific helm chart. Eg:

```yaml
# lifecycle/dev.yaml
global:
  imageRegistry: internal.host.dvd.dev
```

```yaml
# common-values.yaml
[...]
controller:
  # -- Argo Events controller name string
  name: controller-manager
  image:
    repository: "{{ .Values.global.imageRegistry }}/argoproj/argo-events:v1.8.1"
```

```console
$ helm template argo-events -f common-values.yaml -f lifecycle/dev.yaml
```

Will result in
```yaml
[...]
    spec:
      containers:
      - name: controller-manager
        image: internal.host.dvd.dev/argoproj/argo-events:v1.8.1:v1.8.1
        imagePullPolicy: IfNotPresent
        args:
        - controller
        env:
        - name: ARGO_EVENTS_IMAGE
          value: internal.host.dvd.dev/argoproj/argo-events:v1.8.1:v1.8.1
        - name: NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
[...]
```

This is what we do with bitnami charts and it works well.

Thanks,

DVD

pdrastil:
For controller - global option is already supported via:

```yaml
global:
  image:
    repository: internal.host.dvd.dev/argoproj/argo-events
```

and  I don't believe there is much value in repeating yourself in all places with another layer of indirection instead of just overriding global configuration with your custom value.

For JetStream / NATs image versions having some way to specify your custom registry would make sense if you mirror 1:1 the registries that controller is using for EventBuses (btw there are not hosted on quay). Even then users should pin specific versions they are using in their clusters and defaults are provided just for quick start.

```yaml
configs:
  ## NATS event bus
  nats:
    # -- Supported versions of NATS event bus
    # @default -- See [values.yaml]
    versions:
      - version: latest
        natsStreamingImage: internal.host.dvd.dev/nats-streaming:latest
        metricsExporterImage: internal.host.dvd.dev/natsio/prometheus-nats-exporter:latest
      - version: 0.22.1
        natsStreamingImage: internal.host.dvd.dev/nats-streaming:0.22.1
        metricsExporterImage: internal.host.dvd.dev/natsio/prometheus-nats-exporter:0.8.0
```

valleedelisle:
> For controller - global option is already supported via:
> 
> ```yaml
> global:
>   image:
>     repository: internal.host.dvd.dev/argoproj/argo-events
> ```
> 
> and I don't believe there is much value in repeating yourself in all places with another layer of indirection instead of just overriding global configuration with your custom value.

We only have the registry host (eg: `internal.host.dvd.dev`) defined in `global.imageRegistry` because it's actually the value that is used across all our helm charts. That means that we need to create a new global value only for the `argo-events` image. Because we reuse the same lifecycle file across all our deployed helm charts, that means that all of them will get a `global.image.repository` value that will contain the `argo-events` image repo.
 
It's the same for the nats image. We maintain the tag and the repo inside the `argo-events`'s values file and the registry host is shared across them.

Bitnami is `tpl`ing all this without any issue, and I think it's the same with a lot of other helm charts.

valleedelisle:
is there any chance to get this merged? Do we have any issue in templating userinput values? I see this all the time in all other charts.

PR #2350, opened 2023-11-16
chore(argo-workflows): Remove xip url from test to avoid confusion

tico24:
The presence of a xip.io url was confusing to some inexperienced users who thought it a legitimate URL to use to set up Argo Workflows.

To compound matters, xip.io has been offline for a while, so it's evident that the tests don't actually make use of the URL anyway.

This PR changes the url to a placeholder to reduce the confusion.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Issue #970, opened 2021-10-15
Upgrade to Argo-rollouts v2 fails


derjust:
### Describe the bug

```
$ helm upgrade --force --namespace=shared argo-rollouts .
Error: UPGRADE FAILED: failed to replace object: Deployment.apps "argo-rollouts" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"rollouts-controller", "app.kubernetes.io/instance":"argo-rollouts", "app.kubernetes.io/name":"argo-rollouts"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

### Related helm chart

argo-rollouts

### To Reproduce

1. Have an argo 1.0.4 running
2. Try to upgrade to argo 1.1.0 via argo-helm 2.0

### Expected behavior

The upgrade 'just works' and migrates to the new version

mkilchhofer:
Hi @derjust 
Did you read the upgrading notes? This is an expected behavior from our side and we put some information in the README to help users upgrade:
https://github.com/argoproj/argo-helm/tree/master/charts/argo-rollouts#to-200

> ## Upgrading
> 
> ### To 2.0.0
> 
> * The argo-rollouts dashboard is added to the template and can be enabled by setting `dashboard.enabled=true`.
> * There is a breaking change where the selector label `app.kubernetes.io/component: {{ .Values.controller.component }}` is added to rollout's deployment and service in order to distinguish between the controller and the dashboard component.
>   To upgrade an existing installation, please **add the `--force` parameter** to the `helm upgrade` command or **delete the Deployment and Service resource** before you upgrade. This is necessary because Deployment's label selector is immutable.

mkilchhofer:
Oh this looks like an interesting issue. Have to read through tomorrow or so.

When you can't tolerate a small downtime, maybe there are possibilities by deleting the Deployment and using `--cascade=false`. Have to test that.

mubarak-j:
Considering that the operation of `upgrade --force` for helm3 is still up for debate, would it make sense to hold off releases that require the usage of `--force` flag or at least for releases that changes k8s immutable resources?

mubarak-j:
@derjust I'm curious if you found a safe approach to upgrade the helm chart? without deleting deployments managed by agro-rollouts

@mkilchhofer any insights on how o workaround this issue without deleting deployments/manifests managed by argo-rollouts ?

[...]

PR #2407, opened 2024-01-01
chore(argo-cd): Remove deprecated features and redesign Ingress

pdrastil:
It's more than 1 year since these were removed / deprecated and it's probably safe to drop backward compatibility.

Checklist:


tico24 (approved):

charts/argo-cd/templates/argocd-application-controller/statefulset.yaml
```diff
@@ -13,8 +13,6 @@ metadata:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
 spec:
   replicas: {{ .Values.controller.replicas }}
-  # TODO: Remove for breaking release as history limit cannot be patched
-  revisionHistoryLimit: 5
```
mkilchhofer, line 17: If we drop this (and I'd sponsor to do so), we need a major bump and also instructions on howto upgrade.

charts/argo-cd/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: v2.9.3
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 5.52.0
+version: 5.53.0
```
mkilchhofer, line 6: Lets go for a major bump and completely fulfill the Semver spec ;) we don't need another project doing Semver with a custom deprecation policy 🙈

charts/argo-cd/templates/argocd-application-controller/statefulset.yaml
pdrastil: It's mentioned in README that you need to use replace strategy (`helm upgrade --force` or `helm template | kubectl replace -f -`)

charts/argo-cd/Chart.yaml
pdrastil: Bumped

pdrastil:
@mkilchhofer If we want to do major bump - do we want to go through arguments and name things consistently across components. `env`, `envFrom` vs `extraArgs`.

yu-croco (approved):
🚀

pdrastil:
Also another thing that might go into major release is the changes I've mentioned about multiple ingress controller implementations on Slack. Could be good opportunity to clean this up.

mbevc1 (approved):

yu-croco (approved):

charts/argo-cd/templates/argocd-application-controller/statefulset.yaml
mkilchhofer: Helm upgrade force does not work anymore with helmv3+:
- https://github.com/argoproj/argo-helm/issues/970

But ok, we can release that👍

mkilchhofer (approved):

charts/argo-cd/templates/argocd-application-controller/statefulset.yaml
pdrastil: @mkilchhofer I'm going to port also #2415 into this as change in Ingress interface will be breaking

mbevc1 (approved):

pdrastil:
Added breaking changes to ingress rewrite. As it's harder to review:
- Removed catch-all rule from all ingress resources
- Implementation is aligned with Bitnami chart (no options to supply paths or hosts by default)
   - There can be only single path for Argo CD server backend
   - There can be only one host that is configured for SSO login
   - Additional rules or hosts can be provided via `extraHosts`, `extraPaths` or `extraRules` 
   - TLS is now boolean option that expects well-known secret (argocd-server-tls) or custom TLS via `extraTls`
 - AWS & GKE Application Load Balancer options were moved into ingress and each have dedicated implementation that is enabled with `controller: aws` or `controller: gke`
   - GKE must always use port numbers to get valid health checks
   - ALB must have extra service for gRPC

charts/argo-cd/values.yaml
```diff
@@ -1932,29 +1898,47 @@ server:
...
 
-    # -- List of ingress hosts
-    ## Argo Ingress.
-    ## Hostnames must be provided if Ingress is enabled.
-    ## Secrets must be manually created in the namespace
-    hosts: []
-      # - argocd.example.com
+    # -- Argo CD server hostname
+    ## NOTE: Hostname must be provided if Ingress is enabled
+    hostname: argocd.server.local
```
mkilchhofer, line 1921: example.com

charts/argo-cd/values.yaml
```diff
@@ -1964,17 +1948,70 @@ server:
...
-      # - secretName: your-certificate-name
-      #   hosts:
-      #     - argocd.example.com
-
-    # -- Uses `server.service.servicePortHttps` instead `server.service.servicePortHttp`
-    https: false
+    # -- Additional ingress rules
+    # @default -- `[]` (See [values.yaml])
+    extraRules: []
+      # - host: example.server.local
```
mkilchhofer, line 1954: example.com

charts/argo-cd/values.yaml
```diff
@@ -1987,32 +2024,27 @@ server:
...
     pathType: Prefix
+
+    # -- Enable TLS configuration for the hostname defined at `server.ingressGrpc.hostname`
+    ## TLS certificate will be retrieved from a TLS secret with name: `<hostname>-tls`
+    tls: false
+
+    # -- The list of additional hostnames to be covered by ingress record
+    # @default -- `[]` (See [values.yaml])
+    extraHosts: []
+      # - name: grpc.argocd.server.local
```
mkilchhofer, line 2043: example.com

charts/argo-cd/values.yaml
```diff
@@ -2022,15 +2054,25 @@ server:
       #       port:
       #         name: use-annotation
 
-    # -- Ingress TLS configuration for dedicated [gRPC-ingress]
-    tls: []
+    # -- Additional ingress rules
+    # @default -- `[]` (See [values.yaml])
+    extraRules: []
+      # - host: example.server.local
```
mkilchhofer, line 2060: example.com

charts/argo-cd/values.yaml
```diff
@@ -2798,6 +2754,69 @@ applicationSet:
...
+    labels: {}
+    # -- Additional ingress annotations
+    annotations: {}
+
+    # -- Defines which ingress ApplicationSet controller will implement the resource
+    ingressClassName: ""
+
+    # -- Argo CD ApplicationSet hostname
+    ## NOTE: Hostname must be provided if Ingress is enabled
+    hostname: argocd.server.local
```
mkilchhofer, line 2772: example.com

mkilchhofer (changes requested):
Can we use the example.com domain?
-> argocd.example.com

example.com is there for documentation and examples, it is an official reserved domain by IANA:
- https://github.com/argoproj/argo-helm/pull/2352
- https://github.com/argoproj/argo-helm/pull/2350
- IANA RFC https://www.rfc-editor.org/rfc/rfc2606.html

charts/argo-cd/values.yaml
```diff
@@ -1934,29 +1900,47 @@ server:
...
 
+  # Argo CD server ingress configuration
   ingress:
     # -- Enable an ingress resource for the Argo CD server
     enabled: false
-    # -- Additional ingress annotations
-    annotations: {}
+    # -- Specific implementation for ingress controller. One of `generic`, `aws` or `gke`
+    ## Additional configuration might be required in related configuration sections
+    controller: generic
```
pdrastil, line 1909: @mkilchhofer Any opinion on this parameter name that enables specific implementation for backing ingress controller (GKE, AWS maybe something else in the future that have kinda unique setup)

mkilchhofer: Also gave some thoughts about this but didn't found a better name 😁 
names that came in to my mind:
- controller
- type
- flavour
- variant

But I'd say, lets go with `controller`. 👍

pdrastil:
@mkilchhofer Are you OK with current state? Merge is blocked since your last review.

charts/argo-cd/templates/argocd-application-controller/statefulset.yaml
pdrastil: Changed to configuration in values.yaml to allow upgrade without replacing the resource based on discussion with @mkilchhofer

charts/argo-cd/values.yaml
```diff
@@ -577,6 +577,9 @@ controller:
   # Additional replicas will cause sharding of managed clusters across number of replicas.
   replicas: 1
 
+  # -- Maximum number of controller revisions that will be maintained in controller StatefulSet history
+  revisionHistoryLimit: 5
```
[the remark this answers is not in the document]
pdrastil: Problem that I've mentioned to @mkilchhofer is if you change this to different value like null `~` you have to replace whole deployment as this field is immutable. That's why the value is original 5 for upgrade and you can unset this / or use global one.

jmeridth (approved):

mkilchhofer (approved):

jetersen:
Release note did not describe to what degree was changed nor provide an upgrade guide 😓
I read the release notes thinking you only changed ingress but even that broke because of the use of `hostname` which seems unconventional. Very common to use `hosts`
Now left to figure out what changes I had not migrated over the years.

jetersen:
Oh overriding the configs.cm: breaks this condition:

```
{{- if eq (toString (index (coalesce .Values.server.config .Values.configs.cm) "exec.enabled")) "true" }}
```

```
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = Manifest generation error (cached): `helm template . --name-template argocd --namespace argocd --kube-version 1.28 <api versions removed> --include-crds` failed exit status 1: Error: template: argocd-proxy/charts/argo-cd/templates/argocd-server/role.yaml:45:66: executing "argocd-proxy/charts/argo-cd/templates/argocd-server/role.yaml" at <.Values.configs.cm>: nil pointer evaluating interface {}.cm Use --debug flag to render out invalid YAML
```

pdrastil:
@jetersen Hi thanks for feedback. I'll add additional info to release notes. I agree that `hostname` is uncommon, however it makes sense as all components share single domain and some configuration options (SSO, notifications) allow only single URL for callbacks. There is upcoming PR https://github.com/argoproj/argo-helm/pull/2499 that simplifies configuration of all services with this cross-cutting concern from single place and user config will look more like this sample, instead of copy & pasting same value into multiple places.

```yaml
global:
  domain: argocd.example.com

server:
  certificate:
    enabled: true
  ingress:
    enabled: true
    tls: true
```

pdrastil:
@jetersen Have you seen this or you checked release notes that are part of the chart? https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd#600

jetersen:
Usually Renovate will pull the changelogs from ArtifactHub but because your pushing GitHub Releases as well Renovate, we basically do not see the changelog. So me just trusting GitHub Releases missed that you maintain the changelog in README.md
I wonder why the GitHub Releases do not contain the content from the changelog.md
Shouldn't just be a part of the github workflow for releasing? 🤔

PR #2410, opened 2024-01-02
feat(argo-cd): Add optional mapping of argocd-repo-server emptydir to custom volumes

aroundthecode:
Checklist:


aroundthecode:
Hi @pdrastil the idea was to leave default behaviour as-is with emptydir but allow people with specific use case to be able to customize volumes at will.
The big issue with emptydir is that is triggereing "disk pressure" on k8 nodes, so basically is not production ready as well.

In my setup I had to manually modify Helm chart as per PR to let it work otherwise it woud crash my nodes even with few application. 
Using disk mount on tmp folder is the official solution for this so it's quite strange it's not supported at all by helm chart.

snuggie12:
> Hi, thanks for PR but we never merged PVC support as this would be breaking for production deployments that use HPA to run more replicas (typically 30-50 in large scale deployments). For this reason the repo-server was intentionally kept as a stateless. See details in #1648

@pdrastil I agree with @aroundthecode and it would be nice to override this. Currently I have to patch in using ephemeral volume claims which works around the issue you mentioned.
```yaml
$ cat patches/argocd-repo-server-deploy.patch.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: argocd-repo-server
  namespace: argocd
spec:
  template:
    spec:
      volumes:
        - name: tmp
          emptyDir: null
          ephemeral:
            volumeClaimTemplate:
              metadata:
                labels:
                  type: argocd-repo-server-tmp
              spec:
                accessModes:
                  - ReadWriteOnce
                storageClassName: standard-rwo
                resources:
                  requests:
                    storage: 10Gi
```

aroundthecode:
@pdrastil and others, I'd like to kown if this PR can be merged or not.
I'm strugglig a bit to get chart-test passing (sotty this is my first PR on this project), but in the last day nobody let it run and I'm keeping the branch updated for other amin branch modification.

I think this PR allows to keep things "stateless" by default as @pdrastil suggested but also allow people to customize mountpoints followin best practices if needed.

In my company we are currently using a manually modified version of argo-helm project, so having this PR merged would mean we can go on using it and contributing to it, but I'd like to know if there is any interest on maintainer side to this kind of approach.

aroundthecode:
@mbevc1 @mkilchhofer @yu-croco @jmeridth @pdrastil @tico24 any chance to see this approved?
Turned full green yesteday but got again surpassed by other version and had to merge again.
I'm available for any information/discussion.

jmeridth:
As we've had two consumers of this project express the need and one has done the PR to allow the change I'm a fan of merging. This leaves default of emptydir in place. 

@yu-croco @pdrastil @mkilchhofer what say you?

jmeridth (approved):

charts/argo-cd/values.yaml
```diff
@@ -2217,6 +2217,24 @@ repoServer:
   #  - name: cmp-tmp
   #    emptyDir: {}
 
+  # -- Volumes to be used in replacement of emptydir on default volumes
+  existingVolumes: {}
+  #  gpg_keyring:
```
pdrastil, line 2222: @jmeridth I'm generally OK with having this as opt-in. Not sure if we want to have mixed syntax here. We use `camelCase` everywhere but volume suddenly changes to `lower_underscore`.

mbevc1: Looks good and a good shout about naming.

aroundthecode: @pdrastil @mbevc1  volumes name are lower_underscore to keep the very same name as volumeMount names into chat but no problem to change them camelCase if you wish them to match global naming. 
Please just confirm if you want me to change this.

mbevc1: I don't have a strong preference and we should aim for consistency.

aroundthecode: moved to camelCase

pdrastil (approved):

PR #2415, opened 2024-01-06
chore(argo-cd): Remove wildcard catch all ingress rule

pdrastil:
There will be multiple PRs for Ingress to decouple implementations for various controllers in attempt to make this piece more maintainable.

This PR removes ingress rule without specified hostname that acts as a global catch-all rule. It's usually not a good practice to have this in a cluster in first place as it can intercept traffic from another misconfigured application that is behind same Load balancer / IP. If we want to support this we can consider `defaultBackend` section if no hosts are provided, however from security perspective it's better to get 404 or 503 if administrator haven't configured default backend for any non-matching traffic.

Ref:
- https://kubernetes.io/docs/concepts/services-networking/ingress/#default-backend
- https://kubernetes.io/docs/concepts/services-networking/ingress/#single-service-ingress

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: removed
       description: Wildcard catch all ingress rule
+    - kind: added
+      description: Ingress extra rules to allow explicit configuration of catch all rule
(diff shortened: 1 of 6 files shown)


jmeridth:
@pdrastil what do you think about making this a minor version update instead of patch as some people could be, maybe even unwittingly, be using the catch-all situation.  wdyt?

I approve the security best practice approach and getting it in.

pdrastil:
@jmeridth Bumped. Also I would like to have your opinion on guard on top for of ingress. If no host is specified do we want to fail hard or do we want to silently skip rendering and let the user fix it troubleshoot why the ingress is not working.

pdrastil:
I've added extraRules so it's still possibly to explicitly configure the catch-all rule if someone needs it.

pdrastil:
Resolved via https://github.com/argoproj/argo-helm/pull/2407

PR #2451, opened 2024-01-24
fix(argo-workflows): Fix trailing whitespace in controller ConfigMap

SlaterByte:
This commit addresses two places where trailing whitespace may be generated in the workflow-controller ConfigMap.

When the value `.resourceRateLimit` is not null, a trailing whitespace is added after "resourceRateLimit:". This commit trims that whitespace.

When the value `.sso.redirectUrl` is left as the default of empty string, a trailing whitespace is left after "redirectUrl:" as this value is not quoted. This commit pipes this value through `quote` to ensure this whitespace is no longer trailing and instead followed by `"` pair.

The main motivation of this commit is to generate a workflow-controller ConfigMap that is properly pretty-printed, which trailing whitespace prevents.

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.4
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.40.7
+version: 0.40.8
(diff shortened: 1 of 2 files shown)


vladlosev (approved):
LGTM

yu-croco (approved):

PR #2485, opened 2024-02-09
fix(argo-cd): add hostname of extraHost to tls hosts

sjoukedv:
Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.10.0
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 6.0.3
+version: 6.0.4
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: fixed
-      description: Fixed rendering of network policy when applicationset webhook is enabled
+      description: Also added extraHosts defined to the TLS hostnames
--- a/charts/argo-cd/templates/argocd-server/ingress.yaml
+++ b/charts/argo-cd/templates/argocd-server/ingress.yaml
@@ -55,6 +55,11 @@ spec:
     {{- if .Values.server.ingress.tls }}
     - hosts:
       - {{ .Values.server.ingress.hostname }}
+      {{- range .Values.server.ingress.extraHosts }}
+        {{- if .name }}
+      -  {{ .name }} 
+        {{- end }}
+      {{- end }}
       secretName: argocd-server-tls
     {{- end }}
     {{- with .Values.server.ingress.extraTls }}


sjoukedv (commented):
We should add the hosts defined in extraHosts to the list of hostname in the TLS hosts by default

yu-croco:
Hi @sjoukedv . Thank you for your PR.

> I have updated the chart [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog) with all the changes that come with this pull request according to changelog.

Please handle this, remove old one and add new one.

mkilchhofer:
@pdrastil, are you okay with this proposal?

sjoukedv:
> Hi @sjoukedv . Thank you for your PR.
> 
> > I have updated the chart [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog) with all the changes that come with this pull request according to changelog.
> 
> Please handle this, remove old one and add new one.

My bad, fixed now.

I just needed a quick change for the hosts as it didn't make sense for me to define another hostname but not be able to put it in the list of hostname for `tls`.

pdrastil:
@sjoukedv @mkilchhofer It could work - if I consider following scenario:

```yaml
# Ago CD server is terminating TLS
tls: true

# argocd-server-tls secret must contain both CA bundles combined or
# these are just CNAMEs that user can provides with certificate.additionalHosts
extraHosts:
  - name: argocd.public-domain.io
  - name: arogcd.internal-domain.net
```

@sjoukedv Is this your use case? I'm wondering how do you login to UI if running on multiple domains just with local account? I didn't expect that multi-domain setup will be so common.

pdrastil:
Another angle to consider how this will behave on other controllers:
- AWS + GKE implementation will off-load TLS handling via managed certificate with `tls: false`
- What about webhook for ApplicationSet? That one will have same use case as webhook usually shares same domain as UI server with different path.

sjoukedv:
> Another angle to consider how this will behave on other controllers:
> 
> * AWS + GKE implementation will off-load TLS handling via managed certificate with `tls: false`
> * What about webhook for ApplicationSet? That one will have same use case as webhook usually shares same domain as UI server with different path.

@pdrastil I was not aware that the multiple domain setup support is not there yet (I also don't use the alias, but it would be nice to have). I would expect the `argocd-server` HTTP server to return (permanent) redirects to the `server.ingress.hostname` which seems trivial to implement (using the alias instead of redirecting is a different story)?

My use case is indeed an ingress controller (behind a L4 load balancer) where often a tool like cert-manager is used to issue the certificate which respects the list of hosts specified at `.spec.tls[].hosts`. Currently, I could only override the entire block with `tls: false` and `server.ingress.extraTls`. My point is that; I would like to specify additional aliases and it only makes sense to also include these in the TLS hosts block otherwise clients will not be able to validate the certificate as the alias is not in one of the SANs.

We can incrementely support this behaviour for various 'controllers' or put together a single merge request with all the changes.

pdrastil:
@sjoukedv You can run Argo CD with multiple domains, however I know that this doesn't work if you plan to use SSO with Dex or existing OIDC provider as `configs.cm.url` must be set to single domain for authentication redirect and users must always end on this domain otherwise they were not able to login. Running server with TLS termination on multiple domains should be fine on it's own.

pdrastil:
@mkilchhofer I kinda feel we should create similar section in docs for Ingress as we have for HA and showcase how different setups works.

@sjoukedv Can you share your setup with sample with multiple domains? I'm mainly interested in what controller is used, where is the TLS termination, how certs  are provisioned in relation to values.yaml. Thank you

sjoukedv:
> @mkilchhofer I kinda feel we should create similar section in docs for Ingress as we have for HA and showcase how different setups works.
> 
> @sjoukedv Can you share your setup with sample with multiple domains? I'm mainly interested in what controller is used, where is the TLS termination, how certs are provisioned in relation to values.yaml. Thank you

@pdrastil 

In short;

- Load balancer L4
- TLS termination by ingress controller
- Certificates issued by certificate (with e.g. LetsEncrypt or Cloudflare) stored in Kubernetes Secret

You're probably right about the SSO, but I expected when I visit `argocd.alias-domain.nl` (not logged in / or logged in) to be redirect to `argocd.main-domain.nl` where I can complete the SSO on the main domain which is configured for dex.

Redacted `values.yaml`:

```yaml
server:
  ingress:
    tls: true
    enabled: true
    hostname: argocd.example-domain.nl
    extraHosts:
      - name: argocd.main-domain.nl
        path: /

    ingressClassName: nginx
    annotations:
      cert-manager.io/cluster-issuer: "<my-issuer>"
      nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"


configs:
  cm:
    url: https://argocd.main-domain.nl

    dex.config: |
      connectors:
        - type: gitlab
          id: gitlab
          name: GitLab
          config:
            baseURL: https://gitlab.com
            clientID: $GITLAB_APPLICATION_ID
            clientSecret: $GITLAB_CLIENT_SECRET
            redirectURI: http://127.0.0.1:5556/dex/callback
            groups:
            - <my-group>
            useLoginAsID: false
```

pdrastil (approved):

mbevc1 (approved):

pdrastil:
@sjoukedv Thanks for sample. I'll try this as I'm really curious if this works or not. I had a use case when Argo CD was behind GSLB with fail-over and I was running Traefik with Dex + LDAP.

PR #2489, opened 2024-02-09
feat(argo-cd): Add scrapeTimeout property to server and application…

nikolay-o:
I've added scrapeTimeout property to server and application, applicationSet controllers.

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.10.0
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 6.0.6
+version: 6.0.7
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,7 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: fixed
-      description: Dedicated gRPC ingress should route to HTTPS port even when running with insecure server
     - kind: changed
-      description: Dedicated gRPC ingress TLS secret changed to argocd-server-grpc-tls based on feedback
+      description: Added scrapeTimeout property for server and application,applicationSet controllers.
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -727,6 +727,7 @@ NAME: my-release
 | controller.metrics.rules.namespace | string | `""` | PrometheusRule namespace |
 | controller.metrics.rules.selector | object | `{}` | PrometheusRule selector |
 | controller.metrics.rules.spec | list | `[]` | PrometheusRule.Spec for the application controller |
+| controller.metrics.scrapeTimeout | string | `""` | Prometheus ServiceMonitor scrapeTimeout. If empty, Prometheus uses the global scrape timeout unless it is less than the target's scrape interval value in which the latter is used. |
 | controller.metrics.service.annotations | object | `{}` | Metrics service annotations |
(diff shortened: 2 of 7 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: Improved documentation for various ingress setups
```
yu-croco, line 30: Please remove old one.

nikolay-o: Fixed, thanks for review

charts/argo-cd/templates/argocd-application-controller/servicemonitor.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       {{- with .Values.controller.metrics.serviceMonitor.interval }}
       interval: {{ . }}
       {{- end }}
+      scrapeTimeout: {{ .Values.controller.metrics.serviceMonitor.scrapeTimeout }}
```
yu-croco, line 25: maybe it's better to align to other values, like `interval` ? 🤔  WDYT?
```yaml
      {{- with .Values.controller.metrics.serviceMonitor.interval }}
      interval: {{ . }}
      {{- end }}
```

yu-croco: Ah but default value is defined. hmm it doesn't make much difference.

yu-croco (approved):
LGTM

charts/argo-cd/values.yaml
```diff
@@ -758,6 +758,7 @@ controller:
   metrics:
     # -- Deploy metrics service
     enabled: false
+    scrapeTimeout: 10s
```
mkilchhofer, line 761: Missing docs

charts/argo-cd/README.md
```diff
@@ -727,6 +727,7 @@ NAME: my-release
 | controller.metrics.rules.namespace | string | `""` | PrometheusRule namespace |
 | controller.metrics.rules.selector | object | `{}` | PrometheusRule selector |
 | controller.metrics.rules.spec | list | `[]` | PrometheusRule.Spec for the application controller |
+| controller.metrics.scrapeTimeout | string | `"10s"` |  |
```
mkilchhofer, line 730: Missing docs

mkilchhofer (changes requested):

charts/argo-cd/values.yaml
nikolay-o: fixed

charts/argo-cd/README.md
nikolay-o: fixed

charts/argo-cd/templates/argocd-application-controller/servicemonitor.yaml
mkilchhofer: I would also preffer this method and unsetting the default in values.yaml

Prometheus-operator has no default value on ServiceMonitors, the default is inherited from the global config of prom-operator.
Ref: https://github.com/prometheus-community/helm-charts/blob/94330ab7439a5ab2701bbe8605097a904223d62c/charts/kube-prometheus-stack/charts/crds/crds/crd-servicemonitors.yaml#L470-L473

So I suggest to define it like this in values:
```yaml
      # -- (string) Prometheus ServiceMonitor scrapeTimeout. If empty, Prometheus uses the global scrape timeout unless it is less than the target's scrape interval value in which the latter is used.
       scrapeTimeout: ~
```

nikolay-o:
@mkilchhofer I fixed documentation, could you please look?

charts/argo-cd/templates/argocd-application-controller/servicemonitor.yaml
nikolay-o: @mkilchhofer that's make sense, replaced

mkilchhofer (approved):

mkilchhofer:
> @mkilchhofer I fixed documentation, could you please look?

Perfect, thank you so much for your contribution.
Lgtm👍

mbevc1 (approved):

PR #2511, opened 2024-02-14
fix(argo-cd): Fix secret name for applicationset webhook ingress

pdrastil:
Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.10.0
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 6.0.12
+version: 6.0.13
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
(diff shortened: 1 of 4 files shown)


mbevc1 (approved):

yu-croco (approved):

volker-raschek:
Hi @pdrastil, @mbevc1 and @yu-croco, 
this pull request means that a custom secret can no longer be used.

In version 7.5.2 of the helm chart, I am happy that the cert-manager is now primarily used to create certificates, but you have not taken one thing into account. 

There are users like me who generate wildcard certificates from let's encrypt. The certificate, stored in a secret, is synchronized via [reflector](https://github.com/emberstack/kubernetes-reflector) in all necessary namespaces. There it can then be referenced as a secret in the ingress resource. 

Because you have now defined this [hard coded](https://github.com/pdrastil/argo-helm/blob/64b1857a563fdf292a08f1457de7f4f3852f57a0/charts/argo-cd/values.yaml#L2793), the workflow or setup is no longer possible at all and that's very bad.

`secretName: argocd-applicationset-controller-tls`

Questions and answers that may arise.

1. Why must a wildcard certificate be used?
Because otherwise the Let's Encrypt API will reveal public DNS names: https://crt.sh/

2. Why can't you create a second wildcard certificate with the same CommonName / SAN?
Because the client (API's, browsers and so on) would always make a new TLS handshake with HSTS. These TLS handshakes are quite expensive. In my opinion to expensive!

Please provide a fix, so that it's possible to use a custom secret of all ingress resources.

Issue #2527, opened 2024-02-19
re-introduce support for path based Ingress with empty hostname


paolofacchinetti:
### Is your feature request related to a problem?

Since version `6.0.0`, it is no longer possible to leave out the `server.ingress.hosts` field (now renamed to `server.ingress.hostname`) when configuring the ArgoCD ingress resource.

I deploy ArgoCD to a single-node k3s cluster with Traefik as an ingress controller. Leaving out the `hosts` field while specifying a custom argoCD path in both `configs.params.server.rootpath` and `server.ingress.paths` (now renamed to `server.ingress.path`) meant that I could reach my ArgoCD deployment on `<CLUSTER_IP>/argocd`

I've looked into the docs and tried out various configs combination, but by looking at the actual chart source code it doesn't look like my usecase is supported anymore.

Right now, leaving out the `hostname` field means that the default hostname `argocd.example.com` is used, making the ingress useless.

### Related helm chart

argo-cd

### Describe the solution you'd like

Being able to do path based routing only for ingress, while leaving out the `hosts` field

### Additional context

My values.yaml that worked with ArgoCD <  6.0.0

```yaml
configs:
  params:
    server.rootpath: "/argocd"
    server.insecure: true
server:
  ingress:
    enabled: true
    https: false
    paths:
      - "/argocd"
    annotations:
      # annotations for self-signed HTTPS ingress
      traefik.ingress.kubernetes.io/router.entrypoints: websecure
      traefik.ingress.kubernetes.io/router.tls: "true"
```

pdrastil:
Hi @paolofacchinetti your use case should be supported like this with catch-all rule. Removing this was intentional as using ingress without hostname is not really good security practice because all unmatched traffic gets routed to Argo CD server.

```yaml
configs:
  params:
    server.rootpath: "/argocd"
    server.insecure: true
server:
  ingress:
    enabled: true
    annotations:
      # annotations for self-signed HTTPS ingress
      traefik.ingress.kubernetes.io/router.entrypoints: websecure
      traefik.ingress.kubernetes.io/router.tls: "true"
    # Add extra rule for catch-all traffic
    extraRules:
    - http:
        path: /argocd
        backend:
          service:
            name: argocd-server
            port:
              name: http
```

pdrastil:
One trick you might like - for using IPs with fake domain for local setup check https://nip.io/

paolofacchinetti:
Hi @pdrastil, thank you for the very quick feedback!

[...]

PR #2530, opened 2024-02-20
feat(argo-cd): Add templating of ingress extra paths and rules

pdrastil:
Resolves: https://github.com/argoproj/argo-helm/issues/2527

Added `tpl` where user might need to reference backend service name.

Checklist:


mbevc1 (approved):

charts/argo-cd/values.yaml
```diff
@@ -2125,15 +2129,17 @@ server:
...
-      #   http:
-      #       path: /
+      # - http:
+      #     paths:
+      #     - path: /
+      #       pathType: Prefix
       #       backend:
       #         service:
-      #           name: example-svc
+      #           name: '{{ include "argo-cd.server.fullname" . }}'
```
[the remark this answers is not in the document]
pdrastil: It works with Helm 3.14.1. You can try it.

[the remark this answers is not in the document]
mkilchhofer: Alright it works 👍 tested it with helm v3.13.2.

charts/argo-cd/values.yaml
```diff
@@ -2028,15 +2029,17 @@ server:
 
     # -- Additional ingress rules
     # @default -- `[]` (See [values.yaml])
+    ## Note: Supports use of custom Helm templates
```
mkilchhofer, line 2032: Can we move this before the @default line so it is present inside the README?
Maybe we can also use the syntax of Bitnami charts
`(evaluated as a template)`?

mkilchhofer (commented):
LGTM but maybe we should add it to the README inside the parameter table.

charts/argo-cd/values.yaml
pdrastil: @mkilchhofer We have this note in other places such as `extraContainers`. Do you want to change that as well?

yu-croco (approved):

charts/argo-cd/values.yaml
mkilchhofer: Ah, I forgot that 😄 
Lets continue then 👍

PR #2535, opened 2024-02-20
feat(argo-cd): Format redis health check confimap mode in decimal

LucasBoisserie:
To avoid issue with octal value yaml import format redis health check confimap mode in decimal.

## Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.10.1
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 6.2.3
+version: 6.2.4
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Updated Redis image tag to 7.2.4
+    - kind: fixed
+      description: Format redis health check confimap mode in decimal.
--- a/charts/argo-cd/templates/redis/deployment.yaml
+++ b/charts/argo-cd/templates/redis/deployment.yaml
@@ -193,7 +193,7 @@ spec:
         - name: health
           configMap:
             name: {{ include "argo-cd.redis.fullname" . }}-health-configmap
-            defaultMode: 0755
+            defaultMode: 493
       {{- with .Values.redis.volumes }}
         {{- toYaml . | nindent 8}}
       {{- end }}


charts/argo-cd/values.yaml
```diff
@@ -1246,6 +1246,12 @@ redis:
   # - secretRef:
   #     name: secret-name
 
+  # Health Confimap configuration
+  healthConfigmap:
+    volume:
+      # -- Volume permission of the health-configmap for redis deployment
+      defaultMode: 493
```
yu-croco, line 1253: > Add new variable to define volume permission for redis Server health check configmap. (I have to change octal value to decimal value because helm translate it by default)

It's reasonable to make it configurable but I prefer using `0755` as default. 👀  WDYT?

LucasBoisserie: I prefer too `0755` but on my first test with `0755` as default value and I noticed in the manifest that the value has been translated automatically by helm to `493`. To avoid missunderstanding, I change the default value.
I noticed the same behavior with helm-docs and to avoid driffting in readme generation, I decided to to keep the new value.

pdrastil: @yu-croco I have different opinion. 

1) This is implementation detail not a configuration option. The users should not poke and reverse engineer contents of ConfigMap to see what this thing impacts as they cannot change the embedded shell script anyway.
2) The 493 is `r--rwxrwx` so I guess this is attempt to harden the shell script. However this is stored in CM so anyone can usually change this to something dangerous on K8s level as being able to edit CMs is quite usual for most of the users.
3) The nesting of 3 dicts for single scalar value is also nice. These what-if structures that might be extended someday in future with single value were always hard to cleanup. (Last one `webhook.ingress` -> `ingress` went in last breaking change)

LucasBoisserie: @pdrastil
I add the new variable because I have an issue with octal value. My first goal was to have the capability to set it as decimal at helm runtime.
In my case, I use helm to generate manifest and apply it with terraform to have atomic change detection. When terraform decode the yaml, it translate the octal value `0755` in number `755` and fail to apply it because it not a correct value

yu-croco: @LucasBoisserie @pdrastil 
Thank you for your explanation ! I understand it.

pdrastil: @LucasBoisserie Gotcha. This is more about Terraform YAML parsing library that likes to convert things into it's own format without asking (for example: `on` -> `true`) which causes drift for YAML resources in general. For this reason I would just tweak original value to one converted / used by TF without any config option in values.yaml which should solve your problem. 

You will hit this on lots of places and imho I would consider using TF only to bootstrap Argo CD, then ignore all changes in lifecycle and switched to GitOps to avoid dealing with TF YAML drift.

LucasBoisserie: @pdrastil
It's on Terraform level, I find a issue on it with the explanation from terraform team: https://github.com/hashicorp/terraform/issues/22143#issuecomment-513372890

Ok, i update it !

That's exactly what I'm doing. Terraform bootstrap Argo CD and Argo CD deploy all components.

pdrastil: @LucasBoisserie This is off-topic from PR. This is a way I'm completely avoiding the drift or upgrades of YAML resources in TF. I just install argocd once and ignore the helm release changes completely for any future TF runs. Then use Application (app-of-apps) or ApplicationSet where Argo starts to self-manage itself with full-blown config and independent upgrade path. Here is small snippet:

```hcl
resource "helm_relase" "argocd" {
 
 # Complex values are intentionally omitted here
 # values = []

  # Ignore all future changes 
  lifecycle = {
    ignore_changes = all
  }
}

resource "kubectl_manifest" "bootstrap" {
  yaml = <<-EOF
  kind: ApplicationSet
  metadata:
    name: bootstrap
    namespace: argocd
  spec:
    # Your GitOps spec with specific path / branch
    # that will also hydrate Argo with your config values.yaml
  EOF

  depends_on = [ helm_release.argocd ]
}
```

pdrastil (approved):

LucasBoisserie:
@yu-croco , is it good for you ?

yu-croco (approved):
Thank you for your contribution!

PR #2536, opened 2024-02-21
feat(argo-rollouts): Restrict write access to rollout status in the aggregated roles

nedal87:
The status subresource is typically managed automatically by the controller managing the resource, in this case, the Argo Rollouts controller. Allowing users to directly manipulate the status subresource could lead to inconsistencies and unintended behavior, as the controller might overwrite or ignore manually set status updates.

Therefore, it's generally a good practice to restrict permissions to modify the status subresource to the controller or system components responsible for managing the resource, rather than granting such permissions to users directly. This helps ensure the integrity and consistency of the resource's state.
After this change, users will be still able to view the rollout status. 

Current setup: argo-rollouts:edit

Checklist:

### DIFF
--- a/charts/argo-rollouts/templates/aggregate-roles.yaml
+++ b/charts/argo-rollouts/templates/aggregate-roles.yaml
@@ -37,7 +37,6 @@ rules:
   resources:
   - rollouts
   - rollouts/scale
-  - rollouts/status
   - experiments
   - analysistemplates
   - clusteranalysistemplates
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: v1.6.4
+appVersion: v1.6.6
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 2.34.0
+version: 2.35.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-rollouts/assets/logo.png
 keywords:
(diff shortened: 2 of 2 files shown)


charts/argo-rollouts/templates/aggregate-roles.yaml
```diff
@@ -37,7 +37,6 @@ rules:
   resources:
   - rollouts
   - rollouts/scale
-  - rollouts/status
```
yu-croco, line 40: Thank you for your PR.
Since upstream has the permission in latest version, so can you please fix upstream at first ?
*argo-helm follows upstream's manifest.
https://github.com/argoproj/argo-rollouts/blob/v1.6.6/manifests/base/argo-rollouts-aggregate-roles.yaml

nedal87: Hi @yu-croco sure thing, I will open a PR in upstream. Thanks!

jmeridth:
@nedal87 Thank you for the proposal. Closing this one out: the agreed path in the review was to change the upstream manifest first (argoproj/argo-rollouts `manifests/base/argo-rollouts-aggregate-roles.yaml`), and no upstream PR was opened. As of argo-rollouts v1.9.1 the upstream aggregate roles still include `rollouts/status`, and this chart intentionally mirrors upstream, so there is nothing for this PR to track. If you or anyone else wants to pursue this, please start with a PR against argoproj/argo-rollouts and we will gladly mirror it here once released. Also worth noting for any future attempt: the change would need to cover both the edit and admin aggregate roles to be consistent.

PR #2542, opened 2024-02-23
feat(argo-cd): Allow modification of path and port for dex liveness and readiness probes

fboula:
Allow to change the default http path and port for Dex server.

Fixes: https://github.com/argoproj/argo-cd/issues/9091

Checklist:

### DIFF
--- a/charts/argo-cd/templates/dex/deployment.yaml
+++ b/charts/argo-cd/templates/dex/deployment.yaml
@@ -99,8 +99,8 @@ spec:
         {{- if .Values.dex.livenessProbe.enabled }}
         livenessProbe:
           httpGet:
-            path: /healthz/live
-            port: metrics
+            path: {{ .Values.dex.livenessProbe.httpPort }}
+            port: {{ .Values.dex.livenessProbe.httpPath }}
           initialDelaySeconds: {{ .Values.dex.livenessProbe.initialDelaySeconds }}
           periodSeconds: {{ .Values.dex.livenessProbe.periodSeconds }}
           timeoutSeconds: {{ .Values.dex.livenessProbe.timeoutSeconds }}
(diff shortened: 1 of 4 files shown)


charts/argo-cd/templates/dex/deployment.yaml
```diff
@@ -99,8 +99,8 @@ spec:
         {{- if .Values.dex.livenessProbe.enabled }}
         livenessProbe:
           httpGet:
-            path: /healthz/live
-            port: metrics
+            path: {{ .Values.dex.livenessProbe.httpPort }}
+            port: {{ .Values.dex.livenessProbe.httpPath }}
```
pdrastil, line 103: The port name is hardcoded in deployment. This cannot be configured.

pdrastil: Same applies most likely to path. Configuration is automatically generated by init container here.
Ref: https://github.com/argoproj/argo-cd/blob/4dc91dcb9df77071d0d8a83511449e2297d984b1/util/dex/config.go#L44C10-L44C19

fboula: I understand your concern. This is not an ideal solution but for users relying on Google Cloud OIDC, the solution proposed in https://github.com/argoproj/argo-cd/issues/9091 seems to be a viable way to ensure dex server is restarted when it is not able to open the google connector. 

The default values are not introducing any behaviour changes so only chart users specifically changing the proposed values would see a difference.

The change would help, I understand some users may compromise their deployment if they configure a wrong value here, but if I can provide some documentation of the possible values or with a helm template function to ensure the set port is in the list of existing ports

WDYT ?

pdrastil: Hi, thanks for providing context about this change. I think it's ok to allow configuration for this edge case.

pdrastil (approved):

fboula:
I had to rebase my branch, let me know if the semver is ok.

mkilchhofer (approved):

pdrastil (approved):

Issue #2552, opened 2024-02-28
[argo-cd] `aws.backendProtocolVersion` should be `GRPC`


mikutas:
### Describe the bug

Target health status of target group corresponding to [this service](https://github.com/argoproj/argo-helm/blob/2f82fb5992fe1e390d1ebdbc4be6d5d6c6549a37/charts/argo-cd/templates/argocd-server/aws/service.yaml#L1) becomes unhealthy because the value `server.ingress.aws.backendProtocolVersion` is `HTTP2` (i.e. the target group's protocol version is `HTTP2`).
It causes argocd-server pod's "not ready" status in environment which benefit from [pod readiness gate](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.7/deploy/pod_readiness_gate/).

Target health status become healthy after setting `server.ingress.aws.backendProtocolVersion` to `GRPC`.

![image](https://github.com/argoproj/argo-helm/assets/23391543/bdd853f8-1035-4f5b-9d59-33efa442043b)

![image](https://github.com/argoproj/argo-helm/assets/23391543/92b53e0c-230b-448c-aa39-b6b64e387de0)

AWS docs:
- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html#health-check-settings
  - <img width="921" alt="image" src="https://github.com/argoproj/argo-helm/assets/23391543/1f2408db-3bf5-446d-81f3-e9271a5c2c53">
- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-protocol-version

### Related helm chart

argo-cd

### Helm chart version

6.4.1

### To Reproduce

- Use [aws-load-balancer-controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller/) to expose argocd-server
- Set following values to argo-cd chart:
  - `server.ingress.enabled`: true
  - `server.ingress.controller`: aws

### Expected behavior

Health check for target group is successful

### Additional context

My guess is `alb.ingress.kubernetes.io/backend-protocol-version` need not be configurable?

related issue: https://github.com/argoproj/argo-cd/issues/8067

yu-croco:
It's just my guess, but the default value is from sample in upstream, which means there's no deep meaning behind it
*I forgot where the doc was, but the values were the same as argo-helm's one.
[...]

PR #2553, opened 2024-02-28
fix(argo-cd): Pass argocd-server's ALB health check

yu-croco:
Resolves https://github.com/argoproj/argo-helm/issues/2552 

Checklist:


charts/argo-cd/values.yaml
```diff
@@ -2071,13 +2071,14 @@ server:
...
+      ## This tells AWS to send traffic from the ALB using gRPC. Can use HTTP2 as well
+      ## For more information: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html#health-check-settings
+      backendProtocolVersion: GRPC
       # -- Service type for the AWS ALB gRPC service
       ## Can be of type NodePort or ClusterIP depending on which mode you are running.
       ## Instance mode needs type NodePort, IP mode needs type ClusterIP
-      ## Ref: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/how-it-works/#ingress-traffic
-      serviceType: NodePort
+      ## Ref: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.7/how-it-works/#ingress-traffic
+      serviceType: ClusterIP
```
yu-croco, line 2081: [ALB Guide](https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd#aws-application-load-balancer) is also mention this, so I think it's reasonable to change default value, in addition to https://github.com/argoproj/argo-helm/issues/2552 . 🙋 
image

tico24 (approved):
lgtm

charts/argo-cd/values.yaml
pdrastil: @yu-croco I would keep `NodePort` as is. The instance mode is default for AWS load balancer controller. The `ClusterIP` is needed only for `alb.ingress.kubernetes.io/target-type: ip`

yu-croco: Thank you for your review.
Then what about adding comments for `IP mode` ? 🙋

pdrastil: @yu-croco it's already there :)

```yaml
# -- Service type for the AWS ALB gRPC service
## Can be of type NodePort or ClusterIP depending on which mode you are running.
## Instance mode needs type NodePort, IP mode needs type ClusterIP
```

pdrastil: Yes. `NodePort` means that LBs are forwarding traffic to all nodes and then it's routed to Pod via ingress controller. `ClusterIP` for IP mode means that LBs are forwarding traffic to AWS Elastic IP of Pod directly

yu-croco: > it's already there :)

I see. 💡 

>The mode is just for serviceType- both of them can use backendProtocol HTTP2 or GRPC

In the original issue, https://github.com/argoproj/argo-helm/issues/2552, `HTTP2` causes problem for IP mode. 👀

pdrastil: We can change this to `GRPC` as this service is technically gRPC running on port 443 and it's just protocol definition.

pdrastil: The `serviceType` is usually important if you run with different CNI add-on such as Calico or Cilium. In overlay mode with masquerading you are restricted to instance mode.

yu-croco: >>The mode is just for serviceType- both of them can use backendProtocol HTTP2 or GRPC
>
> In the original issue, https://github.com/argoproj/argo-helm/issues/2552, HTTP2 causes problem for IP mode. 👀

My bad, wrong quote. 😢 
It was about `backendProtocolVersion`

>Target health status of target group corresponding to this service becomes unhealthy because the value server.ingress.aws.backendProtocolVersion is HTTP2.
>It causes argocd-server pod's "not ready" status in environment which benefit from [pod readiness gate](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.7/deploy/pod_readiness_gate/).
>
>Target health status become healthy after setting server.ingress.aws.backendProtocolVersion to GRPC.

yu-croco: Hi @mikutas , if you have comments, please follow up.
*I'm sorry for the mess...

mikutas: @pdrastil As [reported](https://github.com/argoproj/argo-helm/issues/2552), if protocol version of target group (for gRPC Service) is HTTP2, the health check setting of the TG become invalid and target become unhealthy.
So aws.backendProtocolVersion is not selectable indeed.
I think this is the point of this issue.

On the other hand, `serviceType` is certainly selectable corresponding to AWS LBC's mode(instance / ip), and it's irrelevant to the issue of unhealthy target.
Therefore, I think the default value of `serviceType` need not to be changed in this PR and keeping `NodePort` as default is reasonable as well as AWS LBC's default mode is `instance`.

mkilchhofer: From my experience with ALB (and AWS EKS) serviceType is for sure relevant.
If you use `ClusterIP` and ALB target-type `ip` the health checks are handled different than in `NodePort` together with `instance`.

- `instance` by default checks traffic path until nodes
- `ip` always tries to check directly to the pod and uses `/` and "traffic port" as the default health-checking parameters

pdrastil: @mikutas Thanks for response. @yu-croco Changing the port to GRPC without any changes to service type should be enough.

mikutas: @mkilchhofer Thank you, my understanding seems incorrect.
More precicely speaking:
- `backendProtocolVersion` = `HTTP2` and `serviceType` = `ClusterIP`  (i.e. LBC is `ip` mode) -> unhealthy (my case)
- `backendProtocolVersion` = `HTTP2` and `serviceType` = `NodePort` (i.e. LBC is `instance` mode) -> healthy (your case)
- `backendProtocolVersion` = `GRPC` and `serviceType` = `ClusterIP` (i.e. LBC is `ip` mode) -> healthy (my case)
- `backendProtocolVersion` = `GRPC` and `serviceType` = `NodePort` (i.e. LBC is `instance` mode) -> healthy (probably?)

yu-croco: >Changing the port to GRPC without any changes to service type should be enough.

@pdrastil @mikutas 
I'm sorry for my mess... I reverted unnecessary diff in https://github.com/argoproj/argo-helm/pull/2553/commits/f91e531fdc6c0322a189ae51632a21979024b611 . 🙋

pdrastil (approved):

PR #2577, opened 2024-03-07
feat(argo-cd): Support setting ArgoCD Notifications secret

### LINKED ISSUE #2575: Support existing secret for ArgoCD Notification

I use a K8s operator to manage my K8s secrets  (namely [external-secrets](https://external-secrets.io/latest/)), and I would like it to create `argocd-notifications-secret` along with its content.  
However, I could not find a way to pass an existing secret to the Helm Chart, and it seems like `argocd-notifications-secret` is hard-coded. 

### Related helm chart

argo-cd

Edit the `values.yaml` with my existing secret. For example:

```
notifications:
  secret:
    # -- Whether helm chart creates notifications controller secret
    create: false

    # -- notifications controller Secret name
    name: "my-secret"

    # -- Whether to use existing secret. If true, assumes that the existing secret name is notifications.secret.name
    useExistingSecret: true
```

What I'm currently doing is setting `notifications.secret.create` to `false`.
However, the Chart still forces me to use `argocd-notifications-secret` as my secret name.

### Additional context

_No response_

ranrubin:
Resolves #2575

Checklist:

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@a6e57359b32af9a54d5b3b6603011f50629a0a05 # v40.1.2
+        uses: renovatebot/github-action@78bdcb3bffa5e95e646183ca0a2ac2895abd6a20 # v40.1.3
         with:
           configurationFile: .github/configs/renovate-config.js
           # renovate: datasource=docker depName=ghcr.io/renovatebot/renovate
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@78bdcb3bffa5e95e646183ca0a2ac2895abd6a20 # v40.1.3
+        uses: renovatebot/github-action@c134f38c01867794a23c35e86660bd9609caf30c # v40.1.4
         with:
           configurationFile: .github/configs/renovate-config.js
           # renovate: datasource=docker depName=ghcr.io/renovatebot/renovate
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -3019,6 +3019,12 @@ notifications:
     # -- Whether helm chart creates notifications controller secret
     create: true
 
+    # -- notifications controller Secret name
+    name: "argocd-notifications-secret"
+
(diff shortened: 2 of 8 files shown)


charts/argo-cd/values.yaml
```diff
@@ -3019,6 +3019,12 @@ notifications:
     # -- Whether helm chart creates notifications controller secret
     create: true
 
+    # -- notifications controller Secret name
+    name: "argocd-notifications-secret"
```
mkilchhofer, line 3023: I dont think that this is configurable. And when it is, we need to pass it somehow to the notification controller, he is talking to the kube-api to read that secret directly.

.github/workflows/renovate.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@a6e57359b32af9a54d5b3b6603011f50629a0a05 # v40.1.2
+        uses: renovatebot/github-action@78bdcb3bffa5e95e646183ca0a2ac2895abd6a20 # v40.1.3
```
yu-croco, line 29: maybe we don't want to add this change in the same PR ?

ranrubin: I didn’t make this change intentionally , I assumed it is part of the project automation. Should I revert this commit then?

yu-croco: Oh it's automation...why it worked... 😮 
It's merged in https://github.com/argoproj/argo-helm/pull/2579, so please fetch remote branch.

charts/argo-cd/values.yaml
```diff
@@ -3019,6 +3019,12 @@ notifications:
...
+
+    # -- Whether to use existing secret. If true, assumes that the existing secret name is notifications.secret.name
+    useExistingSecret: false
```
mkilchhofer, line 3026: Do we really need an additional bool for that. I assume we can add a note on the existing bool `create`.

ranrubin: I agree that having one would be best. 

The reason I added a boolean is the [conditional rule ](https://github.com/argoproj/argo-helm/blob/fc620a3d51afb448fd3a14ed86a8949ed3e47af6/charts/argo-cd/templates/argocd-notifications/clusterrole.yaml#L41) in the ClusterRole. I didn't know the code well enough when I first addressed it to understand why it's there so I didn't want to touch it.  

However, now that we are passing `- --secret-name={{ .Values.notifications.secret.name }}` I guess we can remove the condition completely, and always set the rule :
```
  - apiGroups:
    - ""
    resourceNames:
    - {{ .Values.notifications.secret.name }}
    resources:
    - secrets
    verbs:
    - get
```

Do you agree?

mkilchhofer: Yes I think we can remove that conditional 👍

mkilchhofer (approved):
LGTM. Thanks for this contribution and implementing all changes requested during review 🙏

PR #2589, opened 2024-03-19
docs(argo-cd): update alb ingress docs backend-protocol annotation

andrii29:
`alb.ingress.kubernetes.io/backend-protocol: HTTP` in ALB ingress cause infinite 307 redirect loop

Checklist:

### DIFF
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -190,7 +190,7 @@ server:
     annotations:
       alb.ingress.kubernetes.io/scheme: internal
       alb.ingress.kubernetes.io/target-type: ip
-      alb.ingress.kubernetes.io/backend-protocol: HTTP
+      alb.ingress.kubernetes.io/backend-protocol: HTTPS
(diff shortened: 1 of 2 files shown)


pdrastil:
Hi @andrii29 thanks for PR. Is this really needed with the rest of example that makes the server running in HTTP mode without redirects?

```yaml
configs:
  params:
    server.insecure: true
```

andrii29:
Ty for clarification

pdrastil:
@andrii29 The value depends on the use case. If you want to terminate on Pod with your certs then HTTPS without insecure is correct combination. If you use AWS ACM for certs then you run with insecure and terminate on ALB.

andrii29:
We use both: aws acm and secure backend

pdrastil:
I'm not sure how this should work to be honest as there is always one place where you terminate:

1) HTTPS -> ALB + ACM (termination happens here) -> HTTP -> Server
2) HTTPS -> ALB -> HTTPS -> Server (termination happens here)

Issue #2613, opened 2024-03-29
Generate CRD spec based on upstream tag, similar to kube-prometheus-stack helm chart


DrFaust92:
### Is your feature request related to a problem?

Every time the argocd image is updated, there is a chance that crds will be behind and there might be some mistmatch on the app (as it assume crds of a certain version)

### Related helm chart

argo-cd

### Describe the solution you'd like

have a script to generate CRDs based on upstream tag similar to kube-prometheus-stack helm chart

https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/hack/update_crds.sh

PR #2614, opened 2024-03-29
feat(argo-cd): Generate crd from upstream

DrFaust92:
Closes https://github.com/argoproj/argo-helm/issues/2613

Checklist:


DrFaust92:
need to fix this to support the custom templating (annotations and if to enable crd) and this should be good. but i would love to get input from maintainer if this is a valid approach.

@jmeridth @pdrastil @yu-croco , what do you think?

yu-croco:
Hi @DrFaust92 , thank you for your PR.
IMO:
- Auto generation of CRD sounds nice to me.
- In addition to support `.Values.crds.xxxx` , it's also nice to run this script when [Renovate bumps version](https://github.com/argoproj/argo-helm/blob/main/.github/workflows/renovate.yaml). 👍 

@mbevc1 @mkilchhofer@jmeridth @pdrastil @tico24 
WDYT? 🙋

pdrastil:
@yu-croco Sounds OK, however upstream do not have templating on CRD resources / additional conditional block we have. You can see removed sections in PR. In both cases auto-generation or manual sync I always have to modify these sections by hand. You can see removals in this PR.

DrFaust92:
@pdrastil I'll add logic to template this, just wanted to know if it's worth the effort in the first place 🤗

mkilchhofer:
Maybe we need some yq logic to only copy over the CRD.spec?🤣

PR #2616, opened 2024-03-30
feat(argo-cd): add authentication for builtin Redis

afrimberger:
This PR introduces authentication for the builtin Redis. It can be enabled by setting `redis.auth.enabled=true`. The secrets are auto-generated and stored in a secret.

Checklist:

### DIFF
--- a/charts/argo-cd/templates/redis/secret.yaml
+++ b/charts/argo-cd/templates/redis/secret.yaml
@@ -1,4 +1,22 @@
-# lookup existing secret
+# lookup existing secret with Helm's `lookup` function.
+# At first, it might seem strange to use this function in the ArgoCD ecosystem.
+# However, the purpose of this Helm Chart is to bootstrap ArgoCD. For this use-case
+# you don't have necessarily ArgoCD available, yet.
+#
+# Basically, ArgoCD can be installed with two approaches:
+# 1. Via Helm -> Helm's `lookup` function works as expected
+# 2. Via ArgoCD -> `lookup` will always return an empty value because ArgoCD
+#    calls `helm template` internally. In this case, you need to ignore the password
+#    fields of the secret in `ignoreDifferences`:
+#    ```yaml
+#    ignoreDifferences:
+#    - kind: Secret
+#      name: argocd-redis
+#      jsonPointers:
+#        - /data/redis-password
+#        - /data/redis-password-admin
+#        - /data/redis-password-default
+#    ```
 {{- $secretName :=  include "argo-cd.redis.fullname" . -}}
 {{- $secretObj := (lookup "v1" "Secret" .Release.Namespace $secretName) | default dict }}
 {{- $secretData := (get $secretObj "data") | default dict }}
--- a/charts/argo-cd/templates/redis/secret.yaml
+++ b/charts/argo-cd/templates/redis/secret.yaml
@@ -1,0 +1,60 @@
+# lookup existing secret
+{{- $secretName :=  include "argo-cd.redis.fullname" . -}}
+{{- $secretObj := (lookup "v1" "Secret" .Release.Namespace $secretName) | default dict }}
+{{- $secretData := (get $secretObj "data") | default dict }}
+# generate random password if secret doesn't exist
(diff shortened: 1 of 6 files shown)


charts/argo-cd/templates/redis/secret.yaml
```diff
@@ -0,0 +1,60 @@
+# lookup existing secret
+{{- $secretName :=  include "argo-cd.redis.fullname" . -}}
+{{- $secretObj := (lookup "v1" "Secret" .Release.Namespace $secretName) | default dict }}
```
mkilchhofer, line 3: Using lookup inside the Argo ecosystem is IMHO a bad idea. Argo CD uses helm only as a template engine (argocd only executes `helm template ...`).
And `helm template` does not support looking up values for security reasons.

afrimberger: Hi @mkilchhofer, Thanks for bringing up this topic! If we're talking about ArgoCD in general, I fully agree. However, in this particular case the Helm Chart is meant for bootstrapping ArgoCD. Basically, this can be achieved in two ways:
1. Install directly via Helm -> Lookup works
2. ArgoCD manages ArgoCD -> `ignoreDifferences` for the password fields needs to be specified

I would really love to see a more secure default installation of ArgoCD. Especially, because not all clusters support `NetworkPolicies` out of the box (e.g. Flannel doesn't). 

I suggest to add an example for `ignoreDifferences` and some explanation to the README to address the questions arising by using lookup. WDY?

mkilchhofer: I mean chaning the default installation without being in-line with with upstrean kustomize manifests is not what we want.
Upstream is our reference for the default installation as mentioned in the charts README.

I appreciate a more secure default installation but it has to be promoted first as a PR in the upstream repo (https://github.com/argoproj/argo-cd).

afrimberger: According to the Chart's README:
> Hence, it makes sense to try to achieve a similar output of rendered .yaml resources when calling helm template using the default settings in values.yaml

According to this, the default should be `redis.auth.enabled=false` (as it is in the current implementation). Additionally, a user can easily enable a more secure installation by setting `redis.auth.enabled` to `true`. 
Sounds reasonable to me.

mkilchhofer:
Hi @afrimberger sorry to get back to you soo late. We were working intensively on the recent release to address exactly this Authentication issue.

See:
- https://github.com/argoproj/argo-helm/releases/tag/argo-cd-6.10.0
- #2705
- https://github.com/argoproj/argo-cd/security/advisories/GHSA-9766-5277-j5hr

We are now open for further improvement on this section, can you check which parts of your PR will work with the new setup?

afrimberger:
Hi @mkilchhofer, a security advisory can sometimes speed up things. Well played :-).

> We are now open for further improvement on this section, can you check which parts of your PR will work with the new setup?

According to the least privileges principle the ArgoCD user should *not* have admin rights. The Redis ACLs could look like this (non HA):
```
user default on +@all -@admin -@dangerous ~* &* >{{ $defaultUserPassword | b64dec }}
user admin on +@all ~* &* >{{ $adminUserPassword | b64dec }}
```

To allow an admin user to connect to Redis for debugging purposes, at least two passwords need to be generated.

One option would be to only restrict the default user. Though, this would restrict the debug capabilities.

mkilchhofer:
@github-actions: Not stale

@afrimberger Maybe you can file an issue upstream (https://github.com/argoproj/argo-cd/issues) to enhance the redis access even more?

thecosmicfrog:
@github-actions: Not stale

thecosmicfrog:
@github-actions: Not stale

mkilchhofer:
@afrimberger and also to the maintainers ( @yu-croco @tico24 @mbevc1 @jmeridth )

**Question**:
Do we want to close this? The **authN** part of this PR is already implemented via the secret init.

The PR here goes further with some kind ot **authZ**. Do users need this?

mbevc1:
Sounds good to close this now.

PR #2625, opened 2024-04-04
fix(argo-cd): make automountServiceAccountToken configurable

stefan-caraiman:
Keeps the default behaviour of having automountServiceAccountToken set to true for deployments/statefulsets.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.10.6
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 6.7.10
+version: 6.7.11
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump argo-cd to v2.10.6
+    - kind: added
+      description: Add configurable automountServiceAccountToken
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -725,6 +725,7 @@ NAME: my-release
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | controller.affinity | object | `{}` (defaults to global.affinity preset) | Assign custom [affinity] rules to the deployment |
+| controller.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account into the pod. |
 | controller.clusterRoleRules.enabled | bool | `false` | Enable custom rules for the application controller's ClusterRole resource |
 | controller.clusterRoleRules.rules | list | `[]` | List of custom rules for the application controller's ClusterRole resource |
 | controller.containerPorts.metrics | int | `8082` | Metrics container port |
@@ -806,6 +807,7 @@ NAME: my-release
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | repoServer.affinity | object | `{}` (defaults to global.affinity preset) | Assign custom [affinity] rules to the deployment |
+| repoServer.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account into the pod. |
 | repoServer.autoscaling.behavior | object | `{}` | Configures the scaling behavior of the target in both Up and Down directions. |
 | repoServer.autoscaling.enabled | bool | `false` | Enable Horizontal Pod Autoscaler ([HPA]) for the repo server |
 | repoServer.autoscaling.maxReplicas | int | `5` | Maximum number of replicas for the repo server [HPA] |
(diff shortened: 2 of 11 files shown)


mkilchhofer:
Just to probably learn something new: is there a difference setting it on the deployment/statefulset or ServiceAccount?
Because this toggle is already supported on all ServiceAccounts:

```markdown
$ grep automount README.md
| controller.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
| repoServer.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
| server.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
| dex.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
| redis.serviceAccount.automountServiceAccountToken | bool | `false` | Automount API credentials for the Service Account |
| applicationSet.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
| notifications.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
```

stefan-caraiman:
> Just to probably learn something new: is there a difference setting it on the deployment/statefulset or ServiceAccount? Because this toggle is already supported on all ServiceAccounts:
> 
> ```
> $ grep automount README.md
> | controller.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
> | repoServer.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
> | server.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
> | dex.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
> | redis.serviceAccount.automountServiceAccountToken | bool | `false` | Automount API credentials for the Service Account |
> | applicationSet.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
> | notifications.serviceAccount.automountServiceAccountToken | bool | `true` | Automount API credentials for the Service Account |
> ```

Yep 😄  as per the [K8s docs](https://v1-24.docs.kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server):

```
The pod spec takes precedence over the service account if both specify a automountServiceAccountToken value.
```

Setting automountServiceAccountToken on a ServiceAccount applies the setting to all pods using that account, defaulting them to either mount or not mount the service account token. Setting it on a Deployment or StatefulSet allows for specific control over that particular workload, overriding the ServiceAccount's default.

mkilchhofer:
> Setting automountServiceAccountToken on a ServiceAccount applies the setting to all pods using that account, defaulting them to either mount or not mount the service account token. Setting it on a Deployment or StatefulSet allows for specific control over that particular workload, overriding the ServiceAccount's default.

Ok but do you create your own SA's or why do you like this change? By default every deployment/component of Argo CD has its own SA :)

stefan-caraiman:
Not really no, we do not create our own SA, we rely on the defaults, but due to the following Azure AKS Security policy:
```
"Kubernetes clusters should disable automounting API credentials" - Disable automounting API credentials to prevent a potentially compromised Pod resource to run API commands against Kubernetes clusters.
```

TLDR: We need to basically set `automountServiceAccountToken: false` to both the SA and Deployments/StatefulSets for ArgoCD and then mount the `serviceaccounttoken` through extra `volumes/volumeMounts` for resources that do need access to the API.

Example [values](https://cert-manager.io/docs/installation/best-practice/#best-practice-helm-chart-values) & [issue](https://github.com/cert-manager/cert-manager/issues/5254) from cert-manager on this topic.

In the end the result is the same as with `automountServiceAccountToken: true`, but it is a "hardening" requirement to be compliant with the Azure Security Score 😅

mkilchhofer:
> In the end the result is the same as with `automountServiceAccountToken: true`, but it is a "hardening" requirement to be compliant with the Azure Security Score 😅

Oh boy 🙈 . I now definitively learned something new. But IMHO this is only to satisfy the security scanner, it doesn't really improve security.
There was a good talk at KubeCon: Malicious compliance ;)
[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=9weGi0csBZM)

Lets wait on some other opinions of my colleagues.

stefan-caraiman:
Yep totally agree. No harm though in adding in the option to customize it since it defaults to true just as before. 😄

mbevc1:
Hah, nice work Microsoft :)

Good reference @mkilchhofer :+1: 

Happy we add it in for people on Azure as long as the defaults are not changing.

stefan-caraiman:
@mbevc1 indeed, no defaults are changing so the change is harmless 👍

stefan-caraiman:
@mbevc1 @mkilchhofer just to exemplify further, the following 2 example `values` achieve the same injection of the Service account token into the pod in order to make K8s api calls.

```
# Example 1 ( the default now and with this MR too)
# Automatically injection done as explained here
# https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#serviceaccount-admission-controller
controller:
  automountServiceAccountToken: true
  serviceAccount:
    automountServiceAccountToken: true
```
same as with
```
# Example 2 values for disabling auto-mounting and opting to inject it via volumes instead
# for "hardened" AKS setups
controller:
  automountServiceAccountToken: false
  serviceAccount:
    automountServiceAccountToken: false
  volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: serviceaccount-token
      readOnly: true
  volumes:
    - name: serviceaccount-token
      projected:
        defaultMode: 0444
        sources:
        - serviceAccountToken:
            expirationSeconds: 3607
            path: token
        - configMap:
            name: kube-root-ca.crt
            items:
            - key: ca.crt
              path: ca.crt
        - downwardAPI:
            items:
            - path: namespace
              fieldRef:
                apiVersion: v1
                fieldPath: metadata.namespace
```

yu-croco (approved):
Thank you for your contribution. LGTM.

mbevc1 (approved):

mbevc1:
Thanks @stefan-caraiman

mkilchhofer (approved):

PR #2660, opened 2024-04-23
fix(argo-cd): Fix static value secret name for argocd-server ingress tls secretName

btwseeu78:
Checklist:


btwseeu78:
Ok so let me explain the use case first, we use external secrets operator to pull certificates for diff clusters, then kubed is used to sync across namespaces. All our certificates are shared with multiple namespaces and a specific format to easily differentiate between what certificates for what hosts.Argocd tls means we need to just create another object to store the secrets,though we already have that with a diff name.
Also now we are doing it with extra-tls which does not feel right though it solves the issue.

Also what's the harm having it parameterised with a default values.
Also it's personal but I'm not ok to put ca data in values file and create the secret.

zmx:
We have the same use case as btwseeu78

mkilchhofer:
Somehow related:
- https://github.com/argoproj/argo-helm/pull/2741

mkilchhofer:
@pdrastil In my opinion it totally makes sense to have this configurable. Depending on the method used on the Ingress (e.g. SSL passthrough or not), operator docs uses different names for the secret:
https://argo-cd.readthedocs.io/en/stable/operator-manual/ingress/

So it is possible to use one TLS config for Argo CD's namespace-internal or in-cluster communication and use another TLS config on the Ingress.

PR #2682, opened 2024-05-07
feat(argo-cd): Always install applicationset crd

ebuildy:
ApplicationSet is required by argocd-server, cannot be "opt-out".

If ApplicationSet CRD is not installed, we see lot of errors:

```
W0507 19:44:33.469804 7 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.11/tools/cache/reflector.go:169: failed to list *v1alpha1.ApplicationSet: the server could not find the requested resource (get applicationsets.argoproj.io)
E0507 19:44:33.469828 7 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.11/tools/cache/reflector.go:169: Failed to watch *v1alpha1.ApplicationSet: failed to list *v1alpha1.ApplicationSet: the server could not find the requested resource (get applicationsets.argoproj.io)
```

Relate:
- https://github.com/argoproj/argo-cd/pull/15432

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [ ] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [ ] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #2688, opened 2024-05-12
fix(argo-cd): Always create ApplicationSet as following upstream

yu-croco:
Related to https://github.com/argoproj/argo-helm/pull/2682

Checklist:

### DIFF
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -2638,9 +2638,6 @@ repoServer:
 
 ## ApplicationSet controller
 applicationSet:
-  # -- Enable ApplicationSet controller
-  enabled: true
-
   # -- ApplicationSet controller name string
   name: applicationset-controller
 
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.11.0
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 6.8.1
+version: 6.9.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
-      description: Always install applicationset crd
+      description: Always create ApplicationSet as following upstream
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -278,6 +278,9 @@ For full list of changes please check ArtifactHub [changelog].
 
 Highlighted versions provide information about additional steps that should be performed by user when upgrading to newer version.
 
+### 6.9.0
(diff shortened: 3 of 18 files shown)


charts/argo-cd/values.yaml
```diff
@@ -2638,9 +2638,6 @@ repoServer:
 
 ## ApplicationSet controller
 applicationSet:
-  # -- Enable ApplicationSet controller
-  enabled: true
```
yu-croco, line 2642: Is it better to write changelog in README...? 🤔

mbevc1 (approved):

pdrastil (approved):

charts/argo-cd/values.yaml
pdrastil: I would mention that even in readme as enteprise users with proxy repositories might be suprised that this is now always installed.

yu-croco: @pdrastil 
Thank you for your comment. I will add 🙋

yu-croco: I added in this commit https://github.com/argoproj/argo-helm/pull/2688/commits/822ab42195754e79af05f4adb783295c438e2f61 . 🫡

mkilchhofer (approved):
Thanks @yu-croco 👍👌

rouke-broersma:
Based on the original proposal for merging the appset codebase with the argocd codebase it seems to me like it was intended for appset to be optional. Therefore it seems like this is a bug in argocd and the correct fix would be to remove the requirement in argocd. See: https://github.com/argoproj/argo-cd/pull/8743#discussion_r824240111

pdrastil:
Hi @hsmade I'm not sure why installation would be impossible. As you state you have centrally managed Argo by some team. Same team can also install AppSet controller as they would be cluster admins anyway that are allowed to install / upgrade CRDs. AppSet resource is restricted only to the namespace where Argo is installed and resources are usable only by admins via kubectl or GitOps (you cannot create AppSet resource via UI as end user). In multi-tenant setup it's a good practice to restrict Argo namespace only to admins and do not allow tenants modify it as it contains sensitive integration secrets for whole organisation. Tenants in such case are usually given their own project that allows them specific resources (not AppSet, CRDs or other global resources) / namespaces (not argo, kube-system, etc.) / OIDC binding (not using admin account) so they can self-provision their application. Use of Application in any namespace feature is also tied to correct use of projects if you want to give users access to K8s cluster via kubectl.

hsmade:
>As you state you have centrally managed Argo by some team
No, we install argo on a centrally managed k8s cluster. So we don't own the k8s cluster ourselves, and there's other teams that also install argo on their namespaces. So the argo we install, only manages our own namespace(s). Therefor, the cluster admins limit what CRDs we can access.

felipeng:
> Based on the original proposal for merging the appset codebase with the argocd codebase it seems to me like it was intended for appset to be optional. Therefore it seems like this is a bug in argocd and the correct fix would be to remove the requirement in argocd. See: [argoproj/argo-cd#8743 (comment)](https://github.com/argoproj/argo-cd/pull/8743#discussion_r824240111)
> 
> Wouldn't it make more sense to allow deploying the CRD without the applicationset controller? I don't use the appset controller and I don't want to use the appset controller so this is just wasting resources in my cluster for no benefit.

Same use case here. We don't use AppSet, and it is not possible to avoid its creation. I set AppSet replicas to 0; however, there are other resources created. Not a big deal, but why create Kubernetes objects if they will not be used?
https://github.com/argoproj/argo-helm/issues/4043

PR #2729, opened 2024-05-31
feat(argo-cd): Add option to disable API checks in all servicemonitor

psychomantys:
#### What this PR does / why we need it:

This helm attempts to verify the API version during deployment processes. However, in some environments, direct access to the API is not available beforehand, which complicates the deployment.

This situation is particularly problematic in two notable scenarios:

- **Offline Manifest Generation and Application**: When the manifests are generated in one environment (e.g., a **CI/CD** pipeline or a local machine) and then applied to another cluster, there's no way to verify the API version during the generation phase.
- **Manifest Generation Using Kustomize**: In environments where manifests are created using tools like **kustomize**,  the version check becomes redundant and can lead to deployment issues.

My PR proposes modifications that bypass the API version check under these specific conditions, ensuring smoother and more flexible deployments in environments lacking direct API connectivity.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.11.2
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 7.0.0
+version: 7.0.1
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: Represent cluster credentials as a map
+    - kind: added
+      description: Add option to disable API checks in all servicemonitor
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -803,6 +803,7 @@ NAME: my-release
 | controller.metrics.service.type | string | `"ClusterIP"` | Metrics service type |
 | controller.metrics.serviceMonitor.additionalLabels | object | `{}` | Prometheus ServiceMonitor labels |
 | controller.metrics.serviceMonitor.annotations | object | `{}` | Prometheus ServiceMonitor annotations |
+| controller.metrics.serviceMonitor.disableAPICheck | bool | `false` | Disable API check to enable servicemonitor |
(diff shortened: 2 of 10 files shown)


mkilchhofer:
Hi @psychomantys
Thanks for your contribution. I apreciate everybody who helps us to make the chart better and better.

I am not a fan of this approach. We had this request already one time and declined it.
Background is that all tools around helm (Kustomize, Helmfile, Argo CD, ...) have the option to pass in all the available APIs of kube (CRDs, Core API versions, etc).

> Offline Manifest Generation and Application

For this, you need to execute helm with the parameter `--api-versions`:

```bash
helm template argocd \
  oci://ghcr.io/argoproj/argo-helm/argo-cd \
  --api-versions monitoring.coreos.com/v1 \
  --values argocd-values.yaml
```

> Manifest Generation Using Kustomize

There is first class support for this also in kustomize:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmCharts:
- name: argo-cd
  repo: oci://ghcr.io/argoproj/argo-helm
  version: 7.0.0
  releaseName: argocd
  apiVersions:                 #
  - monitoring.coreos.com/v1   # see this structure
  valuesInline:
    controller:
      metrics:
        enabled: true
        serviceMonitor:
          enabled: true
    server:
      metrics:
        enabled: true
        serviceMonitor:
          enabled: true

    repoServer:
      metrics:
        enabled: true
        serviceMonitor:
          enabled: true

```

psychomantys:
Thank you for your contribution and for sharing your perspective!

I understand your point of view, and i agree that relying solely on external tools for api checks is ideal for wider range of use cases. To address this, I've implemented the functionality with an optional variable. By default, the functionality will be disabled, preserving the current behavior and ensuring compatibility with existing workflows.

For users who prefer this approach, the variable can be easily enabled in the `values.yaml` file. This means no other changes are required for `Applications` using methods like Kustomize, Helmfile or helm.

This approach aims to strike a balance between providing flexibility and maintaining compatibility with existing tooling and workflows.

mkilchhofer:
> I understand your point of view, and i agree that relying solely on external tools for api checks is ideal for wider range of use cases. To address this, I've implemented the functionality with an optional variable.

I think I was clear with my explanation, no? 😃 

We already discussed this inside the maintainers chat and we'd like to decline the idea.
Our opinion is that this is a redundant implentation of a feature which is already provided by Helms core functionality.

I mean it is hard to say "no", we really appreciate every contribution. But supporting all edge cases of users also means more work (due to more testing) for us.
I hope you understand our decision?

Back to topic? What tool(s) are you using?
Maybe we can support you, finding the right parameter for your tool?

PR #2789, opened 2024-06-21
refactor(argo-cd): use dataString in repo credentials

cortopy:
Checklist:

---

Repository credentials [template](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-configs/repository-credentials-secret.yaml) uses `b64enc` function. This makes it working with plugins like [helm-secrets](https://github.com/jkroepke/helm-secrets) very difficult. Multiline secrets like the private key will only be encoded for the first line.

Changing to `stringData` is a non-breaking change but removes any problems around multiline formatting

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.11.3
 kubeVersion: ">=1.23.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 7.2.0
+version: 7.2.1
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,6 +26,8 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
+    - kind: changed
+      description: Repository credentials Secret will use stringData to increase compatibility with plugins like helm-secrets
     - kind: removed
       description: Remove `server.certificate.secretName`, as the expected secret name is static (argocd-server-tls)
     - kind: removed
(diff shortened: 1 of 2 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -26,6 +26,8 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
+    - kind: changed
+      description: Repository credentials Secret will use stringData to increase compatibility with plugins like helm-secrets
     - kind: removed
```
tico24, line 31: You need to remove the old stuff please.

cortopy: thanks @tico24 for the review. What do you mean by the old stuff. This PR only changes something that's there. I'm not sure if I understod this changelog entirely

tico24: You can look at other prs for examples. The changelog should only contain your change. Remove the old stuff.

mbevc1: He was referring to old changelog entries, only the last one needs bo he in the Pr as this is automatically constructed as part of release.

mkilchhofer:
> This makes it working with plugins like [helm-secrets](https://github.com/jkroepke/helm-secrets) very difficult. Multiline secrets like the private key will only be encoded for the first line.

Are you sure, did you test that behavior?
Never seen a difference between using 
- `stringData` without `b64enc`
- `data` plus using `b64enc`

pdrastil:
@mkilchhofer There could be a difference for our templating implementation. I think that we might hit problems with multi-line inputs with line breaks.

Consider input: 

```yaml
key: | 
  foo
  bar
anotherKey: baz
```

result for `stringData` template:

```yaml
stringData:
  key: foo
bar # no identantion or | to generate valid yaml
  anotherKey: baz
```

but `data` will be

```yaml
data:
  key: foo_\n_bar_base64_encoded
  anotherKey: baz_base64_encoded
```

cortopy:
My bad, the problems I was facing were due to misconfig. Apologies.

Issue #2795, opened 2024-06-25
Missing permission in ArgoRollout cluster role


Ahmed-Elkollaly:
### Describe the bug

I can't use the recent feature `scaleDown`, used in migration to Argo rollout, due to missing permission in `clusterrole` in the helm chart in argo-rollout v2.36.1 https://github.com/argoproj/argo-helm/blob/main/charts/argo-rollouts/templates/controller/clusterrole.yaml#L60

https://argo-rollouts.readthedocs.io/en/stable/migrating/
```
  workloadRef:                                 # Reference an existing Deployment using workloadRef field
    apiVersion: apps/v1
    kind: Deployment
    name: rollout-ref-deployment
    scaleDown: onsuccess
```
The error from argo-rollout controller pod
```time="2024-06-25T12:38:07Z" level=error msg="deployments.apps \"rollout-ref-deployment\" is forbidden: User \"system:serviceaccount:argo-rollouts:argo-rollouts\" cannot update resource \"deployments\" in API group \"apps\" in the namespace \"dev\"```

### Related helm chart

argo-rollouts

### Helm chart version

2.36.1

### To Reproduce

Steps to reproduce the issue
1. Apply deployment
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/instance: rollout-canary
  name: rollout-ref-deployment
spec:
  replicas: 1                              
  selector:
    matchLabels:
      app: rollout-ref-deployment
  template:
    metadata:
      labels:
        app: rollout-ref-deployment
    spec:
      containers:
        - name: rollouts-demo
          image: argoproj/rollouts-demo:blue
          imagePullPolicy: Always
          ports:
            - containerPort: 8080
```
            
 2. Apply Rollout
 
```
apiVersion: argoproj.io/v1alpha1               # Create a rollout resource
kind: Rollout
metadata:
  name: rollout-ref-deployment
spec:
  replicas: 5
  selector:
    matchLabels:
      app: rollout-ref-deployment
  workloadRef:                                 # Reference an existing Deployment using workloadRef field
    apiVersion: apps/v1
    kind: Deployment
    name: rollout-ref-deployment
    scaleDown: onsuccess
  strategy:
    canary:
      steps:
        - setWeight: 20
        - pause: {duration: 10s}
```

3. check argo-rollout controller pod logs
[...]

PR #2816, opened 2024-07-09
fix(argo-rollouts): missing permission in cluster role

muma378:
Add missing permissions `apps.Deployment.update` in default SA #2795
Resolves https://github.com/argoproj/argo-helm/issues/2795

Checklist:

### DIFF
--- a/charts/argo-rollouts/templates/controller/role.yaml
+++ b/charts/argo-rollouts/templates/controller/role.yaml
@@ -68,6 +68,7 @@ rules:
   - get
   - list
   - watch
+  - update
 # services patch needed to update selector of canary/stable/active/preview services
 # services create needed to create and delete services for experiments
 - apiGroups:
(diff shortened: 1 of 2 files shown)


yu-croco:
Hi @muma378 ,
Thank you for your PR, but as I mentioned [here](https://github.com/argoproj/argo-helm/issues/2795#issuecomment-2216840460), upstream's manifest is not released, so we will wait for upstream's release.

FYI; when Renovate makes PR as following upstream, we usually handle manifests' update at the same time.
e.g.) https://github.com/argoproj/argo-helm/pull/2794

charts/argo-rollouts/templates/controller/role.yaml
```diff
@@ -68,6 +68,7 @@ rules:
   - get
   - list
   - watch
+  - update
```
yu-croco, line 71: We also have [clusterrole](https://github.com/argoproj/argo-helm/blob/main/charts/argo-rollouts/templates/controller/clusterrole.yaml#L59-L69), so it should be fixed at the same time. 🤔 

*Though upstream is not released yet.

Issue #2824, opened 2024-07-12
Possible Bug in workflow cluster role missing permissions


menzbua:
### Describe the bug

Hi,

I have a problem with Argo workflows HTTP Template. When I run a workflow with using this template, the workflow lasts for 30 seconds or many minutes until it is finished. I researched the behavior and found an error message in the workflow-controller pod:
```
time="2024-07-11T11:35:07.711Z" level=warning msg="error updating taskset" error="failed patching taskset: workflowtasksets.argoproj.io \"http-template-5bskv\" is forbidden: User \"system:serviceaccount:argocd:argo-workflows-workflow-controller\" cannot patch resource \"workflowtasksets/status\" in API group \"argoproj.io\" in the namespace \"argocd\"" namespace=argocd workflow=http-template-5bskv
```
I've installed Argo workflows with the latest Helm-Chart. When I take a look into the file "charts/argo-workflows/templates/controller/workflow-controller-cluster-roles.yaml" I see that the permissions to patch the workflowtasksets is missing. In the namespaced role file "charts/argo-workflows/templates/controller/workflow-role.yaml" the permission is given https://github.com/argoproj/argo-helm/blob/2653aef414ab6a5d8617af75f04190a8f7da28dc/charts/argo-workflows/templates/controller/workflow-role.yaml#L50-L56 I've added the permission manually to the ClusteRole in my cluster, and all HTTP template workflows are now running without errors.

I think that the following permissions should be granted in the ClusterRole too:
```
- apiGroups:
  - argoproj.io
  resources:
    - workflowtasksets/status
    - workflowartifactgctasks/status
  verbs:
    - patch                  
```

I can resolve this issue for myself and send you a pull-request.

### Related helm chart

argo-workflows

### Helm chart version

0.41.11

### To Reproduce

1. Add Workflow with HTTP Template
2. Check the logs of the workflow-controller POD
3. Check the time until the Workflow is finished. It should be a very long period of time if the job is carried out several times.

### Expected behavior

The Workflow with HTTP Template JOB should finish in a few seconds without error messages in the workflow-controller logs.

oEscal:
I am also facing this issue. I think it should be reopened.

derekcha:
I also encountered this issue and resolved it by manually modifying the ClusterRole.
However, I believe the root cause lies in the Helm chart itself, and it should be fixed there.
Therefore, I think this issue should be reopened.

PR #2825, opened 2024-07-12
fix(argo-workflows): add clusterrole to avoid workflow-controller errors

menzbua:
This Pull-Request should fix https://github.com/argoproj/argo-helm/issues/2824

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.8
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.41.11
+version: 0.41.12
 icon: https://argo-workflows.readthedocs.io/en/stable/assets/logo.png
 home: https://github.com/argoproj/argo-helm
 sources:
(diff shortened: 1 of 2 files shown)


tico24:
I don't see this in the upstream clusterrole that the workflow-controller uses. https://github.com/argoproj/argo-workflows/releases/download/v3.5.8/install.yaml

(cluster role starts on line 1016)

In fact those permissions don't appear anywhere in the upstream yaml. This should be fixed upstream before here.

menzbua:
Okay, I will open an issue in the upstream project, and check if we can add the rule there, too.

MikeJansen:
Thank you so much for this!  My simple 2 step HTTP demo workflow went from taking 30 minutes to complete to 50 seconds, and reporting status properly now.  

I just added an argo-fix manifest with a new cluster role and an additional binding  for the controller account until it's fixed upstream in argo-workflows.

MikeJansen:
> Okay, I will open an issue in the upstream project, and check if we can add the rule there, too.

With the latest helm chart this issue still exists.  Did an issue get added upstream?

Issue #2634, opened 2024-04-09
Add gatewayAPI resources as an Ingress alternative.


sboulkour:
### Related helm chart

argo-cd

### Describe the solution you'd like

As gatewayAPI has gone GA in late 2023, many ingress controllers have enabled or are enabling the use of HTTPRoute and other gatewayAPI resources. Also, as it is meant to replace ingress ressources, we would need to have the ability to deploy ArgoCD with ingresses or with gateway API objects.

### Describe alternatives you've considered

Deploying HTTPRoute etc manually alongside the Argo charts. 

### Additional context

- Current implementations https://gateway-api.sigs.k8s.io/implementations/
- Current implementation work being done in ingress-nginx https://github.com/kubernetes/ingress-nginx/issues/7517#issuecomment-2025441175

peschmae:
Currently GRPC is still experimental, but the HTTP(S) routes are GA.

We've already got around to add this as custom templates on an internal setup, and I'll try to check next week, if we can contribute those back upstream, or what might be needed to make this happen.

sboulkour:
Hello everyone,

FTR, Gateway API released v1.1.0 yesterday.

> GRPCRoute has graduated to GA (v1) and is now part of the Standard Channel

https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.1.0

pdrastil:
Hi - I've created draft PR that you can try - feedback from community about initial interface is welcome as lots of ingress controller implementations are still under development.

lucasfcnunes:
Go away, bot!
Can someone pin this issue?

pdrastil:
@lucasfcnunes Pinned - I'm currently experimenting with Contour with latest 1.1 API and this is still planned.

raweber42:
Is this planned to be implemented anytime soon? I am happy to contribute.

PR #2689, opened 2024-05-12
feat(argo-cd): Add Gateway API support

pdrastil:
Resolves: https://github.com/argoproj/argo-helm/issues/2634

Draft to get initial feedback from community

Checklist:

* [ ] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [ ] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [ ] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [ ] Any new values are backwards compatible and/or have sensible default.
* [ ] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #2965, opened 2024-10-04
feat(argo-cd): gateway support

tduverge:
Resolves: https://github.com/argoproj/argo-helm/issues/2634

Adding some changes to https://github.com/argoproj/argo-helm/pull/2689

Especially :
- Support gateway creation
- Support GKE objects associated to gateway
- Rearrange httproute key paths to be clear that httproute is gateway dependent
- Remove `API` suffix from key `gatewayAPI` since it doesn't appear in any other key (like ingress or service)

I've also created a [PR on pdrastil branch](https://github.com/pdrastil/argo-helm/pull/73), you can keep its as you wish.

Checklist:


mkilchhofer:
> Any news on what is preventing this from being merged? This would be a really nice addition.

I think the community did not yet standardize on a pattern how Gateway API is integrated into arbitrary Helm charts. Just why I did not invest a lot if time into this PR:
- Bitnami Helm charts do not offer Gateway API support for their charts
- It is unclear whether an Application chart should deploy a `kind: Gateway` resource. From the ideology (of the Gateway API creators), the Gateway resource is maintained by the cluster operators, not the app operators:
  > image
  >
  > Source: https://gateway-api.sigs.k8s.io/
- I think the two points are the reason why, @pdrastil did not get a review either.

stemarks:
I am also in need of this feature, I may implement it myself for the time being until a decision is made. 

> * It is unclear whether an Application chart should deploy a `kind: Gateway` resource. From the ideology (of the Gateway API creators), the Gateway resource is maintained by the cluster operators, not the app operators:

My personal view is that the app should only create the HTTPRoute, there will likely only be a single gateway shared between multiple services. However there is argument that if this is the only application that requires gateway then it should be included, or if this is the first application to be deployed, which in a lot of scenarios it is, then a gateway would be good here. 

Personally I would plan to deploy ArgoCD and have a chart that deploys my gateway as part of an Argo app then the remote access would work once its up and running. I have multiple apps so need the gateway separated. 

I think it would be good to at least implement HTTPRoute now and have the gateway discussion separately.

tduverge:
Hi.

From my pov, as a GKE user, it's not that clear that gateway should be shared between services. I had a case where a single HTTPRoute with a wrong configuration destroyed all the associated load balancer.

The implematation I made follow the same logic than Ingress for many cloud providers where deploying an Ingress creates a load balancer. This is how it works with the current Ingress already in that chart, and it could be easier to migrate from Ingress to Gateway for users if they can do the same with both.

Moreover, in that PR, the gateway deployment is facultative, as I said [in the note there.](https://github.com/argoproj/argo-helm/pull/2965/files#diff-9eb59d56febda5e405748483497dfdc0c2dfbe9b4f386a12f0180c98d99051e0R2205). This let the ability to anyone to choice what he wants.

Of course, I can separate the gateway if this can accelerate things.

PR #2968, opened 2024-10-07
feat(argo-cd): Expose cluster's project as a label

nbarrientos:
Given that [Argo CD's ApplicationSet cluster generator ](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Generators-Cluster/) allows selecting clusters using arbitrary labels, it'd be useful for operators to have the clusters' project available as label to for instance easily deploy a single application to all the clusters of a given project.

More info: https://github.com/argoproj/argo-helm/discussions/2949

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.12.4
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 7.6.8
+version: 7.6.9
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: added
-      description: Set affinity in redis secret-init job.
+      description: Add the cluster project as a label.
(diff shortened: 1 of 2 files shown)


mkilchhofer:
Are you aware that it is already possible today?

```yaml
  clusterCredentials: {}
    # mycluster:
    #   server: https://mycluster.example.com
    #   labels: {}                               # <-- see here
    #   annotations: {}
    #   config:
    #     bearerToken: "<authentication token>"
    #     tlsClientConfig:
    #       insecure: false
    #       caData: "<base64 encoded certificate>"
```

nbarrientos:
> Are you aware that it is already possible today?

Thanks for taking a look. Yes, I am, but this approach requires setting it in two places, the project itself and then the label. With this patch it's only set in one place and propagated automatically, avoiding duplication and possible misalignments upon (bogus) configuration changes. This is the added value of this patch :)

mkilchhofer:
Can you point me to a resource where the key of the label is documented? :)
I am curious why we should add the label `argocd.argoproj.io/project` on all project scoped clusters (where customers do not use AppSet for example).

nbarrientos:
Hi,

Hehe, I have to confess that made up the resource label key, it's indeed not listed [here](https://argo-cd.readthedocs.io/en/latest/user-guide/annotations-and-labels/) but hey, in my defense it's just an arbitrary label :) I used the `argocd.argoproj.io/` namespace as this information is inherent to Argo CD so it kinda made sense to me. 

Setting it to all clusters could be eventually beneficial when filtering resources using `kubectl`'s label selectors, but being frank at least for me the main use-case is to combine this with AppilicationsSets. I could make the automatic labeling optional (even the label key customizable) so it's not added by default (in view of users not using ApplicationsSets as you're mentioning) if that'd increase the chances of having this merged.

Thanks again.

nbarrientos:
Would a chart-specific label make this more appealing, perhaps? :angel:

Issue #3011, opened 2024-11-07
Support PDB `unhealthyPodEvictionPolicy`


fleuryc-owkin:
### Is your feature request related to a problem?

[kube-linter-action](https://github.com/stackrox/kube-linter-action) complains that `unhealthyPodEvictionPolicy` is not explicitly set :
> (object: xxx/argo-workflows-workflow-controller policy/v1, Kind=PodDisruptionBudget) unhealthyPodEvictionPolicy is not explicitly set (check: pdb-unhealthy-pod-eviction-policy, remediation: Set unhealthyPodEvictionPolicy to AlwaysAllow. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/#unhealthy-pod-eviction-policy for more information.)

### Related helm chart

argo-workflows

### Describe the solution you'd like

Support `unhealthyPodEvictionPolicy` in the PDB definition.

PR #3012, opened 2024-11-07
fix(argo-workflows): Add unhealthyPodEvictionPolicy to workflow controller PDB

fleuryc-owkin:
Resolves https://github.com/argoproj/argo-helm/issues/3011

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.12
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.42.7
+version: 0.42.8
 icon: https://argo-workflows.readthedocs.io/en/stable/assets/logo.png
 home: https://github.com/argoproj/argo-helm
 sources:
@@ -18,3 +18,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: added
       description: Support lifecycle and terminationGracePeriodSeconds for server
+    - kind: added
+      description: explicitly set unhealthyPodEvictionPolicy to IfHealthyBudget in controller PDB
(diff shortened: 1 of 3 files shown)


charts/argo-workflows/templates/controller/workflow-controller-deployment-pdb.yaml
```diff
@@ -14,6 +14,7 @@ spec:
   {{- else }}
   minAvailable: 0
   {{- end }}
+  unhealthyPodEvictionPolicy: IfHealthyBudget
```
tico24, line 17: Feels like this should be user-configurable as some users may well prefer `AlwaysAllow`

tico24 (changes requested):

tico24:
Also this only went stable fairly recently didn't it (maybe 1.31 from memory)? How will this work for older k8s clusters?

charts/argo-workflows/templates/controller/workflow-controller-deployment-pdb.yaml
fleuryc-owkin: Done ✅ 

I ran `./scripts/helm-docs.sh` but the attribute doesn't show up in the doc. What did I do wrong ?

charts/argo-workflows/values.yaml
```diff
@@ -344,6 +344,9 @@ controller:
     enabled: false
     # minAvailable: 1
     # maxUnavailable: 1
+    
+    # -- [unhealthyPodEvictionPolicy] is only valid for v1.31+
```
tico24, line 348: v1.31 of what?
It's valid for earlier versions but it's behind a gate.

fleuryc-owkin: Fixed the comment ✅

charts/argo-workflows/Chart.yaml
```diff
@@ -18,3 +18,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: added
       description: Support lifecycle and terminationGracePeriodSeconds for server
```
yu-croco, line 20: Please remove old one.

PR #3032, opened 2024-11-13
feat(argo-rollouts): configurable automountServiceAccountToken

k-cogswell:
Inspired by https://github.com/argoproj/argo-helm/pull/2625

Adding a configurable option for `automountServiceAccountToken` in argo-rollouts chart. This is my first contribution so please let me know if I am missing anything.

Checklist:

### DIFF
--- a/charts/argo-rollouts/templates/controller/deployment.yaml
+++ b/charts/argo-rollouts/templates/controller/deployment.yaml
@@ -44,6 +44,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "argo-rollouts.serviceAccountName" . }}
+      automountServiceAccountToken: {{ .Values.serviceAccount.automount }}
       containers:
       - image: "{{ .Values.controller.image.registry }}/{{ .Values.controller.image.repository }}:{{ default .Chart.AppVersion .Values.controller.image.tag }}"
         args:
--- a/charts/argo-rollouts/templates/dashboard/deployment.yaml
+++ b/charts/argo-rollouts/templates/dashboard/deployment.yaml
@@ -45,6 +45,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "argo-rollouts.serviceAccountName" . }}-dashboard
+      automountServiceAccountToken: {{ .Values.dashboard.serviceAccount.automount }}
       containers:
       - image: "{{ .Values.dashboard.image.registry }}/{{ .Values.dashboard.image.repository }}:{{ default .Chart.AppVersion .Values.dashboard.image.tag }}"
         imagePullPolicy: {{ .Values.dashboard.image.pullPolicy }}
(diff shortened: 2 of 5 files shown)


charts/argo-rollouts/values.yaml
```diff
@@ -229,6 +229,8 @@ serviceAccount:
   # -- The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
   name: ""
+  # -- Specifies wether a service account should be automounted
+  automount: true
```
yu-croco, line 233: Regarding the comment and default value, please align to https://github.com/argoproj/argo-helm/pull/2625 .
Reason of the default value `false` is also mentioned in the PR.

k-cogswell: @yu-croco The comments and default values have been updated. Thank you for the feedback.

k-cogswell: @yu-croco looking at the referenced PR again, the default value is set to `true`. 

You can also see in this in `main` https://github.com/argoproj/argo-helm/blob/936101865d3090183fc4979546685ece7f78ac23/charts/argo-cd/values.yaml#L834

If you're okay with this I can revert my values file back to the original value of `true`,

mkilchhofer: Yes, lets use `true` as the default. Otherwise it would be a breaking change.

yu-croco: My bad... 🫠  Please revert as `true` .

yu-croco (approved):
Thank you for your contribution. LGTM

charts/argo-rollouts/templates/controller/deployment.yaml
```diff
@@ -44,6 +44,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "argo-rollouts.serviceAccountName" . }}
+      automountServiceAccountToken: {{ .Values.serviceAccount.automount }}
```
mkilchhofer, line 47: Since we are setting the field on the `kind: Deployment` rather than on the `kind: ServiceAccount` can we use `{{ .Values.controller.automountServiceAccountToken }}` instead?

charts/argo-rollouts/templates/dashboard/deployment.yaml
```diff
@@ -45,6 +45,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "argo-rollouts.serviceAccountName" . }}-dashboard
+      automountServiceAccountToken: {{ .Values.dashboard.serviceAccount.automount }}
```
mkilchhofer, line 48: Since we are setting the field on the `kind: Deployment` rather than on the `kind: ServiceAccount` can we use `{{ .Values.dashboard.automountServiceAccountToken }}` instead?

mkilchhofer (changes requested):

jmeridth:
@k-cogswell able to get to the updates that @mkilchhofer is asking for? I ask before this gets marked as stale.  Cheers.

PR #3014, opened 2024-11-07
fix(argo-rollouts): plugin block rendering was incorrect

taer:
Resolves https://github.com/argoproj/argo-helm/issues/2957

Checklist:

* [X] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [X] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [X] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [X] Any new values are backwards compatible and/or have sensible default.
* [X] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [X] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

the config map was getting rendered as 
```
data:
  - location: https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/releases/download/v0.4.0/gatewayapi-plugin-linux-arm64
    name: argoproj-labs/gatewayAPI
```

this changes it to 
```
data:
  trafficRouterPlugins: |-
    - location: https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/releases/download/v0.4.0/gatewayapi-plugin-linux-arm64
      name: argoproj-labs/gatewayAPI
```
as shown in [metricPlugin](https://argoproj.github.io/argo-rollouts/analysis/plugins/#mounting-the-plugin-executable-into-the-rollouts-controller-container) and [trafficPlugin](https://argo-rollouts.readthedocs.io/en/stable/features/traffic-management/plugins/#mounting-the-plugin-executable-into-the-rollouts-controller-container)

PR #3036, opened 2024-11-14
fix(argo-rollouts): Change type of trafficRouterPlugins and trafficRouterPlugins as list

taer:
The change I made here https://github.com/argoproj/argo-helm/pull/3014 broken the examples. 

This works w/ my original change:
```
controller:
  trafficRouterPlugins:
    - name: "argoproj-labs/gatewayAPI"
      location: "https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/releases/download/v0.4.0/gatewayapi-plugin-linux-arm64"
      sha256: "07e2fe515c900899c96777aac0ab490437a3d62c8adb94512ccb8b859325eb11"
```

But the example shows stringification is needed.

I feel having it "look like a real thing instead of a string" is more user friendly.

Checklist:

### DIFF
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v1.7.2
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 2.37.8
+version: 2.37.9
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-rollouts/assets/logo.png
 keywords:
@@ -19,4 +19,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: fixed
-      description: Fixed rendering of plugins in the ConfigMap
+      description: Updated values.yaml examples for the non-stringigied options. No effective change
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v1.7.2
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 2.37.9
+version: 2.38.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-rollouts/assets/logo.png
 keywords:
@@ -19,4 +19,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: fixed
-      description: Updated values.yaml examples for the non-stringigied options. No effective change
+      description: Updated plugin values.yaml example and it's implementation to not need to include the stringification or the plugins block that it used to
--- a/charts/argo-rollouts/values.yaml
+++ b/charts/argo-rollouts/values.yaml
@@ -209,17 +209,15 @@ controller:
 
   # -- Configures 3rd party metric providers for controller
   ## Ref: https://argo-rollouts.readthedocs.io/en/stable/analysis/plugins/
-  metricProviderPlugins: {}
-    # metricProviderPlugins:
-    #   - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
-    #     location: "file://./my-custom-plugin" # supports http(s):// urls and file://
+  metricProviderPlugins: []
+    # - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
+    #   location: "file://./my-custom-plugin" # supports http(s):// urls and file://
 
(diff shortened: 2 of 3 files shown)


taer:
My prior change was basically a breaking one.

I assumed that no one was using the values.yaml for plugins since it wasn't rendering according to the docs.

But that ship sailed, and oops.

I will renumber this build to be a non-patch change...  It's just docs, but follows up to the actual breaking change for anyone who worked around the initial issue.

taer:
If y'all are ok w/ this, we should delete 2.37.8 since it was in fact a breaking change.. Thanks @michel-gleeson

charts/argo-rollouts/values.yaml
```diff
@@ -210,14 +210,14 @@ controller:
   # -- Configures 3rd party metric providers for controller
   ## Ref: https://argo-rollouts.readthedocs.io/en/stable/analysis/plugins/
   metricProviderPlugins: {}
-    # metricProviderPlugins: |-
+    # metricProviderPlugins:
     #   - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
     #     location: "file://./my-custom-plugin" # supports http(s):// urls and file://
```
yu-croco, line 215: Thank you for your PR.

Originally we needed to define like this, so `object` was ok.
```yaml
metricProviderPlugins:
  metricProviderPlugins: |-
    - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
      location: "file://./my-custom-plugin" # supports http(s):// urls and file://
```
But now we define like this.
```yaml
metricProviderPlugins:
  - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
    location: "file://./my-custom-plugin" # supports http(s):// urls and file://
```
So I wonder it's better to change type as `list` ?
```diff
-  metricProviderPlugins: {}
+  metricProviderPlugins: []
-    # metricProviderPlugins: |-
-    # metricProviderPlugins:
-    # - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
-    #   location: "file://./my-custom-plugin" # supports http(s):// urls and file://
+    # - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
+    #   location: "file://./my-custom-plugin" # supports http(s):// urls and file://
```

```

```

charts/argo-rollouts/Chart.yaml
```diff
@@ -19,4 +19,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: fixed
-      description: Fixed rendering of plugins in the ConfigMap
+      description: Updated plugin values.yaml example and it's implementation to not need to include the stringification or the plugins block that it used to
```
yu-croco, line 22: If you agree with https://github.com/argoproj/argo-helm/pull/3036/files#r1843061134 , `Correct type of metricProviderPlugins and trafficRouterPlugins as list` is clearer ?

charts/argo-rollouts/values.yaml
taer: I had always assumed the whole indented block was to be uncommented and de-dented.. :)  
ala
```
values:
    # metricProviderPlugins: |-
    #   - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
    #     location: "file://./my-custom-plugin" # supports http(s):// urls and file://
```
becomes
```
values:
   metricProviderPlugins: |-
     - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
       location: "file://./my-custom-plugin" # supports http(s):// urls and file://
```

mostly since the "default" uncommented version was something you replace w/ the commented out things.

This looks fine I'll make the change on Monday

yu-croco: >I had always assumed the whole indented block was to be uncommented and de-dented.. :)

I am the one who added this attribute in argo-helm as following other existing ones. https://github.com/argoproj/argo-helm/blob/main/charts/argo-rollouts/values.yaml#L30-L39
*Though I should've implemented a better way, trim redundant `metricProviderPlugins` in configmap.yaml.

```yaml
metricProviderPlugins: 
    metricProviderPlugins: |-
      - name: "argoproj-labs/sample-prometheus" # name of the plugin, it must match the name required by the plugin so that it can find its configuration
        location: "file://./my-custom-plugin" # supports http(s):// urls and file://
```

charts/argo-rollouts/Chart.yaml
taer: done

yu-croco, line 22: As a result, maybe `Type of trafficRouterPlugins and trafficRouterPlugins is changed from object to list` is enough ?

yu-croco (approved):

jmeridth (approved):

PR #3048, opened 2024-11-22
fix(argo-workflows): Make Argo Agent and Artifact GC permissions optional for workflows SA

tico24:
This changes the permissions on the workflows Service Account.

Previously we gave permissions for the Argo Agent and Artifact GC to every workflow SA, this was overly permissive. Now with this PR, these things are disabled by default but can be enabled by the user if those features are used.

Checklist:

### DIFF
--- a/charts/argo-workflows/templates/controller/agent-rb.yaml
+++ b/charts/argo-workflows/templates/controller/agent-rb.yaml
@@ -1,0 +1,29 @@
+{{- if .Values.workflow.rbac.agentPermissions -}}
+  {{- range $namespace := or .Values.singleNamespace false | ternary (list "") (append .Values.controller.workflowNamespaces (coalesce .Values.workflow.namespace (include "argo-workflows.namespace" .)) | uniq)  }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ template "argo-workflows.fullname" $ }}-workflow-agent
+  labels:
+    {{- include "argo-workflows.labels" (dict "context" $ "component" $.Values.controller.name "name" $.Values.controller.name) | nindent 4 }}
+    {{- with $namespace }}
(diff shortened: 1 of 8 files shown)


yu-croco (approved):

yu-croco (approved):

jmeridth (approved):

agilgur5:
For reference, this resolves the now public https://github.com/argoproj/argo-helm/security/advisories/GHSA-h974-w8pg-cx73

charts/argo-workflows/templates/controller/agent-rb.yaml
```diff
@@ -0,0 +1,29 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "argo-workflows.fullname" $ }}-workflow-agent
+subjects:
+  - kind: ServiceAccount
+    name: {{ $.Values.workflow.serviceAccount.name }}
+    {{- with $namespace }}
+    namespace: {{ . }}
+    {{- end }}
+  {{- range $.Values.workflow.rbac.serviceAccounts }}
```
agilgur5, line 23: It is similar for the Agent SA, although a bit different, as it is configured by [`spec.serviceAccountName`](https://github.com/argoproj/argo-workflows/blob/1392ef516fd69a2cae875d746c1155f940e4948c/workflow/controller/agent.go#L154) (and at most one Agent runs per Workflow) and so _can_ overlap with other Workflow Pods if those do not have per-template SAs set (i.e. [`spec.templates[].serviceAccountName`](https://argo-workflows.readthedocs.io/en/release-3.5/fields/#template))

agilgur5 (commented):
For best practice least privilege, this RBAC should ideally be attached to separate optional SAs as well (not just Role/RoleBindings). The UX therein is a tad non-trivial (as I mentioned in the advisory), since the user has to know how/when to use those SAs

PR #3075, opened 2024-12-11
fix(argo-cd): Fix repo-server honorLabels config template

symaras:
This fixes a bug introduced in https://github.com/argoproj/argo-helm/commit/0061e14563571f1cd0447fcac0e90407a600bbc4#diff-ef4a32480a793da4e68fa5b4e320523c5a1acf86064769abfc1d97a1462143eeR38 where the honorLabels config for the repo-server was put inside the `with` statement. In that case, if `.Values.repoServer.metrics.serviceMonitor.scheme` is set, the honorLabels value can't be found because the `.Values` object is not accessible from within the `with` block.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: Bump argocd-extension-installer to v0.0.8
+    - kind: changed
+      description: Fix honorLabels config template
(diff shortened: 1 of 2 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
       description: Bump argocd-extension-installer to v0.0.8
```
mbevc1, line 30: You also need to remove old changelog entry

symaras: thanks done

mbevc1 (approved):

yu-croco (approved):

PR #3126, opened 2025-01-16
fix(argo-workflows): container and mainContainer imagePullPolicy

BaptisteDeporte:
Checklist:

I found a strange behavior with the mainContainer "imagePullPolicy" (and the same with executor since they work similary). 

With the current release, if I apply the chart with the following values.yaml
```yaml
mainContainer:
  imagePullPolicy: "Always"
```
My controller config-map does not have the expecting output : 
```yaml
data:
  config: |
    nodeEvents:
      enabled: true
    workflowEvents:
      enabled: true
```

When we only set **imagePullPolicy** option for the **mainContainer** or the **executor**, it’s not applied. 

Plus, the [doc](https://artifacthub.io/packages/helm/argo/argo-workflows#workflow-main-container) specifies that the default value for those fields are "Values.images.pullPolicy". So, I expect the config map to contains according pull policy for executor and main container when I apply the chart with an empty values.yaml, but it's not the case, I obtain the same config map mentioned earlier. 

It’s my first contribution, I have read the CONTRIBUTING.md but any feedback is welcome !

### DIFF
--- a/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
+++ b/charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
@@ -31,7 +31,7 @@ data:
     {{- with .Values.controller.initialDelay }}
     initialDelay: {{ . }}
     {{- end }}
-    {{- if or .Values.mainContainer.resources .Values.mainContainer.env .Values.mainContainer.envFrom .Values.mainContainer.securityContext}}
+    {{- if or .Values.images.pullPolicy .Values.mainContainer}}
     mainContainer:
       imagePullPolicy: {{ default (.Values.images.pullPolicy) .Values.mainContainer.imagePullPolicy }}
       {{- with .Values.mainContainer.resources }}
@@ -47,7 +47,7 @@ data:
       securityContext: {{- toYaml . | nindent 8 }}
       {{- end }}
     {{- end }}
-    {{- if or .Values.executor.resources .Values.executor.env .Values.executor.args .Values.executor.securityContext}}
+    {{- if or .Values.images.pullPolicy .Values.executor}}
     executor:
       imagePullPolicy: {{ default (.Values.images.pullPolicy) .Values.executor.image.pullPolicy }}
       {{- with .Values.executor.resources }}
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.6.2
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.45.4
+version: 0.45.5
 icon: https://argo-workflows.readthedocs.io/en/stable/assets/logo.png
 home: https://github.com/argoproj/argo-helm
 sources:
@@ -16,5 +16,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: added
-      description: Support configuring workflow events in the controller
+    - kind: fixed
+      description: Fix main container and executor pull policy option


charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
```diff
@@ -31,7 +31,7 @@ data:
     {{- with .Values.controller.initialDelay }}
     initialDelay: {{ . }}
     {{- end }}
-    {{- if or .Values.mainContainer.resources .Values.mainContainer.env .Values.mainContainer.envFrom .Values.mainContainer.securityContext}}
+    {{- if or .Values.images.pullPolicy .Values.mainContainer}}
     mainContainer:
       imagePullPolicy: {{ default (.Values.images.pullPolicy) .Values.mainContainer.imagePullPolicy }}
       {{- with .Values.mainContainer.resources }}
```
tico24, line 37: Not sure this will work any more will it?

You need to add the if `.Values.mainContainer.resources` back in here for example. repeat for all the other things.

charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
```diff
@@ -31,7 +31,7 @@ data:
     {{- with .Values.controller.initialDelay }}
     initialDelay: {{ . }}
     {{- end }}
-    {{- if or .Values.mainContainer.resources .Values.mainContainer.env .Values.mainContainer.envFrom .Values.mainContainer.securityContext}}
+    {{- if or .Values.images.pullPolicy .Values.mainContainer}}
```
tico24, line 34: This would also cause errors now if there's data missing. `.Values.mainContainer` is too broad

charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
BaptisteDeporte: Well, I re-read my MR description, and it seems it was unclear my bad. I updated it so please, take a look

With this modification, I expect the config map to contains a "mainContainer" in any case with : 
a. if they are no "mainContainer.imagePullPolicy" in the values files, I expect it to be the same as "Values.images.pullPolicy"
b. if they are a "mainContainer.imagePullPolicy" in the values file, I want the according values to match
(and same with executor)

The issue with the current release is this test :

```yaml
{{- if or .Values.mainContainer.resources .Values.mainContainer.env .Values.mainContainer.envFrom .Values.mainContainer.securityContext}}
```

It don't work if we only set "mainContainer.imagePullPolicy" value. But if we trick the chart with a dummy resources for example, the test is triggered and we got a correct pull policy (the one specified by the user, and the images.pullPolicy as a fallback)

I run the followings tests, and it seems to work as described bellow : 

## With an empty values.yaml

With the current release (0.45.4) we obtain this config-map :
```yaml
data:
  config: |
    nodeEvents:
      enabled: true
    workflowEvents:
      enabled: true
```

And with this chart, this one : 
```yaml
data:
  config: |
    mainContainer:
      imagePullPolicy: Always
    executor:
      imagePullPolicy: Always
    nodeEvents:
      enabled: true
    workflowEvents:
      enabled: true
```

## With a pullPolicy set

With the following values.yaml :
```yaml
mainContainer:
  imagePullPolicy: "IfNotPresent"
```
With the current release, we obtain the same config-map as before. And with this chart we obtain : 

```yaml
data:
  config: |
    mainContainer:
      imagePullPolicy: IfNotPresent
    executor:
      imagePullPolicy: Always
    nodeEvents:
      enabled: true
    workflowEvents:
      enabled: true
```

I made those tests with a freshly installed minikube.

charts/argo-workflows/templates/controller/workflow-controller-config-map.yaml
BaptisteDeporte: Argo Workflow run with an empty values.yaml (or with wrong keys) and is able to run workflow (tested with minikube, latest release and this [dummy workflow](https://raw.githubusercontent.com/argoproj/argo-workflows/master/examples/hello-world.yaml)).

But I understand that we want something more predictable, so I can rollback to the more specific test (with `mainContainer.resources`, `mainContainer.securityContext` etc...)

PR #3135, opened 2025-01-24
feat(github): Analyse past release times

tico24:
I'm far from a python expert so please don't laugh too much.

I wanted to understand whether or not we are releasing at a faster pace than before (given we are a more organised team and we have automation in the likes of Renovate now).

This script calculates the release time for each upstream release and plots the values on a graph.

Checklist:


mkilchhofer:
> I'm far from a python expert so please don't laugh too much.

Sadly my python knowhow is even worse :D I totally missed that Python era - I am still on pure bash or Go ^^

Shall we present the output of this on the GH pages over there?
https://argoproj.github.io/argo-helm/ (maybe under /stats or so)

tico24:
> Shall we present the output of this on the GH pages over there?

short answer is “yes, eventually, but not in this PR”. 

We should use gha to build the data as we go rather than this expensive query for the whole history. Once we have that then we can look into publishing.

tico24:
Could I get some reviews please?

mkilchhofer:
> Could I get some reviews please?

I am okay with that :) 👍 
Do the Argoproj maintainers have any plans on the SLA? I saw the SLA graph and I am curious if it is only an assumption by you ( @tico24 ) or if it comes from the project.

tico24:
The SLA comes from a time that @jmeridth told me. But I can't find it written anywhere.

mkilchhofer:
@tico24 maybe documentation regarding what permissions the PAT requires would be nice, WDYT?

tico24:
Ah yeh I'll do that. It just needs public read, so super minimal.

mkilchhofer:
Do we have a explanation why we have a negative data point?
image

tico24:
> Do we have a explanation why we have **two** negative data points?


>(that negative number for argocd is somehow correct... I released the helm chart before the upstream version was released... I have no idea why).

https://cloud-native.slack.com/archives/C0574J7GR5Z/p1737707991527589

jmeridth (approved):

yu-croco (approved):

PR #3157, opened 2025-02-04
fix(argo-cd): Fix conditional generation of ssh and tls configmaps

DaThumpingRabbit:
Since respectively versions [7.7.11](https://github.com/argoproj/argo-helm/releases/tag/argo-cd-7.7.11) and [7.7.16](https://github.com/argoproj/argo-helm/releases/tag/argo-cd-7.7.16), the configmaps `ssh-known-hosts` and `tls-certs` can be disabled completely from the chart generation  
However, the volumes and volumeMounts of the different elements are not conditioned to be removed if the configmaps are not generated and it causes boot issues due to missing volume configs

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.14.1
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 7.8.0
+version: 7.8.1
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump argo-cd to v2.14.1
+    - kind: fixed
+      description: Make ssh and tls volumes conditional to the configmap creation
--- a/charts/argo-cd/templates/argocd-applicationset/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-applicationset/deployment.yaml
@@ -267,10 +267,14 @@ spec:
             {{- with .Values.applicationSet.extraVolumeMounts }}
               {{- toYaml . | nindent 12 }}
             {{- end }}
+            {{- if .Values.configs.ssh.create }}
             - mountPath: /app/config/ssh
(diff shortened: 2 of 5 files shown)


mkilchhofer:
Hi @DaThumpingRabbit 

> Since respectively versions [7.7.11](https://github.com/argoproj/argo-helm/releases/tag/argo-cd-7.7.11) and [7.7.16](https://github.com/argoproj/argo-helm/releases/tag/argo-cd-7.7.16), the configmaps ssh-known-hosts and tls-certs can be disabled completely from the chart generation

The idea is not to make them optional, the idea was to manage those ConfigMaps outside the Helm chart (some customers want that).
But Argo CD requires those infos, so in my opinion your proposal doesn't make sense. Maybe extending the parameter documentation is better?

DaThumpingRabbit:
Hey @mkilchhofer  
Thanks for the swift answer

I am a bit surprised because none of those configs are actually used in my deployments (I even override the `ssh.knownHosts` to an empty string) so I was happy to see that they were made optional  
I understand that some people want to manage them aside from the chart, but then shouldn't they define `extraVolumes` and `extraVolumeMounts` to reference them ? (although those values do not exist for the server and the repo-server)

This is just the idea I would have regarding this use-case, but let me know if that doesn't make sense for you and I'll close this PR

mkilchhofer:
> I am a bit surprised because none of those configs are actually used in my deployments

Our main goal of argo-helm is that we offer only what is supported from upstream.
Some ConfigMaps and Secrets are directly read via Kubernetes API from argocd-server (without mounting it) while some requires mounting.

You can see that upstream also deploys "empty" ConfigMaps:
- [argocd-cm.yaml](https://github.com/argoproj/argo-cd/blob/v2.14.1/manifests/base/config/argocd-cm.yaml)
- [argocd-cmd-params-cm.yaml](https://github.com/argoproj/argo-cd/blob/v2.14.1/manifests/base/config/argocd-cmd-params-cm.yaml)
- [argocd-rbac-cm.yaml](https://github.com/argoproj/argo-cd/blob/v2.14.1/manifests/base/config/argocd-rbac-cm.yaml)
- [argocd-secret.yaml](https://github.com/argoproj/argo-cd/blob/v2.14.1/manifests/base/config/argocd-secret.yaml)
- [argocd-tls-certs-cm.yaml](https://github.com/argoproj/argo-cd/blob/v2.14.1/manifests/base/config/argocd-tls-certs-cm.yaml)

> but then shouldn't they define extraVolumes and extraVolumeMounts to reference them

In my opinion -> no. No because we also need to support our chart users. When users use custom names for these extra mounts, it is pretty hard for us :)

Does that make sense?

DaThumpingRabbit:
Ah yes I see  
It makes complete sense, thanks for clarifying that !

I will close this PR then and keep my values as they are now

Issue #3174, opened 2025-02-19
Define a liveness probe for the argo-workflows server


Julienlin:
### Is your feature request related to a problem?

We experienced a case where the argo-workflows server died and was not responding. This could have been fixed easily by a restart from initiated by Kubernetes.

### Related helm chart

argo-workflows

### Describe the solution you'd like

We would like to have something similar to what the argo-workflows controller has: 

```yaml
          livenessProbe: {{ .Values.server.livenessProbe | toYaml | nindent 12 }}
```
values.yaml:
```yaml
  # -- Configure liveness [probe] for the server
  # @default -- See [values.yaml]
  livenessProbe:
    httpGet:
      port: 2746
      path: /
    failureThreshold: 3
    initialDelaySeconds: 90
    periodSeconds: 60
    timeoutSeconds: 30
```

### Additional context

It comes to our attention that the argo workflow server does not provide a health check endpoint (see https://github.com/argoproj/argo-workflows/discussions/7066). But we could use the same endpoint as the readiness probe to have something that can be used as a liveness probe.

PR #3175, opened 2025-02-20
feat(argo-workflows): Support livenessProbe to server

yu-croco:
Resolves https://github.com/argoproj/argo-helm/issues/3174

Checklist:

### DIFF
--- a/charts/argo-workflows/templates/server/server-deployment.yaml
+++ b/charts/argo-workflows/templates/server/server-deployment.yaml
@@ -87,16 +87,21 @@ spec:
               {{- end }}
             initialDelaySeconds: 10
             periodSeconds: 20
-          {{- if .Values.server.readinessProbe.enabled }}
-          readinessProbe:
+          {{- if .Values.server.livenessProbe.enabled }}
+          livenessProbe:
             httpGet:
-              path: {{ .Values.server.readinessProbe.httpGet.path }}
-              port: {{ .Values.server.readinessProbe.httpGet.port }}
-            initialDelaySeconds: {{ .Values.server.readinessProbe.initialDelaySeconds }}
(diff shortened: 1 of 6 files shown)


charts/argo-workflows/values.yaml
```diff
@@ -776,6 +776,27 @@ server:
...
+  ## livenessProbe for server
+  ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+  livenessProbe:
+    # -- Enable Kubernetes liveness probe for server
+    enabled: false
+    httpGet:
+      # -- Http port to use for the liveness probe
+      port: 2746
+      # -- Http path to use for the liveness probe
+      path: /
```
tico24, line 788: I'm thinking out loud here, but this might be expensive. If server auth is enabled this would trigger quite a lot of k8s api calls to populate the tables etc.

There's no liveness/readiness probe upstream, right?

Ideally we need to find a quiet endpoint that'll just return a clean http 200

charts/argo-workflows/templates/server/server-deployment.yaml
```diff
@@ -87,6 +87,22 @@ spec:
               {{- end }}
             initialDelaySeconds: 10
             periodSeconds: 20
+          {{- if .Values.server.livenessProbe.enabled }}
+          livenessProbe:
+            httpGet:
+              path: {{ .Values.server.livenessProbe.httpGet.path }}
+              port: {{ .Values.server.livenessProbe.httpGet.port }}
+              {{- if .Values.server.secure }}
+              scheme: HTTPS
```
tico24, line 96: this is only true if https is enabled on the pod. Which I don't think the chart does by default (but upstream does)

tico24 (changes requested):

jmeridth (approved):

PR #3216, opened 2025-03-27
chore(argo-cd): Upgrade redis-ha

alexelent-carta:
diff: [DandyDeveloper/charts@redis-ha-4.29.4...redis-ha-4.33.2](https://github.com/DandyDeveloper/charts/compare/redis-ha-4.29.4...redis-ha-4.33.2)

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #3219, opened 2025-03-29
chore(deps): Support helm dependency update

yu-croco:
Related to https://github.com/argoproj/argo-helm/pull/3216 .

Checklist:

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,6 +9,7 @@
     ":gitSignOff"
   ],
   "labels": ["renovate"],
+  "postUpdateOptions": ["helmUpdateSubChartArchives"],
   "includePaths": [
     "**/charts/argo-workflows/Chart.yaml",
     "**/charts/argo-cd/Chart.yaml",


renovate.json
```diff
@@ -9,6 +9,7 @@
     ":gitSignOff"
   ],
   "labels": ["renovate"],
+  "postUpdateOptions": ["helmUpdateSubChartArchives"],
```
yu-croco, line 12: Ref: https://docs.renovatebot.com/modules/manager/helmv3/

tico24:
Is the problem that "redis-ha" is set to "false" further down in the config? I don't believe you need the thing you've added although I am very occasionally wrong. 

Originally I set renovate to only look for the upstream Argo projects, that's why that line was put in. But it shouldn't be there now.

yu-croco:
@tico24 @mkilchhofer 
Thank you for your reviews.

> https://github.com/argoproj/argo-helm/blob/main/renovate.json#L103

I completely missed this. 😇 


> Maybe it's also excluded since upstream uses a specific version?
> https://github.com/argoproj/argo-cd/blob/master/manifests/ha/base/redis-ha/chart/requirements.yaml

Ah it sounds reasonable. 👀 

Anything we can do in argo-helm...? 🤔

tico24:
Ah yes. That is the reason Past Tim excluded it. Well done. 

So if we are sticking with matching upstream I guess we should do that.

yu-croco:
I see. I don't have strong motivation to update redis-ha immediately, just wondered if I can do anything about the original PR.
Thank you again for the explanation. 😁

Issue #3241, opened 2025-04-09
Documentation and the acutal values.yaml file


ubaidilyas:
### Describe the bug

The default value and the actual value in the values.yaml file are not same. For example 

<html><body>

applicationSet.serviceAccount.automountServiceAccountToken | bool | true | Automount API credentials for the Service Account

</body>
</html>

But in the values file line 934 automountServiceAccountToken: false. 

Please update the documentation as it is confusing.

### Related helm chart

argo-cd

### Helm chart version

7.3.4

### To Reproduce

1. Search for applicationSet.serviceAccount.automountServiceAccountToken and will show default value as true.
2. Search for same in values.yaml file and it is false

### Expected behavior

The values should be same and the applicationSet.serviceAccount.automountServiceAccountToken should be mounted.

PR #3242, opened 2025-04-09
fix(argo-cd): Align default value of automountServiceAccountToken as true

yu-croco:
Resolves https://github.com/argoproj/argo-helm/issues/3241 .

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.14.9
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 7.8.23
+version: 7.8.24
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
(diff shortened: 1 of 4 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: v2.14.9
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 7.8.23
+version: 7.8.24
```
yu-croco, line 6: Is it better to bump minor version since we change the default value ...? 🤔

yu-croco: Thank you for your review. I added a commit for this. 🙋 
https://github.com/argoproj/argo-helm/pull/3242/commits/aca5a9968425ebe1c27b157e8d558f5d7baeaacb

mkilchhofer (commented):
@yu-croco i think it is intented that the token is only mounted where the Argo component is talking to the control plane.
Security scanners recommend this.

Our "issue" is that we allow the setting on the ServiceAccount AND on the Deployment/StatefulSet itself.

Maybe we should prepare a deprecation for this (parameter on the ServiceAccount) and a removal in the next major version?

yu-croco:
Hi @mkilchhofer , I just wondered why `ServiceAccount` is the target instead of Deployment/StatefulSet ? 👀 
>Our "issue" is that we allow the setting on the ServiceAccount AND on the Deployment/StatefulSet itself.
>Maybe we should prepare a deprecation for this (parameter on the ServiceAccount) and a removal in the next major version?

PR #3267, opened 2025-04-27
docs(argo-cd): Update documentation for automountServiceAccountToken

mathieu-clnk:
This is a non-functional change that improves the documentation related to the use of `automountServiceAccountToken: false`.

An example is provided to mount the token as per described by [Kubernetes documentation](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#tokenrequest-api).

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v2.14.11
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 7.8.28
+version: 7.8.29
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
-      description: Bump argo-cd to v2.14.11
+      description: Update automountServiceAccountToken documentation
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -743,12 +743,32 @@ controller:
   #  - mountPath: /usr/local/bin/kubelogin
   #    name: custom-tools
   #    subPath: kubelogin
+  #  - mountPath: /var/run/secrets/kubernetes.io/serviceaccount # Default location of the Service Account token as per described under: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#serviceaccount-admission-controller
+  #    name: kube-api-access-argocd
+  #    readOnly: true
 
   # -- Additional volumes to the application controller pod
   volumes: []
   #  - name: custom-tools
   #    emptyDir: {}
-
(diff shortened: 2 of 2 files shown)


mkilchhofer:
Hi @mathieu-clnk 

I'd like to learn something new, can you explain the use case of manually defining the projected volume mount instead of using the automatic mechanism of Kubernetes by leaving the `automountServiceAccountToken` field on the default value of `true`?

mathieu-clnk:
> Hi @mathieu-clnk
> 
> I'd like to learn something new, can you explain the use case of manually defining the projected volume mount instead of using the automatic mechanism of Kubernetes by leaving the `automountServiceAccountToken` field on the default value of `true`?

Hi @mkilchhofer,
The token `expiration` can be set and the `audience` with the last method. 
In case the Kubernetes cluster is having an internal policy (OPA, Keyverno, ...) which restrict the `automountServiceAccountToken: true` (default option), this avoids to create an exception.

Thank you,

mkilchhofer:
Alright, thanks for clarification.
Besides the expiry field, it sounds more like a Malicious Compliance topic ;)


> > In the end the result is the same as with `automountServiceAccountToken: true`, but it is a "hardening" requirement to be compliant with the Azure Security Score 😅
> 
> Oh boy 🙈 . I now definitively learned something new. But IMHO this is only to satisfy the security scanner, it doesn't really improve security. There was a good talk at KubeCon: Malicious compliance ;) [IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=9weGi0csBZM)
> 
> Lets wait on some other opinions of my colleagues.
>
> _Source: https://github.com/argoproj/argo-helm/pull/2625#issuecomment-2037455276_



Maybe we can add a section to the README to explain that, instead of bloating up the values.yaml with more commented example code which is very DRY-unfriendly. What do you think, @mathieu-clnk ?

mathieu-clnk:
Hi @mkilchhofer,
Thank you for your response.

I have watched the video that you have shared and read the quote (though I'm not sure who it's attributed to).

While I understand your point regarding "malicious compliance," I’m not sure it applies directly in this case. Disabling the service account token is not just an Azure Secure Score recommendation; it's also recommended by Kubernetes itself:  https://kubernetes.io/docs/concepts/security/rbac-good-practices/#hardening

Also, I may have misunderstood what Kubernetes has explained ( https://kubernetes.io/docs/concepts/configuration/secret/#serviceaccount-token-secrets) but I thought it was the recommended way to mount the token:

> In Kubernetes v1.22 and later, the recommended approach is to obtain a short-lived, automatically rotating ServiceAccount token by using the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API instead. You can get these short-lived tokens using the following methods: 
> Request a mounted token in a [projected volume](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume) in your Pod manifest. Kubernetes creates the token and mounts it in the Pod.

I haven´t mentioned it ealier, but this method also avoid than other containers inside the POD (such as the init containers) have this token.

Thank you for your suggestion regarding where to document this in the README. I also hesitate to add too much detail to the values.yaml file, especially if it requires a version bump. 

Do you think this kind of information would fit under this section?
https://github.com/argoproj/argo-helm/blob/main/README.md#charts-use-helm-capabilities

Otherwise, I can remove the reference to the volume altogether and instead leave a note like the following to inform users that setting the value to false isn't sufficient on its own:

> If set to false, please ensure the credentials expected by the Kubernetes client-go library are provided.

mathieu-clnk:
Closing this pull request as it seems not needed.

PR #3276, opened 2025-05-01
fix(argo-workflows): Restart server when configMap is updated

paihu:
Checklist:


paihu:
same as here https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-application-controller/statefulset.yaml#L26-L28

yu-croco:
Hi @paihu , thank you for your PR.
I checked on my local if controller and server watched configmap's update. Server didn't restart but controller restarted.
Are you sure that we need the checksum logic to both of server and controller?

The values.yaml I tried was below.

- before
```yaml
argo-workflows:
  controller:
    configMap:
      create: true
```

- after
```yaml
argo-workflows:
  controller:
    configMap:
      create: true
    persistence:
      connectionPool:
        maxIdleConns: 100
        maxOpenConns: 0

```

paihu:
@yu-croco  I apologize, I created a PR after noticing that the server had not restarted, but I hadn’t checked the behavior of the controller.

yu-croco (dismissed):
Thank you for your contribution. LGTM.

yu-croco (dismissed):

tico24:
What problem is this solving? The CM is mounted to the controller, so that makes sense. It is not mounted to the server.

yu-croco:
@tico24 
Ah I thought it was reasonable since the server refers to controller cm ...? 👀 
https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/templates/server/server-deployment.yaml#L56

vladlosev:
The server [retrieves](https://github.com/argoproj/argo-workflows/blob/e55cc7f4b442d3f240810f9b1a74ce4c0450b821/config/controller.go#L58) the configmap directly via the API. It does not have to have it mapped in a volume.

yu-croco:
Oh, this means that we don't neee to restart server. 👀

paihu:
> The server [retrieves](https://github.com/argoproj/argo-workflows/blob/e55cc7f4b442d3f240810f9b1a74ce4c0450b821/config/controller.go#L58) the configmap directly via the API.

That's true. But Only Startup. https://github.com/argoproj/argo-workflows/blob/main/server/apiserver/argoserver.go#L202
 
The CM is only loaded at startup, so changes to the CM during runtime will not be reflected.
We need to restart server  for the changes to take effect.

yu-croco:
It sounds reasonable to accept this PR, but I am not so familiar with upstream's logic... @tico24 @vladlosev WFYT ? 👀

vladlosev (approved):
I guess it's OK then.

yu-croco (approved):

PR #3290, opened 2025-05-07
chore(argo-workflows): Simplify service account name logic in template

daviderli614:
When `controller.serviceAccount.create` is set to false and `controller.serviceAccount.name` is set to 'test', the workflow-controller deployment will use the serviceAccountName 'test', but the serviceaccount 'test' does not exist. This should always be the default when `controller.serviceAccount.create` is false.

https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/templates/controller/workflow-controller-sa.yaml#L1

```yaml
# values.yaml
controller:
  serviceAccount:
    create: false
    name: "test"
```    

image

```
kubectl get deploy argo-workflows-controller -o jsonpath='{.spec.template.spec.serviceAccountName}'
kubectl get deploy argo-workflows-server -o jsonpath='{.spec.template.spec.serviceAccountName}'
```

- When `.Values.controller.serviceAccount.create` is true:
  - Use `.Values.controller.serviceAccount.name` if set
  - Fallback to `argo-workflows.controller.fullname` if not set
- When false, always return "default" (ignore user-provided name)

In addition, I changed the name in the values to k8s -> kubernetes

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.6.7
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 0.45.14
+version: 0.45.15
 icon: https://argo-workflows.readthedocs.io/en/stable/assets/logo.png
 home: https://github.com/argoproj/argo-helm
 sources:
@@ -17,4 +17,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
-      description: Improve argo-workflow controller clusterrole policy
+      description: Simplify service account name logic in template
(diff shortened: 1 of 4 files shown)


yu-croco:
>When controller.serviceAccount.create is set to false and controller.serviceAccount.name is set to 'test', the workflow-controller deployment will use the serviceAccountName 'test', but the serviceaccount 'test' does not exist

I wonder that users are supposed to prepare for `test` Service Account that they created outside of argo-helm?

daviderli614:
> > When controller.serviceAccount.create is set to false and controller.serviceAccount.name is set to 'test', the workflow-controller deployment will use the serviceAccountName 'test', but the serviceaccount 'test' does not exist
> 
> I wonder that this is for users who prepare for `test` Service Account outside of argo-helm, instead of creating in argo-helm?

Maybe no need to be changed, I close.

PR #3238, opened 2025-04-03
fix(argo-cd): Same Secret name and key for all embedded redis options

mkilchhofer:
Hopefully we now takle our confusing redis-secret handling in case users don't use our default install parameters.

Related to:
- https://github.com/argoproj/argo-helm/pull/3228
- https://github.com/argoproj/argo-helm/pull/3228#issuecomment-2774615386 by @sherif-fanous
- https://github.com/argoproj/argo-helm/pull/3228#issuecomment-2776034887 by @mikutas

@sherif-fanous and @mikutas can we have your eyes on this before we merge?

Other/older references:
- https://github.com/argoproj/argo-helm/pull/2839
- https://github.com/argoproj/argo-helm/pull/3058
- https://github.com/argoproj/argo-helm/issues/3057

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [ ] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #2705, opened 2024-05-21
fix(argo-cd): Enable Redis authentication in the default installation

mkilchhofer:
Checklist:

* [ ] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [ ] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [ ] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [ ] Any new values are backwards compatible and/or have sensible default.
* [ ] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #3294, opened 2025-05-09
fix(argo-cd): Use correct Secret name for optional (external-)Redis auth

### LINKED ISSUE #3293: externalRedis and `.externalRedis.username` use incorrect Secret name
Issue description on this PR:
- https://github.com/argoproj/argo-helm/pull/3238

> @mkilchhofer This still breaks the use case where you are using externalRedis and have set the value `.externalRedis.username`. I see that the change was applied to default to the secret `argocd-redis` for `REDIS_PASSWORD` in case of using externalRedis without existing secret, however it was missed for `REDIS_USERNAME`
> 
> ```diff
>             - name: REDIS_USERNAME
>               valueFrom:
>                 secretKeyRef:
> -                 name: argocd-redis
> +                 name: argo-cd-argocd-redis
>                   key: redis-username
>                   optional: true
> ```
> 
> I see this diff when upgrading to a newer chart which breaks our setup. Now, I am forced to use an existing secret.

_Originally posted by @varkey in https://github.com/argoproj/argo-helm/issues/3238#issuecomment-2862002024_

mkilchhofer:
Fixes #3293 

Appears in:
- https://github.com/argoproj/argo-helm/pull/3238#issuecomment-2862002024

Xref:
- https://github.com/argoproj/argo-helm/issues/2705

Can you please give feedback @varkey since you raised this regression?

Checklist:

### DIFF
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -280,12 +280,13 @@ ipFamilies: {{ toYaml . | nindent 4 }}
 secretKeyRef of env variable REDIS_USERNAME
 */}}
 {{- define "argo-cd.redisUsernameSecretRef" -}}
-    {{- if and .Values.externalRedis.host -}}
-name: {{ default (include "argo-cd.redis.fullname" .) .Values.externalRedis.existingSecret }}
+    {{- if .Values.externalRedis.host -}}
+name: {{ default "argocd-redis" .Values.externalRedis.existingSecret }}
 key: redis-username
-optional: true
+optional: {{ if .Values.externalRedis.username }}false{{ else }}true{{ end }}
+
     {{- else -}}
-name: {{ include "argo-cd.redis.fullname" . }}
+name: "argocd-redis"
 key: redis-username
 optional: true
     {{- end -}}
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.0.0
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 8.0.0
+version: 8.0.1
 home: https://github.com/argoproj/argo-helm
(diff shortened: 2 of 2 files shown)


charts/argo-cd/templates/_helpers.tpl
```diff
@@ -280,12 +280,13 @@ ipFamilies: {{ toYaml . | nindent 4 }}
...
-    {{- if and .Values.externalRedis.host -}}
-name: {{ default (include "argo-cd.redis.fullname" .) .Values.externalRedis.existingSecret }}
+    {{- if .Values.externalRedis.host -}}
+name: {{ default "argocd-redis" .Values.externalRedis.existingSecret }}
 key: redis-username
-optional: true
+optional: {{ if .Values.externalRedis.username }}false{{ else }}true{{ end }}
+
     {{- else -}}
-name: {{ include "argo-cd.redis.fullname" . }}
```
varkey, line 288: Should this be changing? If `externalRedis.host` is not set, it would mean bundled redis is being used right? 🤔

mkilchhofer: I just aligned it with all other Redis secret names. (like the few lines below for the password).

The bundled Redis does not specify a username by default. Also the initial implementation (I implemented it) used a templated Secret name:
https://github.com/argoproj/argo-helm/blob/58926e3d790f7c4c42ea082269cb921fb9afaeb5/charts/argo-cd/templates/argocd-configs/externalredis-secret.yaml#L5

The implementation today is to use a static name:
https://github.com/argoproj/argo-helm/blob/1c937d9100c3e752b91e2d3ced93eab43dade355/charts/argo-cd/templates/argocd-configs/externalredis-secret.yaml#L5

mkilchhofer: Oh and the change from templated to static appears here
https://github.com/argoproj/argo-helm/pull/2705/files#diff-a9b2bd4121ef919395acd4a999dc967702e1089485cf69ad46271a0bb8fdb0c7R201

(where upstream introduced the automatic password generation method)

varkey: Thanks, makes sense! 👍🏼

yu-croco (approved):

mbevc1 (approved):

PR #3443, opened 2025-08-18
fix(argo-cd): Correct targetPort for argocd-server Service

roldyxoriginal:
This PR fixes an issue in the argocd-server Service configuration. Currently, the targetPort uses a dynamic value ({{ .Values.server.containerPorts.server }}) which is unnecessary and doesn't follow best practices.

Proposed Changes:

Modified the targetPort in the argocd-server Service to use the port's name (server) instead of its numerical value.

This change improves code readability and chart resilience, as the Service now directly references the named port in the Deployment, which is the recommended practice.

How was this tested?
I deployed the Helm chart with this change in a Kubernetes cluster.
I verified that the argocd-server Service was correctly created and redirects traffic to the Deployment pod without issues.
I confirmed that the ArgoCD UI is accessible and functional.

Checklist:


yu-croco:
Thank you for your PR. Can you please handle below, the items in your PR description ? You checked them but actually you don't handle them yet.
> I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
> I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).


Also ready means all events will be notified to mainteners, so let us change this PR as draft.

roldyxoriginal:
> Thank you for your PR. Can you please handle below, the items in your PR description ? You checked them but actually you don't handle them yet.
> 
> > I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
> > I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
> 
> Also ready means all events will be notified to mainteners, so let us change this PR as draft.

Hi Aikawa,

I've completed the code change and updated the changelog. Is there anything else I'm missing? The change is quite simple.

I have a question: I'm making this change against the 2.14.11 release, but I've noticed the same issue exists in the latest version, v3.1.0. Would you prefer I open a separate PR targeting the newer version as well?

Thank you for your help.

charts/argo-cd/Chart.yaml
```diff
@@ -3,7 +3,8 @@ appVersion: v3.1.0
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 8.3.0
+version: 7.9.2
+
```
yu-croco, line 7: Why does the version go down ? Also no need to new line.

tico24:
> I have a question: I'm making this change against the 2.14.11 release, but I've noticed the same issue exists in the latest version, v3.1.0. Would you prefer I open a separate PR targeting the newer version as well?

We only support the latest release (3.1.0) so that's what this pr needs to target.

charts/argo-cd/Chart.yaml
roldyxoriginal: I am using the latest version of argocd 2.14 and I need to modify the chart without upgrading to version 3. Is it possible to modify the chart version 7.9.1?
I dont know why that new line appears.

tico24: To reiterate. 2.x is no longer supported by the helm chart.

roldyxoriginal:
> > I have a question: I'm making this change against the 2.14.11 release, but I've noticed the same issue exists in the latest version, v3.1.0. Would you prefer I open a separate PR targeting the newer version as well?
> 
> We only support the latest release (3.1.0) so that's what this pr needs to target.

Ok, I understand. :(
I am going to close this PR, sorry for spend your time.

roldyxoriginal:
This pull request was created to address an issue in an unsupported version. It has been removed.

PR #3452, opened 2025-08-27
chore: replace bitnami redis image with bitnamilegacy redis image

### LINKED ISSUE #3422: Stop using docker.io/bitnami images and possibly bitnami charts
### Describe the bug

Bitnami is changing their public catalog policy:
https://github.com/bitnami/charts/issues/35164
We are no longer using Bitnami images for the Argo CD image, but using Bitnami images and charts in our CI. If we don't make a change till Aug 28, builds will start to fail.

We can temporarily replace `docker.io/bitnami` images with `docker.io/bitnamilegacy` images and figure out a long-term solution later. It seems that only tagged images must be replaced, as the latest images, as per Bitnami statement, will still be available.

We also use Bitnamii charts, and while those are not directly impacted, if they use tagged images pointing to `docker.io/bitnami`, we should be using other charts.

### Related helm chart

argo-cd

### Helm chart version

8.2.7

### To Reproduce

Search for `bitnami` in the codebase and handle according to the description.

### Expected behavior

Switching to `bitnamilegacy` where applicable for docker images and for bitnami charts - if we use them, and they don't use `bitnamilegacy`, replace them with other charts.

### Screenshots

_No response_

### Additional context

_No response_

reggie-k:
Closes #3422 
This is just a temporary fix so that we don't hit rate limits on Aug 28th. Bitnami legacy images will not be getting new stable builds so we need to think of an alternative to using them.

Checklist:

* [ ] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -94,7 +94,7 @@ jobs:
         run: |
           kubectl create namespace redis
           helm repo add bitnami https://charts.bitnami.com/bitnami
-          helm install redis bitnami/redis --wait --namespace redis --set auth.password=argocd --set architecture=standalone
+          helm install redis bitnami/redis --wait --namespace redis --set auth.password=argocd --set architecture=standalone --set global.image.repository=bitnamilegacy/redis
 
       - name: Run chart-testing (install)
         run: ct install --config ./.github/configs/ct-install.yaml


.github/workflows/lint-and-test.yml
```diff
@@ -94,7 +94,7 @@ jobs:
         run: |
           kubectl create namespace redis
           helm repo add bitnami https://charts.bitnami.com/bitnami
-          helm install redis bitnami/redis --wait --namespace redis --set auth.password=argocd --set architecture=standalone
+          helm install redis bitnami/redis --wait --namespace redis --set auth.password=argocd --set architecture=standalone --set global.image.repository=bitnamilegacy/redis
```
nitishfy, line 97: will this override the existing image here - `helm install redis bitnami/redis`?

Also, shouldn't we be using official redis image instead as discussed in meeting?

reggie-k: Yes, the image will be overridden. For now, using the bitnami chart, which will not be impacted at the near future as per their statement, but just with a different image. This is a short term solution.

nitishfy: why can't we simply run `helm install redis bitnamilegacy/redis --wait....` than using the `set` flag?

mkilchhofer:
Already handled via
- https://github.com/argoproj/argo-helm/pull/3497

mkilchhofer:
Already handled via
- https://github.com/argoproj/argo-helm/pull/3497

PR #1758, opened 2023-01-11
feat(argo-cd): Upgrade Argo CD to 2.6.0

pdrastil:
Upgrade to 2.6.0 stable version

Upgrade notes: https://argo-cd.readthedocs.io/en/latest/operator-manual/upgrading/2.5-2.6/

---

Note on DCO:

If the DCO action in the integration test fails, one or more of your commits are not signed off. Please click on the *Details* link next to the DCO action for instructions on how to resolve this.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Changes are automatically published when merged to `main`. They are not published on branches.

PR #3489, opened 2025-09-14
feat(argo-cd)!: Allow override of Applications modifications policy per ApplicationSet by default

### LINKED ISSUE #3488: [argo-cd] Add `applicationsetcontroller.enable.policy.override` to params-cm by default?

I have installed ArgoCD using v8.3.7 (latest as of this writing).

I've configured ApplicationSet to discover and generate Applications. I also wanted to prevent accidental deletion of Application and its resources.
I found out that setting these 2 parameters in ApplicationSet is effective in achieving that:
```yaml
  syncPolicy:
    # Prevents ArgoCD from deleting Application resources when the Application is deleted.
    # https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Application-Deletion/
    preserveResourcesOnDeletion: true
    # Prevents ApplicationSet from deleting Application when it is no longer discovered.
    # https://argo-cd.readthedocs.io/en/latest/operator-manual/applicationset/Controlling-Resource-Modification/
    applicationsSync: create-update
```
So if I want to delete an Application and its resources, I'll have to manually delete using cascading deletion from the UI. This makes dangerous operations safer.

However, after setting `syncPolicy.applicationsSync` in ApplicationSet, I found out that it was not working. ApplicationSet was immediately deleting Applications after they were no longer discovered.
It turns out I also have to set `applicationsetcontroller.enable.policy.override` in `argocd-cmd-params-cm`, because the argo-cd helm chart has `applicationsetcontroller.policy: sync` set by default.
```yaml
configs:
  params:
    applicationsetcontroller.enable.policy.override: true
```

According to the documentation, we have to explicitly set this override parameter to override policy for each ApplicationSet (probably for historical reasons to prevent breaking behavior of ArgoCD):
> If the controller parameter --policy is set, it takes precedence on the field applicationsSync. It is possible to allow per ApplicationSet sync policy by setting variable ARGOCD_APPLICATIONSET_CONTROLLER_ENABLE_POLICY_OVERRIDE to argocd-cmd-params-cm applicationsetcontroller.enable.policy.override or directly with controller parameter --enable-policy-override (default to false).
> https://argo-cd.readthedocs.io/en/latest/operator-manual/applicationset/Controlling-Resource-Modification/

Note that in my use-case, I have multiple ApplicationSets, and I want to prevent accidental resource deletion for one while I want full-automatic cascading deletion for the others.

### Related helm chart

argo-cd

Since this behavior is counter-intuitive, set this by default to values.yaml:
```diff
      ## ApplicationSet Properties
      # -- Modify how application is synced between the generator and the cluster. One of: `sync`, `create-only`, `create-update`, `create-delete`
      applicationsetcontroller.policy: sync
+     # -- Enables override of the policy above per ApplicationSet using syncPolicy.applicationsSync.
+     applicationsetcontroller.enable.policy.override: true
```
https://github.com/argoproj/argo-helm/blob/e338ec1cf33e9992629335611e5e8c88e3587732/charts/argo-cd/values.yaml#L434-L442

Since this is technically a breaking change, we have to bump up major chart version.

I have manually set this parameter in my values.yaml for now, but we have to dig quite deep into the ArgoCD documentation and helm values defaults to find out this behavior.
The implication is also quite important: an user who want to prevent immediate Application deletion is not able to do so because of misconfiguration.

### Additional context

_No response_

motoki317:
## Summary

I have changed the params-cm default so users can use ApplicationSet's `syncPolicy.applicationsSync` override by default, without digging deep into the values.yaml defaults.
The issue is described in details at #3488.

closes #3488

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.1.5
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 8.3.7
+version: 9.0.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
-      description: Bump redis_exporter to v1.77.0
+      description: Allow override of Applications modifications policy per ApplicationSet by default
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -815,6 +815,7 @@ NOTE: Any values you put under `.Values.configs.cm` are passed to argocd-cm Conf
 | configs.gpg.annotations | object | `{}` | Annotations to be added to argocd-gpg-keys-cm configmap |
 | configs.gpg.keys | object | `{}` (See [values.yaml]) | [GnuPG] public keys to add to the keyring |
 | configs.params."application.namespaces" | string | `""` | Enables [Applications in any namespace] |
+| configs.params."applicationsetcontroller.enable.policy.override" | bool | `true` | Enables override of the policy above per ApplicationSet using syncPolicy.applicationsSync. |
 | configs.params."applicationsetcontroller.enable.progressive.syncs" | bool | `false` | Enables use of the Progressive Syncs capability |
 | configs.params."applicationsetcontroller.namespaces" | string | `""` (default is only the ns where the controller is installed) | A list of glob patterns specifying where to look for ApplicationSet resources. (e.g. `"argocd,argocd-appsets-*"`) |
 | configs.params."applicationsetcontroller.policy" | string | `"sync"` | Modify how application is synced between the generator and the cluster. One of: `sync`, `create-only`, `create-update`, `create-delete` |
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -434,6 +434,8 @@ configs:
     ## ApplicationSet Properties
     # -- Modify how application is synced between the generator and the cluster. One of: `sync`, `create-only`, `create-update`, `create-delete`
     applicationsetcontroller.policy: sync
+    # -- Enables override of the policy above per ApplicationSet using syncPolicy.applicationsSync.
+    applicationsetcontroller.enable.policy.override: true
     # -- Enables use of the Progressive Syncs capability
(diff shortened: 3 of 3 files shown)


yu-croco:
Hi @motoki317 , thank you for your PR. We follow upstream's manifests, which means we provide same default value as upstream. If you want to change the default value, could you please approach to upstream at first?
FYI; though current values.yaml doesn't provide the interface for `applicationsetcontroller.enable.policy.override`, you can override in your values.yaml.

motoki317:
> Hi @motoki317 , thank you for your PR. We follow upstream's manifests, which means we provide same default value as upstream. If you want to change the default value, could you please approach to upstream at first?

Hi, thanks for the reply.
By upstream you mean the install.yaml at the [releases](https://github.com/argoproj/argo-cd/releases)? As far as I know, [the upstream cmd-params-cm](https://github.com/argoproj/argo-cd/blob/master/manifests/base/config/argocd-cmd-params-cm.yaml) doesn't contain any values. But this chart contains many "default" values.
It is exactly these non-default values causing problem as described in #3488. So my proposed solution is to either add this new default value in valus.yaml, or remove `applicationsetcontroller.policy` altogether, to align with the upstream install.yaml behavior.
The current chart behavior does not align with the upstream install.yaml as far as I know.

motoki317:
Hi, what do you think about this?
Although this change itself is titled "change default", it's more about a fix for the current chart bug.
The current chart does not align with the upstream behavior and doesn't seem right.

yu-croco:
> So my proposed solution is to either add this new default value in valus.yaml, or remove applicationsetcontroller.policy altogether, to align with the upstream install.yaml behavior.

I agree with your latter suggestion. It's best to align with the upstream default, which is currently an empty string `""`, as shown [here](https://github.com/argoproj/argo-cd/blob/v3.1.5/docs/operator-manual/argocd-cmd-params-cm.yaml#L247).
It seems that the default value of `applicationsetcontroller.policy` was originally `sync` when it was released in [v2.6.0](https://github.com/argoproj/argo-cd/blob/v2.6.0/docs/operator-manual/argocd-cmd-params-cm.yaml#L156) and we inherited it in [this PR](https://github.com/argoproj/argo-helm/pull/1758) correctly. We don't have enough capability to follow the specific values of cmd-params and users can override it by themselves, so I think removing it makes sense.

WDYT? 👀  @mbevc1 @mkilchhofer @jmeridth @pdrastil @tico24

motoki317:
Thanks, I see this was resolved in #3540.

yu-croco:
FYI; we removed default values for `.Values.configs.params.xxxx` in https://github.com/argoproj/argo-helm/pull/3540 but changelog in README was wrong. If you want to update to v9.0.x, please also check v9.0.2 ( https://github.com/argoproj/argo-helm/pull/3546 ) in addition to v9.0.0. 🙇

Issue #3301, opened 2025-05-13
Remove spec.preserveUnknownFields in CRDs


jkleinlercher:
### Is your feature request related to a problem?

argo rollouts stays outofsync because of https://github.com/argoproj/argo-rollouts/blob/master/CHANGELOG-v0.1.0_v1.1.0.md#argo-cd-outofsync-status-on-rollout-v100-crds

### Related helm chart

argo-rollouts

### Describe the solution you'd like

remove spec.preserveUnknownFields in all argo rolloutCRDs

### Additional context

See also https://github.com/argoproj/argo-rollouts/issues/1272

gurpalw:
https://github.com/argoproj/argo-rollouts/pull/4277 was merged 3 weeks ago.

yu-croco:
The point is that upstream needs to bump new version. We follow specific version of upstream, not master branch.

yu-croco:
If you are in a hurry, please ask it in upstream.

yu-croco:
FYI; The CRD in upstream is fixed in v1.9.0-rc3, so hopefully we will close this issue when we release v1.9.0 in Helm Chart.
https://github.com/argoproj/argo-rollouts/blob/v1.9.0-rc3/manifests/crds/cluster-analysis-template-crd.yaml

sule26:
@yu-croco When is the Helm chart release coming out?

yu-croco:
@sule26 

https://github.com/argoproj/argo-helm/issues/3301#issuecomment-3202103844

tristanwl:
Looking forward this this as well!

jmeridth:
Closing as completed. The upstream fix (@yu-croco's close condition from January) shipped and is verified in the current chart: `spec.preserveUnknownFields` is absent from all five CRDs in `charts/argo-rollouts/templates/crds/` as of chart 2.43.0 (appVersion v1.10.0, released this week — the fix has actually been present since the chart's v1.9.0 releases). Upgrading the chart with `installCRDs=true` (the default) updates the CRDs in place, and the Argo CD OutOfSync status on the Rollout CRDs disappears with it. If you manage CRDs outside the chart, apply the updated CRDs manually per the README's CRD section.

PR #3498, opened 2025-09-19
fix: remove preserveUnknownFields to avoid OutOfSync in ArgoCD

Rems08:
# fix(argo-rollouts): remove `spec.preserveUnknownFields` from CRDs to avoid Argo CD OutOfSync

## What

Remove the `spec.preserveUnknownFields: false` key from all Argo Rollouts CRD templates.

## Why

* In CRD `apiextensions.k8s.io/v1`, `spec.preserveUnknownFields` is **deprecated/invalid** and the behavior must be expressed with the OpenAPI extension `x-kubernetes-preserve-unknown-fields` inside the schema. Using the top-level field causes drift between desired and live objects. ([[Kubernetes](https://kubernetes.io/docs/reference/using-api/deprecation-guide/)][1])
* Argo CD 3.0+ explicitly calls out that CRDs containing `spec.preserveUnknownFields: false` will show as **OutOfSync** and recommends removing the field. ([[Argo CD](https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/2.14-3.0/)][2])
* This aligns the chart with Helm/Argo CD guidance for managing CRDs (and avoids unnecessary diff noise during syncs). ([[Argo Project](https://argoproj.github.io/argo-helm/)][3])

## Impact

* No functional change to the CRDs themselves—the field is ignored by modern API servers—but it **eliminates false OutOfSync** reports for users deploying this chart with [Argo CD](https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/2.14-3.0/). ([Argo CD][2])

## How I validated

* Reproduced OutOfSync on CRDs that included `spec.preserveUnknownFields: false`; removing the key restored a clean sync. (There’s also active community demand to drop the field.) ([[GitHub](https://github.com/argoproj/argo-helm/issues/3301)][4])

## Notes for Maintainers

I have **not completed the checklist** (version bump, docs, changelog, etc.). I’m short on time to go through the full contribution process, but this is a minimal, safe fix that prevents avoidable OutOfSync states for everyone using Argo CD. Please apply the change on your side and handle versioning/changelog as you see fit.

### DIFF
--- a/charts/argo-rollouts/templates/crds/analysis-run-crd.yaml
+++ b/charts/argo-rollouts/templates/crds/analysis-run-crd.yaml
@@ -23,7 +23,6 @@ spec:
     shortNames:
     - ar
     singular: analysisrun
-  preserveUnknownFields: false
   scope: Namespaced
   versions:
   - additionalPrinterColumns:
--- a/charts/argo-rollouts/templates/crds/analysis-template-crd.yaml
+++ b/charts/argo-rollouts/templates/crds/analysis-template-crd.yaml
@@ -23,7 +23,6 @@ spec:
     shortNames:
     - at
     singular: analysistemplate
-  preserveUnknownFields: false
   scope: Namespaced
   versions:
   - additionalPrinterColumns:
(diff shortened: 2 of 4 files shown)


yu-croco:
Hi, thank you for your PR. We will follow the upstream when upstream releases new version.

yu-croco:
Since this PR doesn't satisfy the items in [PR template](https://github.com/argoproj/argo-helm/blob/main/.github/pull_request_template.md), let us turn it into draft. Please handle them.

PR #3517, opened 2025-10-04
feat(argo-cd): Add HTTPRoute support for Gateway API

lexfrei:
## Summary

This PR adds support for Kubernetes Gateway API HTTPRoute resource to the Argo CD Helm chart, enabling users to expose Argo CD server using Gateway API instead of traditional Ingress or OpenShift Route.

## Changes

- **New template**: `argocd-server/httproute.yaml` - HTTPRoute resource template following Gateway API v1 specification
- **Configuration**: Added `server.httproute` section in `values.yaml` with support for:
  - `enabled` flag (default: false)
  - `labels` and `annotations` for custom metadata
  - `parentRefs` for Gateway binding (required)
  - `hostname` (defaults to `global.domain`)
  - `path` and `pathType` (default: `/` with `PathPrefix`)
  - `extraHosts` and `extraRules` for advanced routing scenarios
- **Smart port selection**: Automatically selects HTTP or HTTPS port based on `server.insecure` configuration
- **Documentation**: Updated README.md via helm-docs

## Motivation

Gateway API is the successor to Ingress and is becoming the standard for traffic management in Kubernetes. This addition provides users with a modern, vendor-neutral way to expose Argo CD, aligning with the evolving Kubernetes ecosystem.

## Testing

- ✅ Helm lint passed
- ✅ Template rendering verified with sample values
- ✅ Documentation generated successfully
- ✅ Backwards compatible (disabled by default)

## Notes

**AI Disclosure**: This code was developed with assistance from Claude (Anthropic AI assistant) but has been carefully reviewed, tested, and validated by the PR author.

---

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.1.8
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 8.5.8
+version: 8.6.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump argo-cd to v3.1.8
+    - kind: added
+      description: Add HTTPRoute support for Gateway API
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -1120,6 +1120,15 @@ NAME: my-release
 | server.extraArgs | list | `[]` | Additional command line arguments to pass to Argo CD server |
 | server.extraContainers | list | `[]` | Additional containers to be added to the server pod |
 | server.hostNetwork | bool | `false` | Host Network for Server pods |
+| server.httproute.annotations | object | `{}` | Additional HTTPRoute annotations |
+| server.httproute.enabled | bool | `false` | Enable HTTPRoute resource for Argo CD server (Gateway API) |
+| server.httproute.extraHosts | list | `[]` (See [values.yaml]) | The list of additional hostnames to be covered by HTTPRoute |
+| server.httproute.extraRules | list | `[]` (See [values.yaml]) | Additional HTTPRoute rules |
+| server.httproute.hostname | string | `""` (defaults to global.domain) | Argo CD server hostname |
+| server.httproute.labels | object | `{}` | Additional HTTPRoute labels |
+| server.httproute.parentRefs | list | `[]` (See [values.yaml]) | Gateway API parentRefs for the HTTPRoute |
+| server.httproute.path | string | `"/"` | The path to Argo CD server |
+| server.httproute.pathType | string | `"PathPrefix"` | HTTPRoute path type. One of `Exact`, `PathPrefix` or `RegularExpression` |
 | server.image.imagePullPolicy | string | `""` (defaults to global.image.imagePullPolicy) | Image pull policy for the Argo CD server |
 | server.image.repository | string | `""` (defaults to global.image.repository) | Repository to use for the Argo CD server |
 | server.image.tag | string | `""` (defaults to global.image.tag) | Tag to use for the Argo CD server |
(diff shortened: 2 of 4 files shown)


mkilchhofer:
Hi @lexfrei 
Sorry for the late review. We had already multiple PRs related to Gateway API and our helm charts:
- https://github.com/argoproj/argo-helm/pull/2689
- https://github.com/argoproj/argo-helm/pull/2965

The thing is that IMO Gateway API is not yet very common and also it lacks the feature where Developers are still able to control TLS settings (those are currently handled on the Gateway resource which is in the hands of a cluster operator). ListenerSet's are still in the design phase:
- https://github.com/cert-manager/cert-manager/pull/7839

So especially for Argo CD the migration to Gateway API is not very straightforward since we use:
- regular HTTP (but with HTTPS backend). 
  _For this you have to configure `HTTPRoute` (v1) + `BackendTLSPolicy` (v1alpha3)_
- gRPC
  _I assume we need a `kind: GRPCRoute` as well?_

I'd like to hear other opinions from the maintainer colleagues in this repo like @yu-croco @mbevc1 @tico24 .

But IMHO the best would be to raise the topic first in the upstream project (https://github.com/argoproj/argo-cd) and see what needs to be done and especially what needs to be documented in the [Operator manual](https://argo-cd.readthedocs.io/en/stable/operator-manual/ingress/) . My feeling is that not every Gateway API controller will support our default configuration of Argo CD in a similar manner, so it might be that users of the _envoy-gateway_ controller needs additional custom resources like `Backends(.gateway.envoyproxy.io)`.

**Update:**
I mean adding a conditional resource like you did in this PR is not a big deal, but we should make it clear to our chart users, that this might not be the definitive solution and it is a kind of "experimental feature".

mbevc1:
I'm inclined to agree with @mkilchhofer . Even though it'd be nice to see Gateway API used, it still seems it might need some time to mature :smile:

lexfrei:
You’re absolutely right — I totally forgot that Argo CD also exposes gRPC, since I don’t use that part myself.
I agree about the Gateway API maturity and adoption — that’s exactly why I’m trying to help push HTTPRoute support where I can.
Got it, you’re absolutely right — I’ll take some time to rework this and come back to it.

mkilchhofer:
I'd be glad as well if we can offer Gateway API support. I recently searched for a Helm chart "standard" about apps using Gateway API and didn't found much. Used this GitHub query:
```
".Values"  "kind: HTTPRoute" path:**/*.yaml
```

At least I found out that Helm prepared the `helm create mychart` scaffolding command with a HTTPRoute resource:
- https://github.com/helm/helm/pull/12912

I first thought, that this will be only in Helm v4 but it seems that it made it into version 3.19.0:
```bash
$ helm version
version.BuildInfo{Version:"v3.19.0", GitCommit:"3d8990f0836691f0229297773f3524598f46bda6", GitTreeState:"clean", GoVersion:"go1.24.7"}

$ helm create my-httproute-app
Creating my-httproute-app

$ tree my-httproute-app/
my-httproute-app/
├── Chart.yaml
├── charts
├── templates
│   ├── _helpers.tpl
│   ├── deployment.yaml
│   ├── hpa.yaml
│   ├── httproute.yaml      # <---- see here
│   ├── ingress.yaml
│   ├── NOTES.txt
│   ├── service.yaml
│   ├── serviceaccount.yaml
│   └── tests
│       └── test-connection.yaml
└── values.yaml

4 directories, 11 files
```

So I think we are good to go and we can start iteratively (adding BackendTLSPolicy and GRPCRoute later on).

**But:** I'd add a note to the documentation that it is experimental:

```yaml
  httproute:
    # -- Enable HTTPRoute resource for Argo CD server (Gateway API).
    # ADD NOTE HERE
    enabled: false
```

Something that not all use cases are covered today and also that it is unclear what controllers currently are supporting our case. (might depend on what version of the Gateway API they might implement). Cilium's implementation for example does not support BackendTLSPolicy 😆 : 
- https://github.com/cilium/cilium/issues/31352

**Update:**
Oh and while thinking about what to do with this (nice btw) PR, i quickly migrated my homelab Argo CD over to Gateway API (implemented by the [Envoy Gateway](https://gateway.envoyproxy.io/) project):
- https://github.com/mkilchhofer/k8s-gitops-services/pull/905

But for my homelab I also don't use the gRPC stuff :)

lexfrei:
## Added in this update

- **GRPCRoute** (Gateway API v1) - handles gRPC traffic that Argo CD exposes
- **BackendTLSPolicy** (v1alpha3, experimental) - for HTTPS backend connections
- **Documentation** in README.md with configuration examples for all three resources

## Changes

- All templates now follow the Helm scaffolding pattern (matching `helm create` output)
- Added experimental disclaimers in values.yaml noting controller-specific limitations
- Updated Chart.yaml changelog to reflect full Gateway API support scope

## Testing

- `helm lint` passes
- `helm template` verified - all three resources render correctly
- Tested with various configurations

---

@mkilchhofer, you mentioned it would be best to raise this topic in upstream Argo CD and update the [Operator Manual](https://argo-cd.readthedocs.io/en/stable/operator-manual/ingress/).

Am I understanding correctly that the plan should be:

1. Create an issue/discussion in **argoproj/argo-cd** to discuss Gateway API support with upstream maintainers
2. Create a separate PR in **argoproj/argo-cd** adding Gateway API documentation to the Operator Manual
3. Continue iterating on this Helm chart PR based on feedback from both discussions

Or would you prefer a different approach? You mentioned "we can start iteratively" in your last comment, so I want to make sure I'm aligned with what makes sense for the project.

PR #3545, opened 2025-10-20
fix(argo-cd): Fix typo in changelog message for 9.0.0

tgdfool2:
Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [ ] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #3546, opened 2025-10-20
docs(argo-cd): Correct Changelog in README for v9.0.0

yu-croco:
Thanks to @tgdfool2 ( https://github.com/argoproj/argo-helm/pull/3545  ) , I realized the wrong doc in changelog. Thank you so much. In order to deliver faster, please forgive me to make new PR for it.

Checklist:

### DIFF
--- a/charts/argo-cd/README.md.gotmpl
+++ b/charts/argo-cd/README.md.gotmpl
@@ -397,7 +397,7 @@ For full list of changes please check ArtifactHub [changelog].
 Highlighted versions provide information about additional steps that should be performed by user when upgrading to newer version.
 
 ### 9.0.0
-We have removed all parameters under `.Values.configs.cm.params` in this release, with the exception of `create` and `annotations`.
+We have removed all parameters under `.Values.configs.params` in this release, with the exception of `create` and `annotations`.
 This is to ensure better alignment with the upstream project, as tracking changes to their default values within the Helm chart has become challenging.
 
 **Breaking change**
@@ -409,9 +409,8 @@ To restore the previous setting, you can override the argocd-cmd-params-cm Confi
 
 ```yaml
 configs:
-  cm:
-    params:
-      applicationsetcontroller.policy: 'sync'
+  params:
+    applicationsetcontroller.policy: 'sync'
 ```
 
 ### 8.0.0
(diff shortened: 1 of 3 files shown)


charts/argo-cd/README.md.gotmpl
```diff
@@ -409,9 +409,8 @@ To restore the previous setting, you can override the argocd-cmd-params-cm Confi
 
 ```yaml
 configs:
-  cm:
-    params:
-      applicationsetcontroller.policy: 'sync'
+  params:
```
tgdfool2, line 412: Are you sure this is really correct?

> We have removed all parameters under `.Values.configs.params`

So why changing it to `configs.params` here?

yu-croco: Hi @tgdfool2 
We removed it because upstream has default value and they sometimes change the default value. We cannot catch up all the changes, especially this is out of k8s manifest. But we keep providing the interface to override them, so users can use the parameters like older version.

Ref: 
* https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd#900
* https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L418-L420

tgdfool2: Now I see, thanks for clarifying

yu-croco: Maybe it was better to add context like below? 🤔 
```diff
We have removed all parameters under `.Values.configs.params` in this release, with the exception of `create` and `annotations`.
This is to ensure better alignment with the upstream project, as tracking changes to their default values within the Helm chart has become challenging.

+ **Though we removed the parameters from values.yaml in argo-helm, we keep providing the interface to override `.Values.configs.params`. **
```

nbarrientos:
Just wanted to add that today we upgraded our development instances to 9.0.1 and it wasn't obvious from the README that, even though the defaults set by the chart were gone, these defaults were still aligned to ArgoCD's default at the time of removing them. As this was not clear I ended up fishing out all the defaults in ArgoCD itself to compare them to the value that the chart was defauting to, to make sure that there were no implicit configuration changes after upgrading the chart.

PR #3511, opened 2025-09-29
feat(argocd-image-updater): Update dependency argoproj-labs/argocd-image-updater to v0.17.0

yu-croco:
* In addition to https://github.com/argoproj/argo-helm/pull/3510 , I add some upstream's changes.
* diff: https://github.com/argoproj-labs/argocd-image-updater/compare/v0.16.0...v0.17.0

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #3548, opened 2025-10-21
fix(argocd-image-updater): Change default value of `.Values.ingress.enabled` to `false`

yu-croco:
Ref: https://github.com/argoproj/argo-helm/pull/3511#issuecomment-3425553960
Ingress (and service for it) is supposed to be opt-in instead of opt-out. In order to align to other argoproj's projetcs (e.g. argo-cd), I changed the default value from `true` to `false` .

Checklist:

### DIFF
--- a/charts/argocd-image-updater/Chart.yaml
+++ b/charts/argocd-image-updater/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argocd-image-updater
 description: A Helm chart for Argo CD Image Updater, a tool to automatically update the container images of Kubernetes workloads which are managed by Argo CD
 type: application
-version: 0.13.0
+version: 0.14.0
 appVersion: v0.17.0
 home: https://github.com/argoproj-labs/argocd-image-updater
 icon: https://argocd-image-updater.readthedocs.io/en/stable/assets/logo.png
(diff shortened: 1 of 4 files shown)


charts/argocd-image-updater/templates/service.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.ingress.enabled }}
```
yu-croco, line 1: service was also implemented at the same time as ingress, so it doesn't any side effect. 📝

charts/argocd-image-updater/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argocd-image-updater
 description: A Helm chart for Argo CD Image Updater, a tool to automatically update the container images of Kubernetes workloads which are managed by Argo CD
 type: application
-version: 0.13.0
+version: 0.14.0
```
yu-croco, line 5: This release is related to `0.13.0` but it contains breaking change, so I think bumping (at least) minor version makes sense? 🤕

mkilchhofer (approved):
LGTM, thanks for fixing it 🙏

charts/argocd-image-updater/Chart.yaml
mkilchhofer: Since we are on 0.x with the chart version and also the AIU is still in active development:


> **A note on the current status**
> 
> Argo CD Image Updater is under active development.
> 
> You are welcome to test it out on non-critical environments, and of course to contribute by many means.
> 
> There will be (probably a lot of) breaking changes from release to release as development progresses until version 1.0. We will do our best to indicate any > breaking change and how to un-break it in the respective release notes.
> 
> Argo CD Image Updater is not currently compatible with Argo CD Applications in any namespace.
> 
> _Source: https://argocd-image-updater.readthedocs.io/en/stable/_

-> I'd continue without incrementing the major chart version number. :) 👍

jmeridth (approved):

jmeridth:
@mkilchhofer and you discussed the version bump. I concur with just the minor bump.

PR #3559, opened 2025-10-31
feat(argo-workflows): Add support for full CRDs

Joibel:
This change adds the capability to install full CRDs (with complete OpenAPI schemas) instead of the minified ones. This is controlled by a new configuration option crds.full which defaults to false (minified CRDs).

Changes:
- Add new crds.full configuration option in values.yaml (defaults to `false`)
- Create templates/crds-full/ directory with full CRDs from upstream v3.7.3
- Move existing minified CRDs to templates/crds-minified/
- Add conditional logic to install either full or minified CRDs based on config
- Update README.md with documentation about the new option
- Update CONTRIBUTING.md with namespace creation
- Update default namespace documentation for argo-workflows
- Use `/usr/bin/env bash` for better compatibility in scripts

Possibly in 4.0 we (the workflows team) will suggest switching this to default `true`.

If you don't like my "minor fixes along the way" in this PR, I'm happy to split it.

3 of the 8 CRDs are already "full" so we have duplicates of them in the current format. We could put them in a common place if that is preferred. I'm new to this repo and your preferred practices.

Checklist:

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -35,7 +35,7 @@ jobs:
           version: v3.10.1 # Also update in publish.yaml
 
       - name: Set up python
-        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
         with:
           python-version: 3.9
 
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -68,6 +68,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@e12f0178983d466f2f6028f5cc7a6d786fd97f4b # v3.29.5
+        uses: github/codeql-action/upload-sarif@fdbfb4d2750291e159f0156def62b853c2798ca2 # v3.29.5
         with:
           sarif_file: results.sarif
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: v3.2.0
+appVersion: v3.2.1
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 9.1.4
+version: 9.1.5
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
-      description: Bump redis_exporter to v1.80.1
+      description: Bump argo-cd to v3.2.1
--- a/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
@@ -324,6 +324,18 @@ spec:
                 key: reposerver.revision.cache.lock.timeout
                 name: argocd-cmd-params-cm
                 optional: true
+          - name: ARGOCD_REPO_SERVER_ENABLE_BUILTIN_GIT_CONFIG
+            valueFrom:
+              configMapKeyRef:
+                name: argocd-cmd-params-cm
+                key: reposerver.enable.builtin.git.config
+                optional: true
+          - name: ARGOCD_GRPC_MAX_SIZE_MB
+            valueFrom:
+              configMapKeyRef:
+                key: reposerver.grpc.max.size
+                name: argocd-cmd-params-cm
+                optional: true
           - name: ARGOCD_REPO_SERVER_INCLUDE_HIDDEN_DIRECTORIES
             valueFrom:
               configMapKeyRef:


vladlosev:
I would like to sound a note of caution here. Full Argo Workflows CRDs are very large and will slow down the processing substantially. It my work we use A GitOps automated sync deployments, and when we tried to use full CRDs, the process slowed down enough to start timing out. This change does not default to full CRDs,  but there are still costs being added
- The download size increases significantly. That might affect both bandwidth consumed and time spent on downloading the charts. Hopefully, most CD systems cache these downloads, though.
- These CRDs are implemented as direct Go templates. This means Helm will have to read and parse process the entire CRD template just in order to generate an empty output from it. It's quite a bit of CPU and memory consumption. If one is running their CD solution in a container with limited memory requirements, this change is likely to exceed them and cause OOMs.

The former problem is annoying, but the latter can be a real breaking change for some. There are ways to rectify these issues. For example, the actual CRDs can be kept in flat files, and the CRD template could be a short stub loading them via Helm's `File` feature, like this:
```
{{- if .Values.crds.install }}
{{- $pattern := ternary "files/crds/full/*.yaml" "files/crds/minimal/*.yaml" .Values.crds.full }}
{{- range $path, $_ :=  .Files.Glob  $pattern }}
---
{{ $.Files.Get $path }}
{{- end }}
{{- end }}
```
This approach avoids the overhead of processing large CRD templates when they are not required. Incidentally, it will also let us copy the CRDs files verbatim from the Argo Workflows project, reducing maintenance burden of turning them into templates.

Joibel:
@vladlosev, lets hold on merging this, and I'll do that (next week now), it's a very neat pattern.

Thank you.

Joibel:
It doesn't seem it's possible to modify files from Files.Get, which makes adding the annotations impossible - is there a way around that? I've pushed up a new version with something like the modifications suggested, but it's broken CI and it's broken `crds.keep` and `crds.annotations`.

I don't have a way to fix this, but I'm no helm expert - asking for assistance with that.

charts/argo-workflows/templates/crds.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.crds.install }}
+{{- $pattern := ternary "files/crds/full/*.yaml" "files/crds/minimal/*.yaml" .Values.crds.full }}
+{{- range $path, $_ :=  .Files.Glob  $pattern }}
+{{- if not (contains "clusterworkflowtemplates" $path) }}
+---
+{{ $.Files.Get $path }}
```
vladlosev: You may be able do something like this here:
```
{{ regexReplaceAllLiteral "\n  annotations:\n" ($.Files.Get $path) "\n    annotations:\n  helm.sh/resource-policy: keep\n" }}
```
May even abstract it into a function call, e.g.,
```
{{ include "argo-workflows.insertResourcePolicyAnnotation" (dict "input" ($.Files.Get $path) "value" "keep" }
```
The function may look like this:
```
{{- define "argo-workflows.insertResourcePolicyAnnotation" -}}
{{ regexReplaceAllLiteral "\n  annotations:\n" .input (format "\n    annotations:\n  helm.sh/resource-policy: %s\n" .value) }}
{{- end -}}
```

charts/argo-workflows/templates/crds.yaml
```diff
@@ -0,0 +1,14 @@
...
+{{- end }}
+{{- end }}
+{{- if or (.Values.server.clusterWorkflowTemplates.enabled) (.Values.controller.clusterWorkflowTemplates.enabled) }}
```
vladlosev: I would suggest putting this condition into line 4 so that you don't have to complicate the code with two loops:
```
{{- if or (contains "clusterworkflowtemplates" $path | not) .Values.server.clusterWorkflowTemplates.enabled .Values.controller.clusterWorkflowTemplates.enabled }}
```

Joibel: Done

charts/argo-workflows/templates/crds.yaml
Joibel: I have tried to implement this

Joibel:
@vladlosev I have rebased, but the actual changes requested are in https://github.com/argoproj/argo-helm/pull/3559/commits/d34f459d5450d1ebb10540c0806bbbabde0cfeba

charts/argo-workflows/templates/crds.yaml
Joibel: It's a neat trick, I hadn't realised this was possible - thank you.

charts/argo-workflows/README.md
```diff
@@ -2,23 +2,79 @@
...
+
 Some users would prefer to install the CRDs _outside_ of the chart. You can disable the CRD installation of this chart by using `--set crds.install=false` when installing the chart.
 
 Helm cannot upgrade custom resource definitions in the `<chart>/crds` folder [by design](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations). Starting with 3.4.0 (chart version 0.19.0), the CRDs have been moved to `<chart>/templates` to address this design decision.
 
-If you are using Argo Workflows chart version prior to 3.4.0 (chart version 0.19.0) or have elected to manage the Argo Workflows CRDs outside of the chart, please use `kubectl` to upgrade CRDs manually from [templates/crds](templates/crds/) folder or via the manifests from the upstream project repo:
+If you are using Argo Workflows chart version prior to 3.4.0 (chart version 0.19.0) or have elected to manage the Argo Workflows CRDs outside of the chart, please use `kubectl` to upgrade CRDs manually from [templates/crds-full](templates/crds-full/) or [templates/crds-minified](templates/crds-minified/) folders, or via the manifests from the upstream project repo:
```
vladlosev: I believe it should point to `files/crds/full` and `files/crds/minimal`. Otherwise people may get confuse and try to apply those templates directly.

charts/argo-workflows/templates/_helpers.tpl
```diff
@@ -209,3 +209,10 @@ Allows overriding it for multi-namespace deployments in combined charts.
 {{- define "argo-workflows.namespace" -}}
 {{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
 {{- end }}
+
+{{/*
+Insert helm/resource-policy annotations into the workflows CRDs
+*/}}
+{{- define "argo-workflows.insertResourcePolicyAnnotation" -}}
+{{- regexReplaceAllLiteral "\n  annotations:\n    helm.sh/resource-policy: keep\n" .input (ternary "\n  annotations:\n    helm.sh/resource-policy: keep\n" "\n  annotations: {}\n" .value) -}}
+{{- end -}}
```
vladlosev: Please add a terminating newline.

vladlosev (commented):
The new fill CRDs seem to lack the `metadata.name` fields.

charts/argo-workflows/templates/_helpers.tpl
Joibel: Done

charts/argo-workflows/README.md
Joibel: Yes, good catch. Done

Joibel:
> Other issues have been fixed, but the new full CRDs still lack the `metadata.name` fields. They are not valid without it.

Apologies, missed this. Done now

Joibel:
Thank you all

Joibel:
I successfully deployed this via ArgoCD with an application with SSA enabled. `application-controller` needed more memory limit to work, but it did the update. (Kube-prom-stack would deploy with the old memory limit, so this was a bit surprising)

PR #3565, opened 2025-11-03
feat(argo-rollouts): Allow templating in notification templates

cktpw:
Checklist:

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Setup Chart Linting
         id: lint
-        uses: helm/chart-testing-action@0d28d3144d3a25ea2cc349d6e59901c4ff469b3b # v2.7.0
+        uses: helm/chart-testing-action@6ec842c01de15ebb84c8627d2744a0c2f2755c9f # v2.8.0
         with:
           # Note: Also update in scripts/lint.sh
           version: v3.11.0
@@ -70,7 +70,7 @@ jobs:
           fi
 
       - name: Create kind cluster
-        uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+        uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
         if: steps.list-changed.outputs.changed == 'true'
         with:
           config: .github/configs/kind-config.yaml
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@ea850436a5fe75c0925d583c7a02c60a5865461d # v43.0.20
+        uses: renovatebot/github-action@4ebebabcd582dddea1692b05c3d5279f8e372b53 # v44.0.0
         with:
           configurationFile: .github/configs/renovate-config.js
           # renovate: datasource=docker depName=ghcr.io/renovatebot/renovate
--- a/charts/argo-cd/Chart.lock
+++ b/charts/argo-cd/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: redis-ha
   repository: https://dandydeveloper.github.io/charts/
-  version: 4.33.7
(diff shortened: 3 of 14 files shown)


charts/argo-rollouts/templates/controller/notifications-configmap.yaml
```diff
@@ -9,13 +9,25 @@ metadata:
     {{- include "argo-rollouts.labels" . | nindent 4 }}
 data:
   {{- with .Values.notifications.notifiers }}
-    {{- toYaml . | nindent 2 }}
+    {{- if eq .format "template" }}
+      {{- tpl (toYaml .) $ | nindent 2 }}
+    {{- else }}
+      {{- toYaml . | nindent 2 }}
+    {{- end }}
```
yu-croco, line 16: Any problem if we just support `tpl` ? As you can see there is no format parameters in all of argo-helm, though we have some logics to support `tpl`.
```yaml
{{- with .Values.notifications.notifiers }}
  {{- tpl (toYaml .) $ | nindent 2 }}
{{- end }}
```

cktpw: Both Argo notifications and Helm use `{{` and `}}` for interpolation. So if someone wants to use `{{.rollout.metadata.namespace}}` with `tpl` they'll have to escape the `{{` and `}}` with like this: `{{"{{"}}.rollout.metadata.namespace{{"}}"}}`.

Using `tpl` alone would throw errors if someone just do `{{.rollout.metadata.namespace}}` because that doesn't exist in Helm's context.

The reason I had to do this is because rollouts doesn't support the `context.rolloutsUrl` like ArgoCD does with `context.argocdUrl` and I need to provide different Rollout's root URL per Rollout's deployment.

Here's how I did it with the enhancement above. With this method, I can provide a different value for `notifications.rolloutsUrl` for each deployment of Rollouts.

```yaml
  notifications:
    secret:
      create: false

    rolloutsUrl: https://rollouts.staging.example.com

    notifiers:
      service.slack: |
        token: $slack-token

    triggers:
      trigger.on-rollout-updated: |
        - send: [rollout-updated]

    templates:
      format: template
      template.rollout-updated: |
        slack:
          attachments: |
            [{
              "title": "Rollout for {{"{{"}}.rollout.metadata.name{{"}}"}} in progress",
              "title_link": "{{.Values.notifications.rolloutsUrl}}/rollouts/rollout/{{"{{"}}.rollout.metadata.namespace{{"}}"}}/{{"{{"}}.rollout.metadata.name{{"}}"}}",
              "color": "#ffe000",
              "fields": [],
              "mrkdwn_in": ["fields"]
            }]
```

yu-croco: Thank you for your explanation.
This gives a breaking change, which means bumping patch version is not enough, so I am wondering if we can handle it well....  🤔  We want to avoid it to existing users.

cktpw: Where do you see the breaking change? It's defaulted to NOT use `tpl` if `format` is not defined or not equal to `template`. So, this should not break for existing users.

yu-croco: I thought it gave users affect because values.yaml had new attribute and it said that users need to choose from plain or template.
In the aspect of internal logic, it is my misunderstanding.

values.yaml looks requiring  “format” and even not adding the attribute works, so I wonder it's a little confusing?

cktpw: What do you suggest? I can add a note in the variable's comment to say that it's optional? But like most Helm values are default and users only have to define it if they want to override their values. However, currently it shouldn't break anything for existing users.

yu-croco: Ok let's go with current one.

yu-croco:
Can you please handle below? You checked them but they are not handled yet.
> * I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
> * I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).

cktpw:
> Can you please handle below? You checked them but they are not handled yet.
> 
> > * I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
> > * I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).

I'll do them. Sorry I don't do PR to to community repos often.

PR #3585, opened 2025-11-13
fix(argo-cd): make gateway API backend protocol selection work for server

yann-soubeyrand:
When using gateway API (https://gateway-api.sigs.k8s.io/) to configure a Gateway and an HTTPRoute to access Argo CD server, backend protocol selection (https://gateway-api.sigs.k8s.io/guides/backend-protocol/) relies on the appProtocol field of the backend service port (https://gateway-api.sigs.k8s.io/reference/spec/#backendref). Valid values for this field are described in KEP-3726
(https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols).

Since Argo CD server serves web UI using HTTP 1 and GRPC using HTTP 2, both on the same port, two ports redirecting to the same pod port are needed on the service.

Checklist:

### DIFF
--- a/charts/argo-cd/templates/argocd-server/service.yaml
+++ b/charts/argo-cd/templates/argocd-server/service.yaml
@@ -44,6 +44,14 @@ spec:
     {{- if eq .Values.server.service.type "NodePort" }}
     nodePort: {{ .Values.server.service.nodePortHttp }}
     {{- end }}
+  - name: {{ .Values.server.service.servicePortHttp2Name }}
+    protocol: TCP
+    appProtocol: {{ .Values.server.service.servicePortHttp2AppProtocol}}
+    port: {{ .Values.server.service.servicePortHttp2 }}
+    targetPort: {{ .Values.server.containerPorts.server }}
+    {{- if eq .Values.server.service.type "NodePort" }}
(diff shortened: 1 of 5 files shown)


mkilchhofer:
Not stale

charts/argo-cd/templates/argocd-server/service.yaml
```diff
@@ -44,6 +44,14 @@ spec:
...
     nodePort: {{ .Values.server.service.nodePortHttp }}
     {{- end }}
+  - name: {{ .Values.server.service.servicePortHttp2Name }}
+    protocol: TCP
+    appProtocol: {{ .Values.server.service.servicePortHttp2AppProtocol}}
+    port: {{ .Values.server.service.servicePortHttp2 }}
+    targetPort: {{ .Values.server.containerPorts.server }}
+    {{- if eq .Values.server.service.type "NodePort" }}
+    nodePort: {{ .Values.server.service.nodePortHttp2 }}
+    {{- end }}
```
gaffneyd4, line 54: What do you think about toggling between the h2c port AND http port OR https port based on the value of $insecure, since h2c and http should only be needed if the server isn't listening with https. I suppose this would be a more major change though than simply adding a port for h2c.

PR #3629, opened 2025-12-09
fix(argocd-image-updater): missing and typo on configmap name interpolation

### LINKED ISSUE #3628: ArgoCD Image Updater deployment template hardcodes configmap name for interval
https://github.com/argoproj/argo-helm/blob/c14f2b82fb8a4ff39e10b2f89e0d3a11de48b753/charts/argocd-image-updater/templates/deployment.yaml#L46-L51

Unlike the other config values, the `interval`, `webhook.enable`, `webhook.port`, and `webhook.ratelimit-allowed` configuration keys seem to be hardcoded to a specific configmap name (the default `argocd-image-updater-config`. This prevents these values from being defined if you want to customize the configmap name by setting a value for `config.name`. 

Additionally, there is a typo in the valueFrom statement for the `max_concurrent_reconciles` key. `name: a{{ .Values.config.name }}` will set an invalid configmap and cause issues.

I know that you can set the `interval` as an extraArg which takes precedence over the configmap (which provides a workaround for some of this) but that typo cannot be ignored since there is no way to provide a valid value for the configMapKeyRef.name for `max_concurrent_reconciles`.

I have linked a PR that should address this.

schlags:
Fixes #3628 

A couple of the places in the `valueFrom` env definitions were using the default configmap name, and `MAX_CONCURRENT_RECONCILES` had an extra `a` before the `{{ Values` section. This addresses all of those situations so that the configmap name can actually be customized, and MAX_CONCURRENT_RECONCILES can now be used.

Checklist:

	- _NOTE: Seems wrong to try and be backwards compatible with the typo..._

### DIFF
--- a/charts/argocd-image-updater/Chart.yaml
+++ b/charts/argocd-image-updater/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argocd-image-updater
 description: A Helm chart for Argo CD Image Updater, a tool to automatically update the container images of Kubernetes workloads which are managed by Argo CD
 type: application
-version: 1.0.1
+version: 1.0.2
 appVersion: v1.0.1
 home: https://github.com/argoproj-labs/argocd-image-updater
 icon: https://argocd-image-updater.readthedocs.io/en/stable/assets/logo.png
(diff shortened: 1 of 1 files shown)


mkilchhofer (approved):
Thanks for fixing this, @schlags 🙏

yu-croco (approved):

PR #3641, opened 2025-12-21
feat(argo-cd): add minimal RBAC option for server ClusterRole

### LINKED ISSUE #3594: Feature Request: Option to Disable Patch/Delete Permissions for Argo CD Server in Read-Only Mode

Yes. When using Argo CD in read-only mode, the default Helm chart still assigns broad patch/delete permissions to the argocd-server service account. These permissions are unnecessary for our use case and fail internal security/RBAC reviews.

### Related helm chart

argo-cd

I would like a Helm chart option (e.g., server.rbac.minimal: true) that generates a reduced-privilege ClusterRole for argocd-server, removing wildcard patch and delete permissions when Argo CD is running in read-only mode.

The only alternative is to manually override or recreate the argocd-server ClusterRole with reduced permissions, but this is not ideal because it requires maintaining custom RBAC outside the Helm chart and can break during upgrades.

### Additional context

_No response_

Maharshi-Mimo:
Add a new server.rbac.minimal configuration option that restricts argocd-server ClusterRole permissions to a minimal read-only set, removing delete and patch operations on wildcard resources.

Changes:
- Add server.rbac.minimal boolean flag (default: false) to values.yaml
- Update clusterrole.yaml template with conditional logic:
  * When minimal=false (default): uses existing permissions with delete/patch
  * When minimal=true: grants only get, list, watch verbs on wildcard resources
  * Custom rules (when enabled) take precedence over both modes
- Minimal mode removes the following dangerous permissions:
  - delete/* on wildcard resources
  - patch/* on wildcard resources
  - create on events (except pods/exec)
  - create on batch jobs
  - update/patch on applications/applicationsets

Enable with: --set server.rbac.minimal=true

This addresses security concerns by allowing operators to restrict the ArgoCD server's permissions to read-only access when appropriate.
closes #3594 

Tested: Helm template rendering verified with both modes enabled and disabled.

Checklist:

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -28,7 +28,7 @@ jobs:
           persist-credentials: false
 
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@822441559e94f98b67b82d97ab89fe3003b0a247 # v44.2.0
+        uses: renovatebot/github-action@f7fad228a053c69a98e24f8e4f6cf40db8f61e08 # v44.2.1
         with:
           configurationFile: .github/configs/renovate-config.js
           # renovate: datasource=docker depName=ghcr.io/renovatebot/renovate


charts/argo-cd/values.yaml
```diff
@@ -2695,6 +2695,11 @@ server:
     # -- List of custom rules for the server's ClusterRole resource
     rules: []
 
+  ## Server RBAC configuration
+  rbac:
+    # -- Restrict server permissions to minimal set (removes delete/patch on wildcard resources)
+    minimal: false
```
ebuildy, line 2701: minimal is not really explicit, what about "readOnly"?


```suggestion
    readOnly: false
```

mkilchhofer:
Hi @Maharshi-Mimo , Thanks for this PR and your contribution.

I don't think that this makes sense. We provide an option for users who want to harden the Argo CD deployment by using:


```yaml
## Option 1: Cluster Role rules overrides 

server:
  clusterRoleRules:
    rules:
      - # your custom RBAC rules, depending on your use case

controller:
  clusterRoleRules:
    rules:
      - # your custom RBAC rules, depending on your use case
```
or:

```yaml
## Option 2: Role rules overrides (for namespaced installations)

controller:
  roleRules:
    - # your custom RBAC rules, depending on your use case
```

There is no generic "minimal" setup for everyone with hardening intents in my opinion since the access depends on what the user of our Helm chart wants to manage with Argo CD.

Does that make sense?

---

_**Another thing I want to bring up:**_ you clicked every checkbox of the contribution template and clearly not follow what you checked. E.g.
> - I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
> - My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Since you made non-functional changes across all Helm charts, you involved all maintainers for all charts. For the next time, please follow the task checkboxes since our review capacity is limited. 🙏

PR #3676, opened 2026-01-20
feat(argo-cd): Add support to configure Azure workload identity

jagpreetstamber:
Closes: #3679

Checklist:

### DIFF
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -851,8 +851,9 @@ NAME: my-release
 |-----|------|---------|-------------|
 | apiVersionOverrides | object | `{}` |  |
 | azure.workloadIdentity.clientId | string | `nil` | Client Id of the workload identity to be used to pull artifacts |
-| azure.workloadIdentity.enable | bool | `false` | Enable\Disable Azure workload identity |
+| azure.workloadIdentity.enabled | bool | `false` | Enable\Disable Azure workload identity |
 | azure.workloadIdentity.entraSSOClientId | string | `nil` | Client Id of the workload identity to be used to Entra SSO, can be same as clientId |
+| azure.workloadIdentity.tenantId | string | `nil` | Tenant Id of the workload identity. |
 | crds.additionalLabels | object | `{}` | Additional labels to be added to all CRDs |
 | crds.annotations | object | `{}` | Annotations to be added to all CRDs |
 | crds.install | bool | `true` | Install and upgrade CRDs |
--- a/charts/argo-cd/templates/argocd-server/serviceaccount.yaml
+++ b/charts/argo-cd/templates/argocd-server/serviceaccount.yaml
@@ -5,14 +5,14 @@ automountServiceAccountToken: {{ .Values.server.serviceAccount.automountServiceA
 metadata:
   name: {{ include "argo-cd.server.serviceAccountName" . }}
   namespace: {{ include  "argo-cd.namespace" . }}
-  {{- if or .Values.server.serviceAccount.annotations .Values.azure.workloadIdentity.enable }}
+  {{- if or .Values.server.serviceAccount.annotations .Values.azure.workloadIdentity.enabled }}
   annotations:
     {{- with .Values.server.serviceAccount.annotations }}
     {{- range $key, $value := . }}
     {{ $key }}: {{ $value | quote }}
     {{- end }}
     {{- end }}
-    {{- if .Values.azure.workloadIdentity.enable }}
+    {{- if .Values.azure.workloadIdentity.enabled }}
     azure.workload.identity/client-id: {{ .Values.azure.workloadIdentity.entraSSOClientId | quote }}
     {{- if .Values.azure.workloadIdentity.tenantId }}
     azure.workload.identity/tenant-id: {{ .Values.azure.workloadIdentity.tenantId | quote }}
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -4290,7 +4290,13 @@ commitServer:
 
 azure:
   workloadIdentity:
-    enable: false
+    # -- Enable\Disable Azure workload identity
+    # @default -- `false`
+    enabled: false
+    # -- Client Id of the workload identity to be used to pull artifacts
     clientId:
+    # --  Client Id of the workload identity to be used to Entra SSO, can be same as clientId
     entraSSOClientId:
+    # --  Tenant Id of the workload identity.
+    tenantId :
 
--- a/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-repo-server/deployment.yaml
@@ -49,7 +49,7 @@ spec:
         {{- with (mergeOverwrite (deepCopy .Values.global.podLabels) .Values.repoServer.podLabels) }}
           {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if .Values.azure.workloadIdentity.enable }}
+        {{- if .Values.azure.workloadIdentity.enabled }}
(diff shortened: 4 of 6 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -28,3 +28,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: changed
```
tico24: Remove this please

tico24:
Your PR title is also not valid. Thanks

charts/argo-cd/README.md
```diff
@@ -232,35 +233,35 @@ server:
           port: 8080
       frontendConfig:
         redirectToHttps:
-          enabled: true 
+          enabled: true
```
tico24: This blank space isn't needed

charts/argo-cd/README.md
```diff
@@ -232,35 +233,35 @@ server:
...
-    annotations: 
-      # Optional: Add any AKS-specific annotations if needed 
-    extraTls: 
-      - hosts: 
-          - argocd.example.com 
-        # Certificate can be managed by Web Application Routing 
+AKS Web Application Routing uses the standard Kubernetes Ingress specification, so the `generic` controller type is sufficient.  
+webapprouting provides a managed ingress controller based on nginx.  
+
+```yaml
```
tico24: None of these blank spaces are necessary

charts/argo-cd/Chart.yaml
jagpreetstamber: removed

charts/argo-cd/values.yaml
```diff
@@ -4287,3 +4287,10 @@ commitServer:
     # -- Default network policy rules used by commit server
     # @default -- `false` (defaults to global.networkPolicy.create)
     create: false
+
+azure:
+  workloadIdentity:
+    enable: false
```
tico24: Let's stick with `enabled` as is already used in the values.yaml

charts/argo-cd/templates/argocd-server/serviceaccount.yaml
```diff
@@ -5,11 +5,19 @@ automountServiceAccountToken: {{ .Values.server.serviceAccount.automountServiceA
...
   annotations:
+    {{- with .Values.server.serviceAccount.annotations }}
     {{- range $key, $value := . }}
     {{ $key }}: {{ $value | quote }}
     {{- end }}
+    {{- end }}
+    {{- if .Values.azure.workloadIdentity.enable }}
+    azure.workload.identity/client-id: {{ .Values.azure.workloadIdentity.entraSSOClientId | quote }}
+    {{- if .Values.azure.workloadIdentity.tenantId }}
+    azure.workload.identity/tenant-id: {{ .Values.azure.workloadIdentity.tenantId | quote }}
```
tico24, line 18: This doesn't appear to be a thing in the values.yaml

charts/argo-cd/README.md
```diff
@@ -850,6 +850,9 @@ NAME: my-release
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | apiVersionOverrides | object | `{}` |  |
+| azure.workloadIdentity.clientId | string | `nil` | Client Id of the workload identity to be used to pull artifacts |
```
tico24, line 853: This wasn't generated by helm-docs.sh. Your values file has no comments in it.

tico24 (changes requested):
In addition, can you please reference the issue that this closes? Thanks

mkilchhofer:
My question is:
Is it possible to configure it already today?
(Maybe with not so obvious settings)

If so, I'd preffer to keep the chart agnostic as possible. Instead we can create some doc notes in the README or so.

charts/argo-cd/README.md
jagpreetstamber: fixed

charts/argo-cd/templates/argocd-server/serviceaccount.yaml
jagpreetstamber: added

charts/argo-cd/values.yaml
jagpreetstamber: updated

jagpreetstamber:
> My question is: Is it possible to configure it already today? (Maybe with not so obvious settings)
> 
> If so, I'd preffer to keep the chart agnostic as possible. Instead we can create some doc notes in the README or so.

Yes it is possible, but customer have a difficult time to configure it. hence adding it as a native option for easier configurations.

jagpreetstamber:
> In addition, can you please reference the issue that this closes? Thanks

Added the issue in the description

KiPIDesTAN:
I agree with @mkilchhofer. I'm not a maintainer - just someone who uses the argo-cd chart and was looking for guidance on using workload identity for git repos. The comments below are also not to take away from the effort put into the PR.

My personal opinion is that workload identity would be much better supported by documentation on specific outcomes, like "how do I configure the helm chart to use workload identity to access my ADO repo or ACR?" or "how do I configure the helm chart to leverage Entra authentication for user management?" Those are the tough challenges customers have because getting the workload identity token via the Helm chart configs for argo to use is the tricky part.

This PR doesn't supply much value because it only enables workload identity - a straight-forward process that is well documented. The code doesn't show how to use it.

The code also makes an incorrect assumption.

Adding the workload identity annotations and labels to argo server is to support [user authentication with Entra](https://argo-cd.readthedocs.io/en/release-3.0/operator-manual/user-management/microsoft/#using-workload-identity-federation-recommended). Adding the workload identity annotations and labels to the argo repo server supports using [Azure DevOps repos or Azure Container Registries within ArgoCD](https://argo-cd.readthedocs.io/en/release-3.0/user-guide/private-repositories/#azure-container-registryazure-repos-using-azure-workload-identity).

These two capabilities are very different, but the PR assumes the user wants neither or both by sourcing the configuration from the same values. It's fairly common for end users to use Entra authentication for user management and not use Azure DevOps for their repos. 

The PR change is the equivalent of having the values.yaml file below.

```yaml
server:
  podLabels:
    azure.workload.identity/use: "true"
  serviceAccount:
    annotations:
      azure.workload.identity/client-id: "<client-id>"
      azure.workload.identity/tenant-id: "<tenant-id>"
repoServer:
  podLabels:
    azure.workload.identity/use: "true"
  serviceAccount:
    annotations:
      azure.workload.identity/client-id: "<client-id>"
      azure.workload.identity/tenant-id: "<tenant-id>"
```

mkilchhofer:
@KiPIDesTAN 👍 

Especially since there is official documentation to add labels and annotations for exactly this workload identity:
https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/microsoft/#add-credentials-a-new-entra-id-app-registration

I think users who follow this, should be able to "translate" the instructions to our helm chart values/parameters.

jagpreetstamber:
> @KiPIDesTAN 👍
> 
> Especially since there is official documentation to add labels and annotations for exactly this workload identity: https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/microsoft/#add-credentials-a-new-entra-id-app-registration
> 
> I think users who follow this, should be able to "translate" the instructions to our helm chart values/parameters.

we have had multiple customers requesting the easier configurations which resulted in this PR. I am good with the documentation.

@mkilchhofer Shall I update the README with this guidance as existing argocd workload identity docs are not helping much.

jmeridth:
@jagpreetstamber as @mkilchhofer stated, we're going to currently keep the chart agnostic of specific workload identity. I'm going to close this PR. We thank you for your effort but won't be proceeding with it.

If you believe the upstream argo-cd workload identity documentation is "not helping much", may I suggest a PR to those docs to make them help more?

Cheers.

PR #3734, opened 2026-02-17
feat(argo-cd): Allow configuring external redis endpoint from secret

### LINKED ISSUE #3735: argo-cd: Allow External Redis endpoint from secret

We are using Argo CD with AWS Elasticache, our Elasticache is provisioned via crossplane alongside the chart.

To be able to correctly point Argo we need to do two installs, or have Elasticache created separately because of the endpoint.
With this change we allow the redis endpoint when using externalRedis to be passed from the secret, it will be up to the user to ensure it is in host:port format

### Related helm chart

argo-cd

Being able to configure argo-cd to read redis endpoint from a secret.

_No response_

### Additional context

_No response_

jdcmarques:
Description:
We are using Argo CD with AWS Elasticache, our Elasticache is provisioned via crossplane alongside the chart.
To be able to correctly point Argo we need to do two installs, or have Elasticache created separately because of the endpoint.
With this change we allow the redis endpoint when using `externalRedis` to be passed from the secret, it will be up to the user to ensure it is in `host:port`  format

Fixes #3735

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.3.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 9.4.5
+version: 9.4.6
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: added
-      description: Add support for configuring HTTPRoute timeouts
+      description: Allow configuring external redis endpoint from secret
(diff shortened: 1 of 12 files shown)


mkilchhofer:
Sorry for the late reply. Maybe it would make sense to refactor the `externalRedis.existingSecret`-feature completely with a structure like this:

```yaml
externalRedis:

  # ..

  existingSecret:
    name: "asdf"
    serverKey: "xyz"
    usernameKey: "abc"
    passwordKey: "foobar"
```

Hopefully this would solve other feature requests more easily like this one over there:
- https://github.com/argoproj/argo-helm/pull/3849

For sure we'd need to bump the major version of the chart, but that's fully okay for me.

What do you think, @jdcmarques ?

jmeridth:
@jdcmarques Checking in on this one. The feature itself is still wanted (#3735 remains open) and, unlike some other externalRedis asks, it isn't blocked on upstream since Argo CD already honors the `REDIS_SERVER` env var. The open question is the API shape: per @mkilchhofer's comment above, we'd want this as a restructured `externalRedis.existingSecret` object (`name` / `serverKey` / `usernameKey` / `passwordKey`) rather than a flat `existingSecretServerKey` sibling, since that structure also covers related requests and avoids shipping a value we'd deprecate at the next major. That rework would come with a major chart version bump.

Are you up for reworking the PR in that direction? It would also need a rebase, the branch currently conflicts with main. Two review notes to fold in if you do: the `secretKeyRef` swap silently disables the `argocd-cmd-params-cm` `redis.server` override, so the precedence between the two should be documented; and consider whether `optional: false` is the right failure mode when the secret hasn't been provisioned yet (e.g. Crossplane still reconciling).

If we don't hear back, the stale bot will eventually close this, and the design summary above should let anyone pick the work up via #3735. Thanks for the contribution!

PR #3781, opened 2026-03-17
fix(argo-cd): Set ServerSideApply sync option annotation on CRDs

### LINKED ISSUE #3704: argo-cd: ApplicationSets CRD too large for client-side-apply

It seems like the ApplicationSets crd in argocd 3.3.0 is finally too large for client-side-apply and hits the etcd value limit:

```
CustomResourceDefinition.apiextensions.k8s.io "applicationsets.argoproj.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes
```

That means argocd-helm can no longer be installed with a default argocd configuration. We should probably document in the install instructions that you must set server side apply annotations for CRDs now.

### Related helm chart

argo-cd

Either add server side apply requirement to install instructions or set server side apply annotations by default in CRDs.

### Additional context

This is also mentioned in the argocd upgrade guide: https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/3.2-3.3/#applicationset-crd-exceeds-the-size-limit-for-client-side-apply

rouke-broersma:
Checklist:

fix #3704

### DIFF
--- a/charts/argo-cd/README.md.gotmpl
+++ b/charts/argo-cd/README.md.gotmpl
@@ -377,6 +377,8 @@ Changes in the `CustomResourceDefinition` resources shall be fixed easily by cop
 
 ### Custom resource definitions
 
+The chart enables server-side apply for CustomResourceDefinitions when using Argo CD to install Argo CD using the `crds.annotations.argocd.argoproj.io/sync-options` annotation. This default avoids client-side apply size limits that can affect large CRDs (for example, ApplicationSet CRDs) and follows the upstream upgrade guidance: https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/3.2-3.3/#applicationset-crd-exceeds-the-size-limit-for-client-side-apply. If you install the CRDs manually using kubectl make sure to enable server-side apply.
+
 Some users would prefer to install the CRDs _outside_ of the chart. You can disable the CRD installation of this chart by using `--set crds.install=false` when installing the chart.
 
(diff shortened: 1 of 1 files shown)


charts/argo-cd/README.md
```diff
@@ -377,6 +377,8 @@ Changes in the `CustomResourceDefinition` resources shall be fixed easily by cop
 
 ### Custom resource definitions
 
+The chart enables server-side apply for CustomResourceDefinitions by default using the `crds.annotations.argocd.argoproj.io/sync-options` annotation. This default avoids client-side apply size limits that can affect large CRDs (for example, ApplicationSet CRDs) and follows the upstream upgrade guidance: https://argo-cd.readthedocs.io/en/stable/operator-manual/upgrading/3.2-3.3/#applicationset-crd-exceeds-the-size-limit-for-client-side-apply.
```
tico24: This paragraph should mention that it specifically relates to installing through Argo CD. Otherwise the annotations are ignored. The implication at the moment is that this annotaion magically works for everyone whether they use argo cd to install the chart or not.

rouke-broersma: done. I see that you implemented the same by setting the annotations on the CRDs directly, is that preferred? I added them to the default values should someone for some reason want to disable server side apply, or want to enable another sync option but I am fine with them being on the CRDs as well.

tico24: oh. oops. that's actually 100% my fault. I genuinely thought that I'd just copied what you'd done.

I will update my approach. Yours is better.

rouke-broersma: Ah I should've added this in the template instead, once second

tico24:
CI is failing because I think you forgot to run helm docs after making that readme change

east4ming:
@rouke-broersma 

Will this change affect the cr of application and appset? Because I have `serversideapply=fals`e for the appset cr.
image

rouke-broersma:
This change affects the installation of the argocd crds themselves, not crs.

PR #3793, opened 2026-03-20
feat(argo-rollouts): add controller.createConfigmap flag

a1exx-k:
## Summary

The `argo-rollouts-config` ConfigMap name is hardcoded in the chart template with no way to disable its creation. This causes a Helm ownership conflict when deploying multiple `argo-rollouts` releases in the same namespace — a pattern used when running several controller instances with different `--instance-id` values to manage separate sets of Rollout resources.

## Changes

- `templates/controller/configmap.yaml`: wrap the ConfigMap in `{{- if .Values.controller.createConfigmap }}`
- `values.yaml`: add `controller.createConfigmap: true` (default preserves existing behaviour)

## Usage

A secondary controller instance (e.g. a `beta` release) can skip creating the ConfigMap and read the one owned by the primary release:

```yaml
controller:
  createConfigmap: false
  extraArgs:
    - --instance-id=beta
```

## Relation to argoproj/argo-rollouts#4664

[argoproj/argo-rollouts#4664](https://github.com/argoproj/argo-rollouts/pull/4664) proposes a `--configmap-name` flag for the controller binary, which would allow each instance to own a fully isolated ConfigMap. These two changes are complementary:

- **Until `--configmap-name` lands**: set `createConfigmap: false` on secondary releases to avoid the Helm conflict; all instances share the primary ConfigMap.
- **Once `--configmap-name` lands**: set `createConfigmap: true` on each release and point each controller at its own ConfigMap via `extraArgs`, achieving full isolation.

Checklist:

### DIFF
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -19,4 +19,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: added
-      description: Added flag to optionally disable argo-rollouts-config ConfigMap creation 
+      description: Added flag to optionally disable argo-rollouts-config ConfigMap creation


tico24:
Please put the checklist back that your AI removed and do the things. Keep this in draft until done please.

a1exx-k:
@tico24 all addressed.

charts/argo-rollouts/values.yaml
```diff
@@ -251,6 +251,12 @@ controller:
     # - name: "argoproj-labs/sample-nginx" # name of the plugin, it must match the name required by the plugin so it can find it's configuration
     #   location: "file://./my-custom-plugin" # supports http(s):// urls and file://
 
+  # -- Whether to create the argo-rollouts-config ConfigMap. Set to false when running multiple
```
tico24, line 254: you have not run helm-docs.sh

charts/argo-rollouts/templates/controller/configmap.yaml
```diff
@@ -19,3 +20,4 @@ data:
   trafficRouterPlugins: |-
     {{- toYaml . | nindent 4 }}
   {{- end }}
+{{- end }}
```
tico24: nit: I'd prefer an empty line at the end, but there no strict guide for that in this repo. But seeming as you'll have to push the docs anyway you might as well?

charts/argo-rollouts/values.yaml
a1exx-k: Missed that, my bad! Addresed.

charts/argo-rollouts/templates/controller/configmap.yaml
a1exx-k: np, addressed

PR #3794, opened 2026-03-20
ci(github): add draft PR lifecycle workflows

jmeridth:
## What

Add two GitHub Actions workflows to enforce a draft-first PR lifecycle. All newly opened PRs are automatically converted to draft and labeled with `Mark Ready When Ready`. Once all checks pass, the label triggers automatic promotion to ready for review.

## Why

Reduces manual toil for contributors -- no need to remember to open as draft or manually mark ready after CI passes. The automation handles the full lifecycle by default.

## Notes

- Both workflows use `pull_request_target` instead of `pull_request` so the `GITHUB_TOKEN` retains write permissions for PRs from forks. GitHub caps fork PR tokens at read-only on `pull_request` events regardless of the `permissions` block.
- `pull_request_target` is safe here because neither workflow checks out or executes code from the PR branch -- they only invoke pinned-SHA actions.
- `default-pr-to-draft.yaml` triggers only on `opened` to avoid fighting with `mark-ready-when-ready.yaml` when a PR is promoted -- no re-draft loop possible.
- `mark-ready-when-ready.yaml` renamed from `.yml` and updated to comply with GHA standards: `permissions: {}` at workflow level, job-level permissions with comments, corrected concurrency group key.
- `contents: write` on the mark-ready job is retained -- required internally by the `kenyonj/mark-ready-when-ready` action.
- The `Mark Ready When Ready` label must exist in the repo (already created).

### DIFF
--- a/.github/workflows/default-pr-to-draft.yaml
+++ b/.github/workflows/default-pr-to-draft.yaml
@@ -1,0 +1,37 @@
+name: Default PR to Draft
+
+on:
+  pull_request_target:
+    types: [opened]
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request_target' }}
+
+jobs:
+  default-to-draft:
+    name: Convert PR to draft and add label
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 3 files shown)


yu-croco (approved):
🚀

tico24 (approved):

PR #3806, opened 2026-03-26
feat(argo-cd): update chart versioning and testing logic

abuzersaj:
Updating version for CI/CD test.


jmeridth:
@abuzersaj I respect your role as a pentester. We need it. But if you're going to do it, please coordinate or add info in your PR description expressing that intent. Otherwise you'll get flagged as a bad actor. 🙇

PR #3693, opened 2026-01-28
feat(argo-cd): Add AWS TargetGroupConfiguration support for Gateway API

ArieLevs:
AWS Load Balancer Controller does not support the standard Gateway API BackendTLSPolicy for HTTPS backends.
This adds support for the AWS-specific TargetGroupConfiguration CRD (gateway.k8s.aws/v1beta1) as an alternative.

Changes:
  - Add templates/argocd-server/aws/targetgroupconfiguration.yaml
  - Add server.aws.targetGroupConfiguration values with defaultConfiguration
    and routeConfigurations support
  - Update backendTLSPolicy comment to note AWS incompatibility
  - Document AWS ALB Gateway API usage in README.md
---

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #3858, opened 2026-04-28
feat(argo-cd): add AWS Gateway API support for gRPC backend routing

nooop3:
Description:

Adds AWS Load Balancer Controller Gateway API support for Argo CD server gRPC traffic.

Summary:
  - Adds an AWS-specific HTTPRoute for gRPC requests, matching Content-Type: application/grpc*.
  - Adds a TargetGroupConfiguration template for AWS ALB target group protocol settings.
  - Introduces server.httproute.aws values for enabling the feature, customizing route rules, labels, annotations, and target group route  configuration.
  - Defaults the gRPC route target group to protocolVersion: GRPC while preserving HTTP/1 behavior for non-gRPC HTTPRoute traffic.

References:

- https://github.com/argoproj/argo-helm/issues/3550#issuecomment-4332481448

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.3.8
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 9.5.5
+version: 9.5.6
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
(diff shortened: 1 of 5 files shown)


mkilchhofer:
I assume that this PR aims to add the exact same we already declined, right?
- https://github.com/argoproj/argo-helm/pull/3693

If so: I'd like to decline this PR here as well as Gateway API should solve the annotation hell on the Ingresses using a generic high-level resource. We don't want to have opinionated config again for every cloud service provider.

nooop3:
Yes, both would add the AWS Load Balancer Controller CRD `TargetGroupConfiguration`.
This PR aligns with the Ingress approach to support gRPC routing with the same hostname for `HTTPRoute` on AWS.
Alternatively, if we don’t rely on cloud provider-specific resources, we can achieve the same result by defining Kubernetes objects directly via extraObjects, which is my current approach.

Thanks for your time, close now.

Issue #3815, opened 2026-03-31
argocd image updater ServiceMonitor


frittenlab:
### Describe the bug

I am facing issue with the current ServiceMonitor for the argocd image updater. The argocd image updater service is configured to use port 8443 with HTTPS as the metrics port. The default ServiceMonitor is not aware of this. So Prometheus tries to scrape the service with http which does not work. In the argocd image updater logs I can see the request from Prometheus:

`2026/03/28 11:42:06 http: TLS handshake error from 10.42.0.151:40176`

I also tested the metrics endpoint with curl:

```
curl 10.42.8.140:8443/metrics
Client sent an HTTP request to an HTTPS server.
```

I added the following to the existing ServiceMonitor spec field:

```
spec:
  endpoints:
  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    interval: 30s
    path: /metrics
    port: metrics
    scheme: https
    tlsConfig:
      insecureSkipVerify: true
```

This allows Prometheus to scrape the metrics via the HTTP port 8443 without an error. I can create a PR for the helm chart if you are interested. 

### Related helm chart

argocd-image-updater

### Helm chart version

1.1.4

### To Reproduce

1. install the argocd image updater helm chart with the ServiceMonitor enable
2. Wait for Prometheus to scrape the ServiceMonitor
3. Observe the target is down in Prometheus

### Expected behavior

Prometheus should be able to scrape the Service with the default ServiceMonitor that is created by the helm chart

ystkfujii:
Hi @frittenlab , 
I noticed that you mentioned you could create a PR for this. Are you still interested in working on it?
If not, I’d be happy to pick this up.

frittenlab:
@ystkfujii i am really busy this week. You are welcome to pick up the issue. Thanks for your help

ystkfujii:
This issue may be closed by https://github.com/argoproj/argo-helm/pull/3874

jmeridth:
[...]

PR #3874, opened 2026-05-05
fix(argocd-image-updater): Add ServiceMonitor TLS and authorization options

ystkfujii:
## Summary

This PR adds optional ServiceMonitor settings for Argo CD Image Updater metrics scraping:

- `metrics.serviceMonitor.scheme`
- `metrics.serviceMonitor.tlsConfig`
- `metrics.serviceMonitor.authorization`

This is related to #3815, where the metrics endpoint is served over HTTPS on port `8443`, but the generated ServiceMonitor currently does not allow configuring the scrape scheme or TLS settings.

With this change, users can configure HTTPS metrics scraping, for example:

```yaml
metrics:
  enabled: true
  serviceMonitor:
    enabled: true
    scheme: https
    tlsConfig:
      insecureSkipVerify: true
```

This PR intentionally does not change the default ServiceMonitor behavior. If scheme is not specified, Prometheus Operator defaults to HTTP. Changing the chart default to HTTPS would be a behavior change for existing users, and making insecureSkipVerify: true the default would not be a safe default.

This PR also uses authorization instead of bearerTokenFile, since bearerTokenFile is deprecated in the Prometheus Operator API.

Checklist:

* [ ] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/charts/argocd-image-updater/Chart.yaml
+++ b/charts/argocd-image-updater/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: argocd-image-updater
 description: A Helm chart for Argo CD Image Updater, a tool to automatically update the container images of Kubernetes workloads which are managed by Argo CD
 type: application
-version: 1.1.5
+version: 1.1.6
 appVersion: v1.1.1
 home: https://github.com/argoproj-labs/argocd-image-updater
 icon: https://argocd-image-updater.readthedocs.io/en/stable/assets/logo.png
@@ -19,4 +19,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: fixed
-      description: Remove duplicate nodePort key from Service template when service.type is NodePort
+      description: Add ServiceMonitor options for secure metrics scraping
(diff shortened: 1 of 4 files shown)


r0bj:
Maybe we should change the default ServiceMonitor behavior, since it is currently broken with the default values.
In my opinion, it provides a better user experience when the defaults just work. For example, a configuration like this:
```
metrics:
  enabled: true
  serviceMonitor:
    enabled: true
```
should generate a working setup, but currently it does not.

What do you think?

ystkfujii:
@r0bj 

I agree that the default configuration should work out of the box.

I’ve updated the ServiceMonitor defaults for `scheme` and `tlsConfig`, so it now uses HTTPS by default.

However, making it fully work also requires an `authorization` setting with a token Secret. I’m not sure what the best default behavior should be here, since creating the Secret by default would mean creating a long-lived credential automatically.

What do you think about keeping authorization configurable and letting users provide the Secret explicitly?

jmeridth (approved):

mbevc1 (approved):

PR #3882, opened 2026-05-09
feat(argo-workflows): allow configuring the container and pod-level securityContext for the crd upgrade job

### LINKED ISSUE #3844: Support for running the CRD-install job as a non-root user using a more restrictive securityContext

Currently, the install-crds job runs using the default security context: https://github.com/argoproj/argo-helm/blob/main/charts/argo-workflows/templates/crds-install-job.yaml#L87, which doesn't explicitely allow it to run as non-root to comply with strict security policies.

### Related helm chart

argo-workflows

It would be great to run it using a restrictive securityContext by default.

_No response_

### Additional context

I'd be happy to contribute this feature!

chgl:
Closes #3844 

Checklist:

### DIFF
--- a/charts/argo-workflows/values.yaml
+++ b/charts/argo-workflows/values.yaml
@@ -47,8 +47,6 @@ crds:
     tolerations: []
     # -- Image pull secrets for the CRD install Job
     imagePullSecrets: []
-    # -- Extra environment variables to provide to the CRD install Job container
-    extraEnv: []
     # -- Security context for the CRD install Job container
     securityContext:
       readOnlyRootFilesystem: true
@@ -61,6 +59,8 @@ crds:
       capabilities:
         drop:
           - ALL
+    # -- Extra environment variables to provide to the CRD install Job container
+    extraEnv: []
     # - name: HTTP_PROXY
     #   value: "http://proxy.example.com:8080"
     # - name: HTTPS_PROXY
--- a/charts/argo-workflows/values.yaml
+++ b/charts/argo-workflows/values.yaml
@@ -34,12 +34,33 @@ crds:
     resources: {}
     # -- Node selector for the CRD install Job
     nodeSelector: {}
+    # -- Pod security context for the CRD install Job pod
+    podSecurityContext:
+      runAsNonRoot: true
+      fsGroup: 65532
+      runAsUser: 65532
+      runAsGroup: 65532
+      fsGroupChangePolicy: "OnRootMismatch"
+      seccompProfile:
+        type: RuntimeDefault
     # -- Tolerations for the CRD install Job
     tolerations: []
     # -- Image pull secrets for the CRD install Job
     imagePullSecrets: []
     # -- Extra environment variables to provide to the CRD install Job container
     extraEnv: []
+    # -- Security context for the CRD install Job container
+    securityContext:
+      readOnlyRootFilesystem: true
+      runAsNonRoot: true
+      allowPrivilegeEscalation: false
+      runAsUser: 65532
+      runAsGroup: 65532
+      seccompProfile:
+        type: RuntimeDefault
+      capabilities:
+        drop:
+          - ALL
     # - name: HTTP_PROXY
     #   value: "http://proxy.example.com:8080"
     # - name: HTTPS_PROXY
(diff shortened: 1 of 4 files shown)


charts/argo-workflows/values.yaml
```diff
@@ -34,12 +34,33 @@ crds:
...
+      seccompProfile:
+        type: RuntimeDefault
     # -- Tolerations for the CRD install Job
     tolerations: []
     # -- Image pull secrets for the CRD install Job
     imagePullSecrets: []
     # -- Extra environment variables to provide to the CRD install Job container
     extraEnv: []
+    # -- Security context for the CRD install Job container
+    securityContext:
```
tico24, line 44: This block landed between `extraEnv: []` and its example comments, so the `# - name: HTTP_PROXY` lines just below now read as if they belong to `securityContext`. Worth putting `securityContext` above `extraEnv`, or moving the proxy examples back up next to it.

charts/argo-workflows/values.yaml
```diff
@@ -34,12 +34,33 @@ crds:
...
+      readOnlyRootFilesystem: true
```
tico24, line 45: kubectl writes a discovery/http cache under `$HOME/.kube/cache`. With `readOnlyRootFilesystem: true` and no writable mount, that write fails. I think it's non-fatal (kubectl falls back to uncached), but have you run this against a real cluster with `crds.full` to confirm the apply still succeeds? CI might not exercise the hook job.

chgl: Good point. I tested it locally against a KinD cluster and the job still ran successfully (when explicitly setting Chart.version to 1.0.14), logging just:

```
customresourcedefinition.apiextensions.k8s.io/clusterworkflowtemplates.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/cronworkflows.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workflowartifactgctasks.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workfloweventbindings.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workflows.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workflowtaskresults.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workflowtasksets.argoproj.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/workflowtemplates.argoproj.io serverside-applied
```

charts/argo-workflows/values.yaml
```diff
@@ -34,10 +34,31 @@ crds:
     resources: {}
     # -- Node selector for the CRD install Job
     nodeSelector: {}
+    # -- Pod security context for the CRD install Job pod
+    podSecurityContext:
```
tico24: This is the only populated `podSecurityContext` default in the chart; controller and server both default to `{}`. The container `securityContext` below already sets `runAsUser`/`runAsGroup`/`runAsNonRoot`/`seccompProfile`, so the only thing this block adds on top is `fsGroup`/`fsGroupChangePolicy`, which only affects volume ownership. The one volume here is the `hostPath` mount, and it's `readOnly: true`, so fsGroup never does anything. Was the populated default intentional, or would `{}` (matching the other components) do the same job?

chgl: It's intentional but definitely redundant: I generally just set it out of consistently setting secure defaults in any security context I find. But matching it with the defaults in the rest of the chart makes more sense - I can update it.

charts/argo-workflows/values.yaml
chgl: Unrelated to this PR: should we remove the `--` here so the `upgradeJob` object isn't added to the README.md - and updated every time one of its fields is changed.

tico24: yeh we probably should. good spot

tico24 (approved):

rotarur:
when will this be released? I need the crd job security contexts

rotarur:
cc @vladlosev @yu-croco @jmeridth could you please review this PR?

jmeridth (approved):

jmeridth:
@rotarur https://github.com/argoproj/argo-helm/releases/tag/argo-workflows-1.0.17

rotarur:
> @rotarur https://github.com/argoproj/argo-helm/releases/tag/argo-workflows-1.0.17

thank you sir 🙇

PR #3912, opened 2026-06-03
fix(argo-cd): add namespace to rolebinding subjects

Rasnar:
The rolebinding subjects were missing the namespace field, which caused issues when the rolebinding was applied in a different namespace than the default. The service account itself already has the correct namespace set, see: https://github.com/argoproj/argo-helm/blob/dacd465d6f0c53fa52e2b2741a6a75790dd5d589/charts/argo-cd/templates/redis-secret-init/serviceaccount.yaml#L7

We saw this issue because of the breaking change in kustomize v5.8+, see: 
https://github.com/kubernetes-sigs/kustomize/issues/6058

somewhat related to these too:
https://github.com/argoproj/argo-cd/issues/17803#issuecomment-3860882581
https://github.com/kubernetes-sigs/kustomize/issues/5566

Let me know if my fix is just plain wrong, but I'm pretty sure this was just working with kustomize by luck before (before the breaking changes introduced in v5.8+). Currently we need to fix this with such patch in our deployments:
```
  - target:
      kind: RoleBinding
      name: argocd-redis-secret-init
    patch: |-
      kind: RoleBinding
      metadata:
        name: argocd-redis-secret-init
      subjects:
        - kind: ServiceAccount
          name: argocd-redis-secret-init
          namespace: argocd
```
even though, we were passing it explicitly in the `Kustomization.yaml`:
```
helmCharts:
  - name: argo-cd
    repo: <some-custom-repo>
    version: 9.5.17
    releaseName: argocd
    namespace: argocd
```

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.4.3
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 9.5.17
+version: 9.5.18
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: fixed
-      description: Fix env vars ordering to allow user overrides via controller.env.
+      description: Add namespace to redis-secret-init rolebinding subjects.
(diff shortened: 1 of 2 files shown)


tico24 (approved):
Fix looks good, happy to take it. One small clarification on the framing for anyone reading later: on a RoleBinding (unlike a ClusterRoleBinding) the subject namespace isn't actually required. If you leave it empty and the RBAC authorizer defaults it to the binding's own namespace, which here is the same `argo-cd.namespace` the ServiceAccount uses. So a plain `helm install` was never broken by this.

What you're really working around is the kustomize v5.8 change in kubernetes-sigs/kustomize#6058, where it stopped filling in the namespace on fields that didn't already have one and left the subject bare. Setting it explicitly is the right fix for that, and it lines up with what the metadata namespace and the ServiceAccount already do here.

yu-croco (approved):

PR #3940, opened 2026-06-22
fix(argo-workflows): skip static credential rendering when useSDKCreds is true

daigo424:
## What this PR does

When `artifactRepository.s3.useSDKCreds: true` is set (e.g. for IRSA / IAM Roles for Service Accounts on EKS), the chart still renders `accessKeySecret` and `secretKeySecret` because `useStaticCredentials` defaults to `true`.

Since `accessKeySecret` is not provided by the user in this case, the template fails with a nil pointer error:

> `nil pointer evaluating interface {}.key`

This makes the Helm release unrenderable, causing ArgoCD to show `ComparisonError` and the `argo` namespace to never be created.

## Changes

Conditionally skip the `accessKeySecret` / `secretKeySecret` / `sessionTokenSecret` block when `useSDKCreds: true`, regardless of the `useStaticCredentials` value.

```diff
- {{- if .Values.useStaticCredentials }}
+ {{- if and .Values.useStaticCredentials (not .Values.artifactRepository.s3.useSDKCreds) }}
```

## Backward compatibility

- Users relying on static credentials (`useStaticCredentials: true` with `accessKeySecret` set) are unaffected.
- Users using SDK-based auth (`useSDKCreds: true`) no longer need to set `useStaticCredentials: false` explicitly or provide dummy credential values as a workaround.

Checklist:

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v4.0.6
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 1.0.16
+version: 1.0.17
 icon: https://argo-workflows.readthedocs.io/en/stable/assets/logo.png
 home: https://github.com/argoproj/argo-helm
 sources:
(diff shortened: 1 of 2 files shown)


vladlosev (dismissed):
LGTM

vladlosev (approved):

jmeridth (approved):

PR #3943, opened 2026-06-23
feat(argo-workflows): add envFrom to controller and argo-server containers

somaz94:
## Summary

Adds an optional `controller.envFrom` and `server.envFrom` to the argo-workflows chart, so the workflow-controller and argo-server containers can load environment variables from ConfigMaps/Secrets via `envFrom`.

This mirrors patterns already in the chart and the wider repo:
- the Workflow main container already has `mainContainer.envFrom` (added in #1736)
- the sibling `argo-events` chart exposes per-controller `envFrom`
- both `controller` and `server` already expose `extraEnv`, so `envFrom` is the natural companion

Default is `[]`, so the rendered output is unchanged for existing users (the field only renders when set).

## Validation (local)

- `helm lint`: 1 chart linted, 0 failed
- `./scripts/helm-docs.sh` (helm-docs v1.9.1): README regenerated, diff is only the two new rows
- `helm template`:
  - default: no `envFrom` rendered, default manifests byte-identical except the `helm.sh/chart` version label
  - `controller.envFrom` set: renders `envFrom` on the controller container
  - `server.envFrom` set: renders `envFrom` on the argo-server container

## Changes

- `values.yaml`: `controller.envFrom` + `server.envFrom` (default `[]`)
- `templates/controller/workflow-controller-deployment.yaml`, `templates/server/server-deployment.yaml`: wire `envFrom`
- `README.md`: helm-docs rows
- `Chart.yaml`: version 1.0.22 -> 1.0.23 + changelog annotation

PR #3951, opened 2026-07-02
feat(argo-cd): support custom secret name for redis password

maarous:
### Summary

This MR adds a new `redis.existingSecret` value to the argo-cd chart, allowing users to reference a pre-existing Kubernetes Secret for the Redis password instead of the hardcoded `argocd-redis` secret name.

### Motivation

Currently, the Redis deployment template hardcodes `argocd-redis` as the secret name in the `REDIS_PASSWORD` `secretKeyRef`. Users who manage their Redis credentials externally have no way to point the chart at their own secret and are forced to name it `argocd-redis`. This change makes the secret name configurable while keeping the existing behavior as the default.

### Checklist


jmeridth:
@maarous Thanks for the contribution, and sorry it sat two months without feedback. The use case is legitimate, but the PR as written has a functional gap that would break Redis auth for anyone who uses it, so it needs changes before it can move forward.

**The Redis clients don't follow the new secret name.** This PR changes where the Redis server (and exporter sidecar) reads `REDIS_PASSWORD`, but argocd-server, repo-server, and the application controller get their `REDIS_PASSWORD` via the `argo-cd.redisPasswordSecretRef` helper in `_helpers.tpl`. In exactly the scenario this PR targets (embedded Redis with `redisSecretInit.enabled: false`), that helper still points at the hard-coded `argocd-redis` secret with `optional: true`. So with `redis.existingSecret: my-secret` and no `argocd-redis` secret present, the server enforces the password from your secret while every client silently connects without one, and auth fails cluster-wide.

To make this work end to end:

1. Update the final branch of `argo-cd.redisPasswordSecretRef` to honor the new value, roughly `name: {{ default "argocd-redis" .Values.redis.existingSecret }}`.
2. In that branch, make `optional: false` when `redis.existingSecret` is set. An explicitly named secret that's missing is a configuration error and should fail loudly, not degrade to no-auth.
3. Bump the chart version in `Chart.yaml` and add the `artifacthub.io/changes` changelog entry, otherwise lint CI will fail (see the PR checklist in CONTRIBUTING).

Two smaller notes: please keep the value docs explicit that the secret key must be `auth` and that this only applies to embedded Redis (external Redis already has `externalRedis.existingSecret`), and be aware there's a broader discussion in #3734 about restructuring redis secret configuration, so keeping this change minimal and consistent with the existing flat style is the right call for now.

Happy to review once updated.

maarous:
Hi @jmeridth , thanks for the detailed review!
No worries at all about the delay, it was summer and everyone deserves a holiday :smile: 

I've pushed updates addressing all the points. Ready for another look whenever you have time. Hopefully a bit sooner than two months this time :sweat_smile:

tico24 (commented):
The checklist from the PR template is missing from the description. Can you put it back and tick what applies? ([template](https://github.com/argoproj/argo-helm/blob/main/.github/pull_request_template.md))

charts/argo-cd/templates/_helpers.tpl
```diff
@@ -345,9 +345,10 @@ optional: false # Secret is not optional in this case !
 
     {{- else -}}
     {{- /* All other use cases (e.g. disabled pre-install Job) */ -}}
-name: "argocd-redis"
+    {{- /* Secret is required when specifying redis.existingSecret, otherwise it is optional */ -}}
+name: {{ default "argocd-redis" .Values.redis.existingSecret }}
```
tico24, line 349: What about redis-ha? I rendered it with `redis-ha.enabled=true`, `redisSecretInit.enabled=false` and `redis.existingSecret=my-secret`. argocd-server, repo-server and the application controller all read `my-secret`, but redis-ha still reads `redis-ha.existingSecret`, which defaults to `argocd-redis`. So unless someone sets both, the clients and Redis are looking at different passwords. Should this branch leave redis-ha alone, or should the `redis.existingSecret` docs say to set `redis-ha.existingSecret` to the same name?

jmeridth:
@maarous confirming the two items from my August review are both resolved exactly as discussed: the `argo-cd.redisPasswordSecretRef` fallback branch now honors `redis.existingSecret`, and `optional` flips to `false` when the value is set so a missing named secret fails loudly. Thank you for the thorough rework. @tico24's redis-ha question is the remaining substantive item. One mechanical note while you're addressing it: per the convention discussion on #4060 (new backwards-compatible values ship as patches, matching #3943/#3967/#3982/#3659), this should be a patch bump on current main rather than `10.10.0`.

charts/argo-cd/templates/_helpers.tpl
maarous: I've added the following check to the helpers file (in line 346) to prevent this issue:
```go
    {{- else if (index .Values "redis-ha").enabled -}}
    {{- /* Embedded redis-ha with disabled pre-install Job: follow the Secret used by the redis-ha subchart */ -}}
name: {{ index .Values "redis-ha" "existingSecret" }}
key: auth
optional: true
```

maarous: I added also the PR template. Thanks @tico24 for the review !

maarous:
@jmeridth I've pushed an update addressing all the points.

charts/argo-cd/templates/_helpers.tpl
tico24: This breaks if `redis-ha.existingSecret` is a template. redis-ha runs it through `tpl`, so `"{{ .Release.Name }}-creds"` works on its side, but here it goes in as-is and the chart won't render (`YAML parse error on argo-cd/templates/argocd-application-controller/statefulset.yaml`).

It also ignores `redis-ha.authKey`. Set that to `password` and Redis reads `password` from the secret, but argocd-server, repo-server and the controller still read `auth`.

And if someone sets `existingSecret: null` so redis-ha creates its own secret, those three end up with an empty `name:`.

Could the name go through `tpl` and fall back to the redis-ha fullname, and the key come from `authKey`? Worth mentioning the redis-ha change in the changelog too, it only covers `redis.existingSecret` at the moment.

maarous: @tico24 done !

PR #3888, opened 2026-05-12
fix(argo-cd): support tpl in configs.cm.url

ystkfujii:
## Summary

Fixes #3875.

This change allows `configs.cm.url` in the `argo-cd` chart to use Helm template expressions.

## Testing

Tested with `helm template`.

Used the following test values:

```yaml
global:
  clusterFqdn: mynewfqdn.something

server:
  ingress:
    enabled: true
    controller: generic
    hostname: "{{ .Values.global.clusterFqdn }}"
    path: /argocd

configs:
  cm:
    url: "https://{{ .Values.global.clusterFqdn }}/argocd"
```

Renderd the manifests with:

```console
$ helm template argocd . \
  --namespace argocd \
  -f  values-3875.yam \
  --show-only templates/argocd-server/ingress.yaml \
  --show-only templates/argocd-configs/argocd-cm.yaml \
  > rendered.yaml
```

Confirmed that the argocd-cm URL is also rendered with the templated value:

```console
$ yq 'select(.kind == "ConfigMap" and .metadata.name == "argocd-cm") | .data.url' rendered.yaml
https://mynewfqdn.something/argocd
```

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #3964, opened 2026-07-09
feat(argo-cd): allow appending additional resource.exclusions without overriding the chart defaults

### LINKED ISSUE #3291: Allow appending to default resource.exclusion list

Currently we can either use the default `resource.exclusion` list or overwrite it ourselves. 

If we want to stay automatically up to date with the provided ArgoCD default exclusions, yet add some of our own, this is not currently possible with the chart.

### Related helm chart

argo-cd

I would like to have some property like `additionalExclusions` where we can provide a list of exclusions to be appended to the default list.

Every release, manually check if the default list changed and add this to our overwritten list (higher maintenance burden)

### Additional context

_No response_

walsm232:
Closes #3291

## Overview

`configs.cm."resource.exclusions"` is a single YAML-in-string block containing Argo CD's default resource exclusion list. Overriding it is all-or-nothing: to add one exclusion of your own on top of the defaults, you have to copy the entire default list into your values file, which then drifts out of sync as the chart's defaults change.

This adds `configs.cm.resourceExclusionsAdditional`, a plain list that gets appended onto the default `resource.exclusions` string at render time, so the chart's default list stays intact and users can layer their own entries on top of it, adapted here since `resource.exclusions` is a stringified block rather than a real Helm list.

### Prior art

Exposing a separate `*Additional`/`*Include` value that appends to chart-managed defaults (rather than replacing them) is an established pattern in other popular charts:

- **Kyverno** added [`config.resourceFiltersInclude` / `resourceFiltersIncludeNamespaces`](https://github.com/kyverno/kyverno/blob/main/charts/kyverno/values.yaml) so users can extend the default `resourceFilters` list without duplicating it, the same all-or-nothing problem this PR addresses (see kyverno/kyverno#9529).
- **kube-prometheus-stack** exposes [`prometheus.prometheusSpec.additionalScrapeConfigs`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml) to append user scrape configs on top of the chart-managed defaults.

### Backwards compatibility

**Non-breaking:** the new value defaults to `[]`, which is falsy in the merge template, so nothing changes when it's left unset. Verified the rendered `argocd-cm` output is byte-identical to before this change (aside from the chart version label) with default values.

### Testing

```
helm template charts/argo-cd
  --set 'configs.cm.resourceExclusionsAdditional[0].apiGroups[0]=example.com'
  --set 'configs.cm.resourceExclusionsAdditional[0].kinds[0]=MyCustomResource'
  -s templates/argocd-configs/argocd-cm.yaml
```

Confirms the appended entry renders correctly and the resulting `resource.exclusions` string still parses as valid YAML.

Checklist:

* [X] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.4.3
+version: 10.5.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: added
-      description: Added dnsPolicy/dnsConfig support for the secret initialization of Redis. Added hostNetwork support for Redis and the Redis secret initialization.
+      description: Add configs.cm.resourceExclusionsAdditional to append custom resource.exclusions entries without overriding the chart defaults
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -1044,6 +1044,7 @@ NAME: my-release
 | configs.cm."timeout.reconciliation.jitter" | string | `"60s"` | Maximum jitter added to the reconciliation timeout to spread out refreshes and reduce repo-server load |
 | configs.cm.annotations | object | `{}` | Annotations to be added to argocd-cm configmap |
 | configs.cm.create | bool | `true` | Create the argocd-cm configmap for [declarative setup] |
+| configs.cm.resourceExclusionsAdditional | list | `[]` | Additional resource exclusions to append to the default `resource.exclusions` list above, so that the defaults can be kept up to date without needing to duplicate/override them. These entries are always appended, never substituted: if you also set `resource.exclusions` yourself, they are appended to your value rather than to the chart defaults. |
 | configs.cmp.annotations | object | `{}` | Annotations to be added to argocd-cmp-cm configmap |
 | configs.cmp.create | bool | `false` | Create the argocd-cmp-cm configmap |
 | configs.cmp.plugins | object | `{}` | Plugin yaml files to be added to argocd-cmp-cm |
--- a/charts/argo-cd/templates/_helpers.tpl
+++ b/charts/argo-cd/templates/_helpers.tpl
@@ -216,9 +216,15 @@ Argo Configuration Preset Values (Influenced by Values configuration)
 Merge Argo Configuration with Preset Configuration
 */}}
 {{- define "argo-cd.config.cm" -}}
-{{- $config := omit .Values.configs.cm "create" "annotations" -}}
+{{- $config := omit .Values.configs.cm "create" "annotations" "resourceExclusionsAdditional" -}}
 {{- $preset := include "argo-cd.config.cm.presets" . | fromYaml | default dict -}}
-{{- range $key, $value := mergeOverwrite $preset $config }}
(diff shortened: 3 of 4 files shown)


walsm232:
I also just wanted to mention that I considered a cleaner-looking implementation using Helm's list functions directly:

```
{{- $existing := get $merged "resource.exclusions" | default "" | fromYamlArray }}
{{- $combined := concat $existing .Values.configs.cm.resourceExclusionsAdditional }}
{{- $_ := set $merged "resource.exclusions" (toYaml $combined) }}
```

This parses the existing value into a real list, concatenates it with the additional entries, and serializes the combined list once. It's more idiomatic Helm I think, but it has a downside for this chart specifically: the default `resource.exclusions` block has `###` comments documenting why each group of resources is excluded (network resources, Kyverno reporting resources, etc). Parsing that string into a list and re-serializing it with `toYaml` discards those comments.

I verified this with `helm template` against the chart's actual defaults: every `###` comment is gone from the rendered `argocd-cm` output once it goes through that round trip.

Since those comments might be meant to survive into the rendered `ConfigMap` (so someone running `kubectl get cm argocd-cm -o yaml` can see why each block exists), I kept `resource.exclusions` as an opaque string and appended to it as text instead. This never touches the existing string's contents, only appends to it, so the default comments are preserved regardless of whether `resourceExclusionsAdditional` is set.

If you folks reviewing this are fine with losing those comments in the rendered `ConfigMap`, I'm happy to switch to the `fromYamlArray/concat/toYaml` version instead, it's a smaller diff and probably easier to follow at a glance to be honest.

jmeridth:
@walsm232 Triaged this one today. The approach looks right to me: it matches the consensus in #3291 (string concatenation, as @jemag suggested), and your reasoning for rejecting the `fromYamlArray`/`concat` alternative holds up, since parsing and re-serializing would strip the explanatory `###` comments from the rendered ConfigMap. The meta-key inside `configs.cm` also follows existing precedent (`create` and `annotations` are omitted the same way). @mkilchhofer noted in the issue that he wanted exactly this during the v3 prep, so tagging him to confirm the direction.

Two asks to get it moving:

1. **Rebase onto current main.** The chart is at 10.4.2 now, so the branch conflicts (likely just Chart.yaml version and changelog).
2. **One docs sentence.** The append also applies when a user overrides `resource.exclusions` themselves, which I think is the right semantic, but it's worth stating explicitly in the value's comment so nobody is surprised.

Sequencing note: #3888 patches the same `argo-cd.config.cm` helper block. Since your change restructures the loop and that one is a 3-line insert, the plan is to land this one first so the trivial rebase falls on #3888.

Thanks for picking this up and for un-staling the issue.

walsm232:
Thanks @jmeridth for taking the time to review!

Both asks should hopefully be addressed now:

- **Rebased onto current `main`**
- **Added the docs sentence:** _"These entries are always appended, never substituted: if you also set `resource.exclusions` yourself, they are appended to your value rather than to the chart defaults."_

and I also re-verified after the rebase:
- With `resource.exclusions` overridden and `resourceExclusionsAdditional` set, the custom entry appends to the user's value (which matches the documented behaviour).
- The combined string still parses as valid YAML.
- With the value unset, rendered `argocd-cm` is identical to `main` apart from the `helm.sh/chart` version label.

jmeridth (approved):
Both asks addressed: rebased onto current main with a correct minor bump, and the append-semantics docs sentence is in the values comment and regenerated README. Helper logic matches what was endorsed in the earlier review, and CI is fully green including the install test. Thanks @walsm232 for the quick turnaround and the re-verification notes.

Merge-order reminder for whoever merges: #3888 patches the same `argo-cd.config.cm` helper and takes the trivial rebase after this lands, per the plan. Also note #4026 targets chart 10.5.0 as well, so the second of the two to merge needs a version bump adjustment.

mbevc1 (approved):

PR #3991, opened 2026-07-31
feat(argo-cd): add optional h2c service port for Gateway API gRPC backend protocol selection

peiyuko:
In insecure mode the Argo CD server multiplexes the web UI (HTTP/1.1 only) and
gRPC (HTTP/2 only) on a single container port, and both the HTTPRoute and the
GRPCRoute target `servicePortHttp`. `appProtocol` is single-valued per Service
port, so that one port cannot describe both: leaving it unset relies on the
implementation guessing the backend protocol, while setting `kubernetes.io/h2c`
on it would break the UI.

Per [GEP-1911], implementations only *MAY* infer the backend protocol from the
Route type, so gRPC works on some controllers and not others. This adds an
optional second Service port targeting the same container port, whose
`appProtocol` ([KEP-3726]) and port name can be set to declare the protocol
explicitly instead. The GRPCRoute uses that port when set, while the HTTPRoute
keeps using the http port.

`servicePortHttp2` is empty by default, so the rendered output is unchanged and
the GRPCRoute keeps targeting `servicePortHttp` — controllers that already infer
h2c correctly need no changes. The port is only rendered when `server.insecure`
is `true`, since h2c does not apply to a TLS backend.

This continues #3585, which was closed by the stale bot, and applies the review
feedback on that PR to gate the port on `server.insecure`. The existing https
port is kept rather than replaced, since removing it would be a breaking change.

[GEP-1911]: https://gateway-api.sigs.k8s.io/geps/gep-1911/
[KEP-3726]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.0
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.3.1
+version: 10.3.2
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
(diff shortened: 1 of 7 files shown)


Untersander:
Great PR, thank you for the implementation. I came here to fix exactly the same issue.
Looks like you fixed it how I would have done. Hope someone finds time to review this.

jmeridth:
@peiyuko conflicts. We've also release argo-cd 3.5.0 chart. Make sure you're still good. We'll get reviews on this PR once you've rebased/merged.

peiyuko:
Conflicts resolved, thanks!

tico24 (dismissed):

tico24 (approved):

jmeridth (approved):

PR #4004, opened 2026-08-08
chore(github): switch CI to Helm 4 and bump chart-testing to v3.14.0

jmeridth:
Kudos to @tico24 for noticing this.

## What/Why

Argo CD 3.5 bundles Helm 4 (`hack/tool-versions.sh` pins `helm4_version=4.2.1`), so published charts are rendered by Helm 4 at runtime. This switches CI from Helm 3.10.1 to 4.2.3 so we lint and install-test with the same major version, and bumps chart-testing v3.11.0 to v3.14.0 (workflow and local lint image) as the release most likely to be Helm 4 compatible. Helm 3 also enters its end-of-support window soon.

Per review feedback, `scripts/lint.sh` now installs the CI Helm version into the chart-testing container before linting, since the ct image bundles its own Helm 3. While making that change we found the script's unquoted `docker run ... -c cd /workdir ct lint ...` meant sh only ever executed `cd`, so the local lint had been a silent no-op; the quoting is fixed and it now runs for real.

## Proof it works

All 6 charts pass `helm lint` and `helm template` under Helm 4.2.3 locally, with default values and every `ci/*-values.yaml` variant. The repaired `scripts/lint.sh` was run end to end: Helm 4.2.3 installed inside the ct v3.14.0 container and `ct lint` passed against all six charts, so the ct-on-Helm-4 path has now been exercised for real (locally; CI's changed-chart detection still skips it in this PR since no charts are touched).

## Risk + AI role

Low: CI-only change, no chart content touched. Charts stay installable with Helm 3 (apiVersion v2, unchanged packaging); what changes is that CI no longer guards Helm 3 behavior. AI-assisted (version research, edits, local verification), directed and reviewed by a maintainer.

## Review focus

Whether ct v3.11.0 to v3.14.0 brings behavior changes we care about, and whether dropping Helm 3 from CI deserves a note in README or release notes for consumers still installing with Helm 3.

---

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning) (N/A: no chart changes)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation) (N/A)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog) (N/A)
* [x] Any new values are backwards compatible and/or have sensible default. (N/A)
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests) (N/A: no chart changes)

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # This script runs the chart-testing tool locally. It simulates the linting that is also done by the github action. Run this without any errors before pushing.
+# Note: The chart-testing image ships its own Helm binary (still 3.x), while CI runs Helm 4. Rendering differences are only caught in CI.
 # Reference: https://github.com/helm/chart-testing
 set -eux
 
@@ -9,7 +10,7 @@ echo -e "\n-- Linting all Helm Charts --\n"
 docker run \
      -v "$SRCROOT:/workdir" \
      --entrypoint /bin/sh \
-     quay.io/helmpack/chart-testing:v3.10.0 \
+     quay.io/helmpack/chart-testing:v3.14.0 \
      -c cd /workdir \
      ct lint \
      --config .github/configs/ct-lint.yaml \
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,18 +1,24 @@
 #!/usr/bin/env bash
 # This script runs the chart-testing tool locally. It simulates the linting that is also done by the github action. Run this without any errors before pushing.
-# Note: The chart-testing image ships its own Helm binary (still 3.x), while CI runs Helm 4. Rendering differences are only caught in CI.
 # Reference: https://github.com/helm/chart-testing
 set -eux
 
 SRCROOT="$(cd "$(dirname "$0")/.." && pwd)"
 
+# Note: Also update in .github/workflows/lint-and-test.yml
+CT_VERSION=v3.14.0
+HELM_VERSION=v4.2.3
+
+# The chart-testing image ships its own Helm 3 binary, so install the Helm version used by CI before linting.
 echo -e "\n-- Linting all Helm Charts --\n"
 docker run \
      -v "$SRCROOT:/workdir" \
      --entrypoint /bin/sh \
-     quay.io/helmpack/chart-testing:v3.14.0 \
-     -c cd /workdir \
-     ct lint \
-     --config .github/configs/ct-lint.yaml \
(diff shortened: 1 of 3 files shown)


scripts/lint.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # This script runs the chart-testing tool locally. It simulates the linting that is also done by the github action. Run this without any errors before pushing.
+# Note: The chart-testing image ships its own Helm binary (still 3.x), while CI runs Helm 4. Rendering differences are only caught in CI.
```
tico24: The line above still says this simulates the github action, which it won't after this. Worth bumping helm inside the container too, or more trouble than it's worth?

tico24 (dismissed):
Rendered all six charts on 3.19.1 and 4.2.3 and they come out identical, bar Helm 4 adding a blank line before each `---`. `repo add`, `dependency build` and `package` all behave on 4.2.3 too, so I don't think anyone still installing with Helm 3 is going to notice.

ct v3.14.0 does still vendor helm.sh/helm/v3 3.18.6, but it shells out to whatever `helm` is on PATH, so CI should pick up 4. Not actually exercised here though: list-changed found no charts, so `ct lint` never touched one. Happy for the first chart PR after this to be the smoke test?

scripts/lint.sh
jmeridth: Worth it for completion. Will do once I wake up.

jmeridth: @tico24 Done in https://github.com/argoproj/argo-helm/commit/f138c37f. The ct image bundles its own Helm 3, so the script now installs the CI Helm version (v4.2.3) into the container before running ct lint. While in there I found the unquoted `-c cd /workdir ...` meant sh only ever executed `cd`, so the local script had been a no-op that always passed. With that fixed, ct lint v3.14.0 with Helm 4 has now run for real against all six charts locally, all clean, which also covers the smoke test from your review.

tico24 (approved):
Ran the fixed script and helm v4.2.3 does end up at `/usr/local/bin/helm` inside the ct v3.14.0 container, with ct picking it up.

It still prints `No chart changes detected` on this branch though, so I re-ran the same container command with `--all`: all six charts lint clean, every `ci/*-values.yaml` included.

yu-croco (approved):

PR #4008, opened 2026-08-10
feat(argo-events): Update NATS images and add Renovate coverage for JetStream dependencies

suryaval:
Checklist:

This MR is to ensure the helm chart has the changes performed here: https://github.com/argoproj/argo-events/pull/4134

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,6 +15,7 @@
     "**/charts/argo-cd/Chart.yaml",
     "**/charts/argo-cd/values.yaml",
     "**/charts/argo-events/Chart.yaml",
+    "**/charts/argo-events/values.yaml",
     "**/charts/argo-rollouts/Chart.yaml",
     "**/charts/argocd-image-updater/Chart.yaml",
     "**/.github/workflows/renovate.yaml"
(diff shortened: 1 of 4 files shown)


tico24:
It would be better to get this at the source and have renovate do the bumping like it does for redis exporter (to pick a random example)

suryaval:
@tico24 my miss - i accidentally moved it to ready for review. 
can you please elaborate how renovate picks it up.. the changes have been done on argo-events [repo](https://github.com/argoproj/argo-events/pull/4134)
any guidance on how i can run(force) renovate targeting this NATS images?

suryaval:
@tico24 Great point — I've added Renovate custom regex managers for the NATS JetStream images in `charts/argo-events/values.yaml` so they'll be bumped automatically going forward. Specifically:

- `nats` (the core NATS server image)
- `natsio/prometheus-nats-exporter`
- `natsio/nats-server-config-reloader`

These follow the same `postUpgradeTasks` pattern as existing entries (chart version bump via `renovate-bump-version.sh` + `helm-docs.sh` regeneration).

The manual bump in this PR sets the baseline for Renovate to track from.

tico24 (commented):
The changelog points at argo-events#4134, but I can't see it in v1.9.11, which is what this chart pins. It merged to master on 7 Aug and v1.9.11 went out on 13 Jul, so I don't think it's in a release yet. Almost certainly me misreading something, but can you point me at where it landed? We've parked chart PRs waiting on an upstream release before.

Separately, `latest` moving from nats:2.10.10 to 2.14.4 changes what existing installs get on upgrade, so 2.4.25 as a patch doesn't look right to me ([versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)).

renovate.json
```diff
@@ -70,6 +71,33 @@
       "depNameTemplate": "argoproj-labs/argocd-image-updater",
       "datasourceTemplate": "github-releases"
     },
+    {
+      "customType": "regex",
+      "fileMatch": ["charts/argo-events/values.yaml$"],
+      "matchStrings": [
+        "natsImage: (?<depName>nats):(?<currentValue>[^\\s]+)"
```
tico24, line 78: These match every `natsImage:` in the file, not just the `latest` entry (10 of them). Wouldn't Renovate then try to bump the pinned 2.8.1 and 2.9.x entries too? They're there so people can stay on an older NATS.

jmeridth:
@suryaval Thanks for the work here, and for engaging quickly on the Renovate suggestion. Closing this one out after verifying the upstream state; here's the reasoning:

**The image bump tracks unreleased upstream.** @tico24's read was right: the `nats:2.14.4` / exporter `0.20.1` / reloader `0.23.0` set comes from argo-events#4134, which is only on master. The chart pins v1.9.11, and that tag ships `nats:2.10.29` (exporter/reloader `0.14.0`). The chart *was* out of sync with v1.9.11, but the correct sync is #4045, which just got approved. Your 2.14.4 set will arrive naturally when the chart moves to the next argo-events release.

**The Renovate regex managers would cause ongoing problems.** They match every `natsImage:`/`metricsExporterImage:`/`configReloaderImage:` line in values.yaml, including the legacy pinned entries (2.8.x, 2.9.x) that exist so old eventbus pins keep working, so Renovate would perpetually open PRs bumping historical entries. More fundamentally, the JetStream block is mirrored upstream config (argo-events' tested version matrix), not a chart-owned dependency like redis_exporter, so it should only change when the pinned argo-events version changes, not whenever a new NATS image lands on Docker Hub.

If you want to pursue automation for this, the direction that fits the chart's mirror-upstream policy is syncing the JetStream block from the upstream tag as part of the appVersion Renovate bump (e.g. a `postUpgradeTasks` script that pulls `controller-config.yaml` from the new tag). That would be a genuinely useful contribution; happy to discuss in an issue first.

PR #4011, opened 2026-08-11
feat(argo-cd): add hostnetwork to redis

SkalaNetworks:
This PR simply adds the ability to toggle hostNetwork on the Redis deployed by the chart.
Useful when installing ArgoCD on an empty cluster (with no CNI) if ArgoCD is to install the CNI itself.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.4.2
+version: 10.4.3
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump redis_exporter to v1.90.0
+    - kind: added
+      description: Added dnsPolicy/dnsConfig support for the secret initialization of Redis. Added hostNetwork support for Redis and the Redis secret initialization.
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -1679,6 +1679,7 @@ NAME: my-release
 | redis.exporter.resources | object | `{}` | Resource limits and requests for redis-exporter sidecar |
 | redis.extraArgs | list | `[]` | Additional command line arguments to pass to redis-server |
 | redis.extraContainers | list | `[]` | Additional containers to be added to the redis pod |
+| redis.hostNetwork | bool | `false` | Host Network for redis pods |
 | redis.image.imagePullPolicy | string | `""` (defaults to global.image.imagePullPolicy) | Redis image pull policy |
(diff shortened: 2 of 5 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: added
-      description: Add an optional h2c server service port so Gateway API implementations can select the backend protocol for gRPC traffic when running in insecure mode
+      description: Added hostNetwork support for the Redis deployment
```
tico24, line 32: This also adds `redisSecretInit.hostNetwork`, which the description doesn't mention. ([changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog))

tico24 (commented):
`--set redis.hostNetwork=true` renders `hostNetwork: true` with `dnsPolicy: ClusterFirst` immediately under it, which is the combination that usually breaks cluster DNS resolution in the pod. Should enabling hostNetwork flip the default to `ClusterFirstWithHostNet`, or is leaving that to the user deliberate? The secret-init Job has no `dnsPolicy` value at all, so there's no way to change it there.

SkalaNetworks:
@tico24 Yes, the dns policy may seem "wrong", but I think there's several arguments against flipping it automatically:
- other components that already support hostNetwork don't do it (repo-server, argocd-server, application-controller...)
- insome cases `ClusterFirstWithHostNet` is actually detrimental. I don't see many uses for a hostNetwork redis, but the one I have involves using ArgoCD to bootstrap the very cluster on which it is installed, including CoreDNS/the CNI. `ClusterFirstWithHostNet` will try to resolve DNS through CoreDNS, which doesn't exist. So trying to pull a repository from the internet will fail, but if I set the policy to something like "Default", I inherit the DNS of the host, which is gonna work fine.

So I think we should let users choose the policy that best fits their use.

I fixed the changelog to include all elements. I also added dns config for the redis job, that way we cover everything.

SkalaNetworks:
@tico24 does this look good to you?

SkalaNetworks:
@jmeridth could you please rerun the pipeline? The CI wasn't happy with the lack of a new line at the end of a file

jmeridth:
@SkalaNetworks Apologies for the slow response on the rerun ping. Triaged this properly today; here's where it stands.

On the `dnsPolicy` question: I checked the chart and I'm with you. No other component auto-flips to `ClusterFirstWithHostNet` when `hostNetwork` is enabled (repoServer, server, and the application controller all follow the same conditional-hostNetwork + explicit-dnsPolicy pattern this PR uses), and for your bootstrap use case the flip would be actively wrong since there's no cluster DNS before the CNI exists. Silently coupling the two values would be the surprising behavior; users who want `ClusterFirstWithHostNet` can set it explicitly. @tico24, flagging so you can resolve your thread if that reasoning works for you.

The rerun request is moot now: main has moved to chart `10.4.2` (redis_exporter and argo-cd v3.5.2 bumps), so the branch conflicts and needs a rebase anyway. Please rebase and bump to `10.4.3` with your changelog entry; that will retrigger CI including the newline fix.

Implementation looks consistent and complete otherwise (redis already had `dnsPolicy`/`dnsConfig`, so this fills exactly the missing pieces). Ready to approve once the rebase is up and CI is green.

SkalaNetworks:
@jmeridth Hi, I bumped the chart to 10.4.3

jmeridth (approved):
Rebase to 10.4.3 confirmed and CI is fully green including chart-test. dnsPolicy reasoning settled earlier in the thread (consistent with repoServer/server/controller precedent, no auto-flip). LGTM, thanks @SkalaNetworks for the quick turnaround and the patience on the rerun.

mbevc1 (approved):

PR #3977, opened 2026-07-23
feat(argo-workflows): support Argo Workflows v4.1 features

Joibel:
Adds chart support for new configuration introduced in Argo Workflows v4.1.0

- `controller.initlessPod.enabled` — opt-in beta [init-less pod layout](https://argo-workflows.readthedocs.io/en/latest/initless-pod/) (argoproj/argo-workflows#16161). Documented with the `ImageVolume` feature-gate requirement (beta in Kubernetes v1.33-1.35, on by default from v1.36) and caveats.
- `controller.disableAgentPodCreation` — disable agent pods for HTTP and Plugin templates (argoproj/argo-workflows#15844).
- `artifactRepository.s3.addressingStyle` — virtual-hosted-style S3 support (argoproj/argo-workflows#15734).
- Make `persistence.postgresql.passwordSecret` optional in the controller and server ClusterRoles. The new Azure Entra (`azureToken`, argoproj/argo-workflows#15511) and AWS RDS IAM (`awsRDSToken`, argoproj/argo-workflows#15833) token authentication is used without a password secret, and the templates previously failed rendering with a nil pointer in that case. The persistence config itself passes through unchanged; commented examples added to values.yaml.
- Documentation: OpenTelemetry tracing setup via the existing `controller.extraEnv` / `executor.env` / `mainContainer.env` hooks (tracing is configured purely with `OTEL_*` environment variables upstream), and MariaDB as a supported archive database.

All new values default to off, so rendering is unchanged for existing users on v4.0.x.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

[...]

PR #4012, opened 2026-08-11
feat(argo-workflows): install full CRDs with the crdinstaller image

Joibel:
Switch the full-CRD install hook Job to the upstream `quay.io/argoproj/argo-workflows-crdinstaller` image (argoproj/argo-workflows#16622) instead of a stock kubectl image pointed at CRD URLs on raw.githubusercontent.com.

The image is kubectl plus the full CRDs for its own version, with `apply --server-side --force-conflicts -f /crds/full/` as its default command. So the Job needs no command, no CRD URLs and no GitHub egress, and the CRDs always match the version they're tagged with.

- `crds.upgradeJob.image` gains `registry` and its `tag` now defaults to the app version, like the other images
- `crds.upgradeJob.crdBaseURL` and `crds.upgradeJob.hostPath` are removed — `hostPath` only existed so CI could feed the Job from the kind nodes
- `runAsUser`/`runAsGroup` move to 8737 to match the image, and `/tmp` is an emptyDir because the image sets `HOME=/tmp` for kubectl's discovery cache
- the 11MB local copy of the full CRDs is dropped, along with the half of `update-argo-workflows-crds.sh` that synced it. Released charts pin their URLs to their own tag, where the files remain.

Checklist:

* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog)
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#dco)


tico24 (commented):
`quay.io/argoproj/argo-workflows-crdinstaller` only has `v4.1.0` and `latest` on it at the moment, so with appVersion still on v4.0.8 the Job asks for a tag that doesn't exist. I assume that's the CI failure you're warning about. #3977 is still sat on v4.0.8 and conflicting though, so is there a separate appVersion bump coming, or does this one pick it up?

Other thing: the old docs told people to override the kubectl image. Rendering with `--set crds.upgradeJob.image.repository=my-registry.internal/kubectl` now gets me `quay.io/my-registry.internal/kubectl:v1.36.3`, so anyone who did that has their pre-install hook fail on upgrade, and `crdBaseURL`/`hostPath` are silently ignored rather than erroring. Does that make it a major with an `Upgrading` section rather than 1.1.0? ([versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning))

Joibel:
> `quay.io/argoproj/argo-workflows-crdinstaller` only has `v4.1.0` and `latest` on it at the moment, so with appVersion still on v4.0.8 the Job asks for a tag that doesn't exist. I assume that's the CI failure you're warning about. #3977 is still sat on v4.0.8 and conflicting though, so is there a separate appVersion bump coming, or does this one pick it up?
> 
> Other thing: the old docs told people to override the kubectl image. Rendering with `--set crds.upgradeJob.image.repository=my-registry.internal/kubectl` now gets me `quay.io/my-registry.internal/kubectl:v1.36.3`, so anyone who did that has their pre-install hook fail on upgrade, and `crdBaseURL`/`hostPath` are silently ignored rather than erroring. Does that make it a major with an `Upgrading` section rather than 1.1.0? ([versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning))

Hello Tim's robot.

Slack chat says I'm expecting someone to do the usual automated bump PR to 4.1.0, and someone who know's what they're doing to tell me what they'd like the ordering of these to be. Failing finding someone who knows what they're doing, we can make it up as we go along with the argo-helm maintainers.

You tell me if you'd like this to be a v2.0.0 - happy to do that. I don't think continuing to support the old CRD installation mechanism is a sane path, and crdBaseURL and hostPath ignoring seem a fine thing to do given that.

tico24:
my bot leaked :(

tico24:
I do think it probably is a 2.0.0, which is annoying but that's how the numbers work.

Joibel:
> I do think it probably is a 2.0.0, which is annoying but that's how the numbers work.

Done. This is on top of #3977, so merge that first and then this will make sense again. Can't github stack them as is.

jmeridth (dismissed):

tico24 (approved):

jmeridth (approved):

PR #4014, opened 2026-08-11
fix(argo-cd): default HTTPRoute hostnames to global.domain when unset

### LINKED ISSUE #3972: [Documentation] [argocd] httproute example does not provide hostnames
### Describe the bug

In the chart documentation of argocd, the first examples of values to enable httproute creation do not include `hostnames` field.  
To be fair it actually works fine without if you do not have another route like that.  
But in order to avoid some basic routing issues I suggest to mention the `hostnames` field in the documentation.  
Another _fix_ could be to add the `global.domain` value to `hostnames` list by default if the list is empty.

### Related helm chart

argo-cd

### Helm chart version

10.1.2

### To Reproduce

1. Copy httproute config from documentation to the values file
2. deploy argo-cd with these values

### Expected behavior

We should have a working httproute regarding the provided route, and not a route that can be overtaken by another route.

### Screenshots

_No response_

### Additional context

_No response_

### LINKED ISSUE #3990: gateway(argo-cd): spec.hostnames may not set
### Describe the bug

When server.httproute and server.httpRoute.hostNames are not set, the spec.hostnames are not set in the generated HTTP route's YAML manifest.
The value of spec.hostnames should be set to the value of global.domain by default.

### Related helm chart

argo-cd

### Helm chart version

10.2.1

### To Reproduce

As above

### Expected behavior

The value of spec.hostnames should be set to the value of global.domain by default.

### Screenshots

_No response_

### Additional context

_No response_

lemon0333:
Defaults the argo-cd `server` and `applicationSet` HTTPRoute `spec.hostnames` to `global.domain` when `hostnames` are not set and no ListenerSet is enabled. Previously the rendered HTTPRoute omitted `spec.hostnames` entirely in that case. This mirrors the existing ListenerSet branch, which already falls back to `global.domain`.

Verified with `helm lint` and `helm template`:
- `hostnames` unset → `spec.hostnames: ["<global.domain>"]`
- `hostnames` set → uses the provided values (unchanged)
- omitted when `global.domain` is also empty

Closes #3990

Checklist:

Closes #3972

### DIFF
--- a/charts/argo-cd/templates/argocd-server/httproute.yaml
+++ b/charts/argo-cd/templates/argocd-server/httproute.yaml
@@ -33,6 +33,9 @@ spec:
   {{- else if .Values.server.listenerset.enabled }}
   hostnames:
     - {{ tpl (.Values.server.listenerset.hostname | default .Values.global.domain) $ | quote }}
+  {{- else if .Values.global.domain }}
+  hostnames:
+    - {{ tpl .Values.global.domain $ | quote }}
   {{- end }}
   rules:
   {{- range .Values.server.httproute.rules }}
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.6.4
+version: 10.6.5
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
-      description: Hash only the data/stringData of ConfigMaps and Secrets in checksum/* pod annotations so a chart version bump alone no longer restarts all Argo CD pods. Adopting this release recomputes every checksum/* annotation once, so all Argo CD pods roll a single time on this specific upgrade; subsequent version bumps no longer roll them.
+      description: HTTPRoute now defaults spec.hostnames to global.domain when httproute.hostnames is unset. This changes behavior — an omitted hostnames previously matched all hostnames of the bound listener, and now narrows to global.domain (default placeholder argocd.example.com). To restore match-all, set httproute.hostnames explicitly or set global.domain to "".
--- a/charts/argo-cd/templates/argocd-applicationset/httproute.yaml
+++ b/charts/argo-cd/templates/argocd-applicationset/httproute.yaml
@@ -31,6 +31,9 @@ spec:
   {{- else if .Values.applicationSet.listenerset.enabled }}
   hostnames:
     - {{ tpl (.Values.applicationSet.listenerset.hostname | default .Values.global.domain) $ | quote }}
+  {{- else if .Values.global.domain }}
+  hostnames:
+    - {{ tpl .Values.global.domain $ | quote }}
   {{- end }}
   rules:
     {{- range .Values.applicationSet.httproute.rules }}


jmeridth:
@lemon0333 Triaged this today. The change is directionally right, it's what #3990 asked for, and it matches the ingress and ListenerSet conventions. The implementation is clean. One substantive ask and one mechanical one before approval:

**1. Reframe the changelog: this is `kind: changed`, not `fixed`.** In Gateway API, an omitted `spec.hostnames` is meaningful config: the route matches all hostnames of the bound listener, letting the listener own hostname policy. And since `global.domain` defaults to the placeholder `argocd.example.com`, your new branch is effectively always taken. Any existing user who deliberately omitted `hostnames` will have their route silently narrowed to `argocd.example.com` on upgrade and stop matching real traffic. The escape hatches exist (set `httproute.hostnames` explicitly, or `global.domain: ""`), but nobody will find them behind a changelog line that says "fixed". Please make the entry `kind: changed` and spell out the new default plus how to restore match-all behavior.

**2. Rebase.** Main has moved to `10.4.2`, so bump to `10.4.3` with the updated changelog entry.

For most users who've set `global.domain` to their real domain this default is exactly what they wanted, so I'm fine shipping it without a major bump as long as the changelog is honest about the behavior change. Ready to approve once both are pushed.

lemon0333:
Thanks for the review, @jmeridth!

- Rebased onto `main` (chart was at `10.6.1`) to clear the conflicts.
- Bumped `charts/argo-cd` to **`10.6.4`** (skipping `10.6.2`/`10.6.3`/`10.7.0` to avoid colliding with sibling PRs in flight).
- Reframed the changelog from `kind: fixed` to `kind: changed`, since this is a behavior change rather than a bugfix. The single entry now spells it out:

  > HTTPRoute now defaults `spec.hostnames` to `global.domain` when `httproute.hostnames` is unset. This changes behavior — an omitted `hostnames` previously matched all hostnames of the bound listener, and now narrows to `global.domain` (default placeholder `argocd.example.com`). To restore match-all, set `httproute.hostnames` explicitly or set `global.domain` to `""`.

`helm lint charts/argo-cd` passes clean. No `values.yaml` changes, so the README is unchanged.

jmeridth:
@lemon0333 one more time on the Chart.yaml. sorry

lemon0333:
@jmeridth No worries — re-rebased onto current `main` (now argo-cd chart 10.6.3 after the sibling PR merged). Bumped `version` to **10.6.5**, and the `Chart.yaml` conflict is cleared. The changelog entry (`kind: changed`, HTTPRoute hostnames) is unchanged, and remains the single `artifacthub.io/changes` entry. Force-pushed the rebased branch.

charts/argo-cd/templates/argocd-server/httproute.yaml
```diff
@@ -33,6 +33,9 @@ spec:
   {{- else if .Values.server.listenerset.enabled }}
   hostnames:
     - {{ tpl (.Values.server.listenerset.hostname | default .Values.global.domain) $ | quote }}
+  {{- else if .Values.global.domain }}
```
tico24, line 36: Doesn't this break anyone sitting on the default `global.domain`? With `server.httproute.enabled=true` and nothing else set, the route goes from no `hostnames` at all, which matches every hostname on the listener, to `hostnames: ["argocd.example.com"]`. Your changelog says as much, so should this be a major with an `Upgrading` section rather than a patch? ([versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning))

lemon0333:
Re-rebased onto `main` again (it moved to 10.6.4 after the sibling merge) — Chart.yaml conflict cleared, version stays **10.6.5**, `kind: changed` changelog entry unchanged.

jmeridth:
@tico24 agreeing with you, and revising my earlier position on the record: I had accepted `kind: changed` with an honest changelog as sufficient, but you're right that by [our versioning convention](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning) a default that turns a match-all route into `hostnames: ["argocd.example.com"]` for anyone sitting on the default `global.domain` is a breaking change, and breaking changes mean a major with an `Upgrading` section, not a patch.

@lemon0333 two asks, which conveniently collapse into one rebase since #4033 took `10.7.0`:

1. Re-target to **`11.0.0`**.
2. Add an `## Upgrading to 11.0.0` entry to the chart README (gotmpl) covering: HTTPRoutes with `hostnames` unset now default to `global.domain`; to restore match-all behavior set `httproute.hostnames` explicitly or `global.domain: ""`. The changelog text you already wrote is the right content, it just also belongs in the Upgrading section.

The implementation itself needs no changes. Sorry for the moving target on versions this week; the merge queue has been busy.

PR #4018, opened 2026-08-13
fix(argo-cd): add app.kubernetes.io/part-of: argocd to secretTem…

### LINKED ISSUE #4017: regression with version 3.5.0+ of argocd application: when using cert-manager and server.certificate.enabled no longer hot reloads on certificate rotation, causing the argocd-server to present an expired certificate
### Describe the bug

It seems that in a recent change(https://github.com/argoproj/argo-cd/pull/27213) the hot reloading of secrets logic has been changed to prevent excessive reloading. This seems to  have had the unintended side effect that the combination of using argo-cd helm chart in combination with it's own `server.certificate.enabled` value to no longer be hot reloaded on rotation.

We've encountered this in our setup because we have a rotation speed of once every 48h and after 48h the argocd server would respond with the previous certificate that had bin expired. Manually restarting the argocd server would resolve the issue.

As a workaround we used the `extraObjects` value to add a certificate object but wit the following added:

```yaml
...
    spec:
      ...
      secretName: argocd-server-tls
      secretTemplate:
        labels:
```

When this lable is present on the generated secret, hot reloading works again and we don't see this error anymore.

Suggestion:

add this per default or make secretTemplate a configurable option so that the shipped `server.certificate` values become useful again for fast rotating secrets..

### Related helm chart

argo-cd

### Helm chart version

10.3.3

### To Reproduce

1: Deploy using the shipped certificate logic

```yaml
  server:
   
    certificate:
      enabled: true

      additionalHosts: [ ]
  
      duration: "48h"
      issuer:
        kind: ClusterIssuer
        name: my-issuer

    
      privateKey:
        rotationPolicy: Always
        encoding: PKCS1
        algorithm: RSA
        size: 4096
      annotations:
        cert-manager.io/issue-temporary-certificate: "true"
      usages:
      - server auth
```

2: Delete the generate secret associated with the certificate to force cert-manager of issuing a new one.
3: verify in browser that argocd-server is presenting the old certificate.

### Expected behavior

The shipped certificate logic should be hot reloaded like all other argocd secrets:
```yaml
labels:
```

### Screenshots

_No response_

### Additional context

_No response_

martijnhybrit:
…plate.label of the server.certificate generated certificate so hot reloading works on certificate rotation

Checklist:

Closes #4017

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.7.1
+version: 10.7.2
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: security
-      description: Bump the redis-ha Redis image to 8.6.4-alpine so it no longer ships an OpenSSL version affected by the RSASVE uninitialized-memory issue (reported by m98anu)
+    - kind: fixed
+      description: "add app.kubernetes.io/part-of: argocd to secretTemplate.label of the server.certificate generated certificate so hot reloading works on certificate rotation"
--- a/charts/argo-cd/templates/argocd-server/certificate.yaml
+++ b/charts/argo-cd/templates/argocd-server/certificate.yaml
@@ -13,8 +13,10 @@ metadata:
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.server.name "name" .Values.server.name) | nindent 4 }}
 spec:
-  {{- with .Values.server.certificate.secretTemplateAnnotations }}
   secretTemplate:
+    labels:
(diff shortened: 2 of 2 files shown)


charts/argo-cd/Chart.yaml
```diff
@@ -22,6 +22,9 @@ dependencies:
     repository: https://dandydeveloper.github.io/charts/
     condition: redis-ha.enabled
 annotations:
+  artifacthub.io/changes: |
+    - kind: fixed
+      description: add app.kubernetes.io/part-of: argocd to secretTemplate.label of the server.certificate generated certificate so hot reloading works on certificate rotation
```
yu-croco: Pls modify the original ones, in line 31-33.

jmeridth:
@martijnhybrit Triaged this today. First, the substance: the fix is correct, and I verified it against upstream source so nobody has to re-litigate it. Argo CD's settings manager lists and watches secrets via the selector `app.kubernetes.io/part-of=argocd` (`util/settings/settings.go`, `partOfArgoCDSelector`, defined right next to the `argocd-server-tls` constant), so a cert-manager-generated secret without that label is invisible to the informer and rotated certs go stale until pod restart. Making the label unconditional in `secretTemplate` is the right call, and scoping it to the server cert only is also correct since the applicationSet webhook cert is consumed as a mounted volume, not through the settings manager.

Three mechanical items are blocking it:

1. **Address @yu-croco's review comment.** Your change added a second `artifacthub.io/changes` block at the top of the annotations; Chart.yaml already has one further down. That's a duplicate YAML map key. Replace the content of the existing entry instead of adding a new block.
2. **Quote the changelog description.** The text contains `app.kubernetes.io/part-of: argocd`, and the embedded colon-space breaks YAML parsing of the changes block. Wrap the description in quotes or reword it.
3. **Rebase.** Main has moved to chart `10.4.2`, so rebase and bump to `10.4.3`.

Once those three are pushed this is ready to approve. Thanks for the fix; it closes a real gap in the `server.certificate` feature.

tico24 (approved):

mbevc1 (approved):

PR #4019, opened 2026-08-14
fix(argo-rollouts): remove default controller readinessProbe to align with upstream

mitsu3s:
Upstream removed the controller readinessProbe from the install manifests in argoproj/argo-rollouts#4704 (released in v1.9.1), since probing `/metrics` is expensive and process health is already covered by the liveness probe on `/healthz`. This PR aligns the chart with upstream: the default `controller.readinessProbe` is removed, and the probe is rendered only when the value is set.

Users who want the previous behavior can set `controller.readinessProbe` explicitly (partial overrides now require the full probe definition).

Checklist:

### DIFF
--- a/charts/argo-rollouts/values.yaml
+++ b/charts/argo-rollouts/values.yaml
@@ -210,6 +210,9 @@ controller:
     successThreshold: 1
     timeoutSeconds: 10
 
+  # -- Configure readiness [probe] for the controller
+  readinessProbe: {}
+
   ## Configure Pod Disruption Budget for the controller
   pdb:
     # -- Labels to be added to controller [Pod Disruption Budget]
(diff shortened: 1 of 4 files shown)


jmeridth (dismissed):

charts/argo-rollouts/values.yaml
```diff
@@ -210,18 +210,6 @@ controller:
     successThreshold: 1
     timeoutSeconds: 10
 
-  # -- Configure readiness [probe] for the controller
-  # @default -- See [values.yaml]
-  readinessProbe:
```
yu-croco, line 215: nits; I wonder that we can leave the minimum interface since we provide it at `.Values.controller.readinessProbe` in `charts/argo-rollouts/templates/controller/deployment.yaml`. 🤔 
```yaml
# -- Configure readiness [probe] for the controller
readinessProbe {}
```

mitsu3s: Good point, thank you! Kept `readinessProbe: {}` as a documented placeholder.

yu-croco (approved):

jmeridth (approved):

PR #4021, opened 2026-08-15
chore(argo-cd): Add ReferenceGrant for Gateway API

yanjunding:
Checklist:

### DIFF
--- a/charts/argo-cd/templates/argocd-server/certificate.yaml
+++ b/charts/argo-cd/templates/argocd-server/certificate.yaml
@@ -47,4 +47,19 @@ spec:
   usages:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+---
+{{- if and .Values.server.certificate.referencegrant.enabled (and .Values.server.httproute.enabled .Values.server.httproute.parentRefs) -}}
+apiVersion: gateway.networking.k8s.io/v1
+kind: ReferenceGrant
+metadata:
+  name: {{ include "argo-cd.server.fullname" . }}
+  namespace: {{ include  "argo-cd.namespace" . }}
+spec:
+  from:
+  {{- toYaml .Values.server.httproute.parentRefs | nindent 4 }}
+  to:
+  - group: ""
+    kind: Secret
+    name: argocd-server-tls
+{{- end }}
 {{- end }}
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -2439,6 +2439,9 @@ server:
     usages: []
     # -- Annotations that allow the certificate to be composed from data residing in existing Kubernetes Resources
     secretTemplateAnnotations: {}
+    # ReferenceGrant for Gateway API
+    referencegrant:
+      enabled: false
 
   # TLS certificate configuration via Secret
   ## Ref: https://argo-cd.readthedocs.io/en/stable/operator-manual/tls/#tls-certificates-used-by-argocd-server
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -27,4 +27,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: changed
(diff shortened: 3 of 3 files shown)


jmeridth:
@yanjunding Thanks for the contribution, and apologies for the two-week silence. The feature is in scope: the chart has HTTPRoute/ListenerSet support, and a Gateway in another namespace genuinely needs a ReferenceGrant to read `argocd-server-tls`. But the implementation has a correctness bug that needs a redesign rather than polish, so requesting changes.

**The core issue: `spec.from` is built from the wrong object.** The template dumps `server.httproute.parentRefs` verbatim into the ReferenceGrant's `from`. Those schemas are incompatible: `ReferenceGrantFrom` allows exactly three fields, all required (`group`, `kind`, `namespace`), while a `parentRef` requires `name` (forbidden in ReferenceGrant) and treats `group`/`kind`/`namespace` as optional with implicit defaults. A typical parentRef like `- name: my-gateway, namespace: gateway-ns` renders a ReferenceGrant with an illegal `name` field and missing required `group`/`kind`. It passes `helm template` but the API server rejects it at apply time. Also note a parentRef without `namespace` targets the release namespace, where no grant is needed at all, another sign parentRefs is the wrong source.

**Suggested shape:**

1. Give it a dedicated template (`templates/argocd-server/referencegrant.yaml`) instead of appending inside `certificate.yaml`. As written, a stray empty `---` document renders whenever the certificate is enabled but the grant is not, and the coupling to `server.certificate.enabled` locks out users who bring their own `argocd-server-tls` secret without cert-manager but still have a cross-namespace Gateway.
2. Make the config explicit, e.g. a `referenceGrant` block where the user lists Gateway namespaces, rendered as proper `{group: gateway.networking.k8s.io, kind: Gateway, namespace: <x>}` entries.
3. Mechanical items: camelCase the value (`referenceGrant`), add the `# --` helm-docs comment marker so the value is documented, regenerate the README (`./scripts/helm-docs.sh`), and rebase (main is at chart `10.4.2`).

For what it's worth, hardcoding `argocd-server-tls` in `to:` is fine; the Certificate template hardcodes the same secretName.

Happy to review the reworked version.

jmeridth (changes requested):
Need changes

charts/argo-cd/templates/argocd-server/certificate.yaml
```diff
@@ -47,4 +47,19 @@ spec:
...
+---
+{{- if and .Values.server.certificate.referencegrant.enabled (and .Values.server.httproute.enabled .Values.server.httproute.parentRefs) -}}
+apiVersion: gateway.networking.k8s.io/v1
+kind: ReferenceGrant
+metadata:
+  name: {{ include "argo-cd.server.fullname" . }}
+  namespace: {{ include  "argo-cd.namespace" . }}
+spec:
+  from:
+  {{- toYaml .Values.server.httproute.parentRefs | nindent 4 }}
```
tico24, line 59: Rendered this with `referencegrant.enabled=true` and a `parentRefs` entry with a `sectionName`, and the grant comes out with `from: [{name, namespace, sectionName}]`. ReferenceGrant's `from` only has `group`, `kind` and `namespace`, and the first two are required, so the API server rejects it. Shouldn't `from` be the Gateway rather than a copy of `parentRefs`?

charts/argo-cd/values.yaml
```diff
@@ -2439,6 +2439,9 @@ server:
     usages: []
     # -- Annotations that allow the certificate to be composed from data residing in existing Kubernetes Resources
     secretTemplateAnnotations: {}
+    # ReferenceGrant for Gateway API
```
tico24, line 2442: The new value needs a `# --` comment or helm-docs won't give it a description, and `README.md` needs regenerating with `./scripts/helm-docs.sh`. You're also on 10.3.4 where main's on 10.7.1, so it wants a rebase and a re-bump. None of that is showing red because the chart checks haven't run on this PR yet. ([contributing](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md))

PR #4024, opened 2026-08-17
fix(argo-events): RoleBinding missing metadata.namespace under controller.rbac.namespaced

### LINKED ISSUE #4023: argo-events chart 2.4.24: controller.rbac.namespaced RoleBinding template missing metadata.namespace
When setting `controller.rbac.namespaced: true` with `controller.rbac.managedNamespace: <ns>`
on the argo-events chart (tested at chart version 2.4.24), the rendered `RoleBinding` for the
controller has no `metadata.namespace` field. `kubectl apply` (and ArgoCD's own Kustomize/Helm
render+apply) then places the object in whatever namespace the apply context defaults to,
not the namespace actually managed by the Role — in our case it landed in `default` instead
of the target `argo-events` namespace.

Effect observed: the controller pod itself comes up `Running`, but every `list`/`watch` call
against its own CRs (`EventBus`, `EventSource`, `Sensor`) and against core resources in the
intended namespace fails with `forbidden ... in the namespace "<intended-ns>"`, because the
RoleBinding granting those permissions is silently sitting in the wrong namespace instead.
`kubectl get rolebinding -n <intended-ns>` finds nothing; `-n default` (or whatever the apply
context's default namespace is) finds the misplaced object.

Expected: the `RoleBinding` template should set `metadata.namespace` to
`.Values.controller.rbac.managedNamespace` (or the release namespace, matching whatever the
paired `Role` template already does) when namespaced RBAC is enabled, the same way the `Role`
template itself appears to.

Workaround used: reverted to the chart's default cluster-wide `ClusterRole`/
`ClusterRoleBinding` instead of the namespaced option — not filing a PR since we haven't
narrowed the exact template line, just a live-verified defect report.

landif:
## What

`charts/argo-events/templates/argo-events-controller/rbac.yaml`'s `Role`/`ClusterRole`
block correctly conditions `metadata.namespace` on `controller.rbac.namespaced`:

```yaml
metadata:
  name: {{ include "argo-events.controller.fullname" . }}
  {{- if .Values.controller.rbac.namespaced }}
  namespace: {{ include "argo-events.namespace" . | quote }}
  {{- end }}
```

The sibling `RoleBinding`/`ClusterRoleBinding` block right below it has no such condition
at all — `metadata.namespace` is simply absent, in both modes.

## Impact

With `controller.rbac.namespaced: true`, the rendered `RoleBinding` has no
`metadata.namespace`, so `kubectl apply`/`helm install` places it wherever the apply
context's own default namespace is, not the chart's release namespace. The controller pod
still comes up `Running`, but every `list`/`watch` call it makes against its own CRs
(`EventSource`, `Sensor`, `EventBus`) and core resources fails `forbidden … in the
namespace "<release-namespace>"`, since the RoleBinding granting those permissions isn't
actually bound in that namespace.

Found running this chart (`v2.4.24`) in production with the namespaced RBAC option enabled,
as a lower-privilege alternative to the chart's default cluster-wide `ClusterRole`/
`ClusterRoleBinding`.

## Fix

Add the identical conditional already used one block above, to the `RoleBinding`/
`ClusterRoleBinding` block's `metadata`.

## Verified

```
$ helm template test . --namespace argo-events --set controller.rbac.namespaced=true -s templates/argo-events-controller/rbac.yaml
kind: RoleBinding
metadata:
  name: test-argo-events-controller-manager
  namespace: "argo-events"
  ...
```

Default (cluster-wide) mode confirmed unaffected — `ClusterRoleBinding` still renders with
no `metadata.namespace`, as expected for a cluster-scoped object.

Chart version bumped `2.4.24` → `2.4.25` per `CONTRIBUTING.md`.

Closes #4023.

### DIFF
--- a/charts/argo-events/Chart.yaml
+++ b/charts/argo-events/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v1.9.11
 description: A Helm chart for Argo Events, the event-driven workflow automation framework
 name: argo-events
-version: 2.4.24
+version: 2.4.25
 home: https://github.com/argoproj/argo-helm
 icon: https://avatars.githubusercontent.com/u/30269780?s=200&v=4
 keywords:
--- a/charts/argo-events/templates/argo-events-controller/rbac.yaml
+++ b/charts/argo-events/templates/argo-events-controller/rbac.yaml
@@ -113,6 +113,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: {{ .Values.controller.rbac.namespaced | ternary "RoleBinding" "ClusterRoleBinding" }}
 metadata:
   name: {{ include "argo-events.controller.fullname" . }}
+  {{- if .Values.controller.rbac.namespaced }}
+  namespace: {{ include "argo-events.namespace" . | quote }}
+  {{- end }}
   labels:
     {{- include "argo-events.labels" (dict "context" . "component" .Values.controller.name "name" .Values.controller.name) | nindent 4 }}
 roleRef:


jmeridth:
@landif Triaged this today. The fix is verified and wanted: the Role/ClusterRole block conditions `metadata.namespace` on `controller.rbac.namespaced` while the sibling RoleBinding block omits it entirely, so with `namespaced: true` the binding lands wherever the apply context points (bites with `helm template | kubectl apply` or `namespaceOverride`). Your three-line change mirrors the Role block exactly, which is the right shape.

Two things to get it over the line:

1. **DCO**: your commit isn't signed off, so the DCO check fails. `git commit --amend -s` and force-push (the check's *Details* link has the full instructions).
2. **Mark the PR ready for review** when you consider it complete; it's still in draft, so it isn't in anyone's review queue.

Once those are done and chart-test runs green, this is ready to approve.

jmeridth (dismissed):

charts/argo-events/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v1.9.11
 description: A Helm chart for Argo Events, the event-driven workflow automation framework
 name: argo-events
-version: 2.4.24
+version: 2.4.25
 home: https://github.com/argoproj/argo-helm
 icon: https://avatars.githubusercontent.com/u/30269780?s=200&v=4
 keywords:
```
tico24, line 8: Missing a changelog here

tico24:
Please put the checkboxes back into your PR description and then do the things they say. thanks

PR #4029, opened 2026-08-21
fix(argo-rollouts): harden dashboard security context

WoahitsJoe:
## Summary

- prevent privilege escalation in the dashboard container by default
- drop all Linux capabilities from the dashboard container
- use a read-only root filesystem and the RuntimeDefault seccomp profile
- rebase onto current main and bump the argo-rollouts chart to 2.43.1
- regenerate documentation

This aligns the optional dashboard with the Kubernetes Restricted Pod Security Standard controls already configured for the controller while preserving value overrides.

## Testing

- `helm lint charts/argo-rollouts -f charts/argo-rollouts/ci/enable-dashboard-values.yaml`
- rendered the dashboard and verified the complete default security context, including `readOnlyRootFilesystem: true`
- the repository-wide `./scripts/lint.sh` requires Docker, which is unavailable in the current WSL environment; chart CI will run it

## Checklist

* [x] I have bumped the chart version according to versioning
* [x] I have updated the documentation according to documentation
* [x] I have updated the chart changelog with all changes in this pull request
* [x] Any new values are backwards compatible and/or have sensible defaults
* [x] I have signed off all commits as required by DCO
* [x] I have created a separate pull request for this chart
* [x] My local chart lint and render are green

### DIFF
--- a/charts/argo-rollouts/values.yaml
+++ b/charts/argo-rollouts/values.yaml
@@ -394,7 +394,14 @@ dashboard:
   podSecurityContext:
     runAsNonRoot: true
   # -- Security Context to set on container level
-  containerSecurityContext: {}
+  containerSecurityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+        - ALL
+    readOnlyRootFilesystem: true
+    seccompProfile:
+      type: RuntimeDefault
   service:
     # -- Sets the type of the Service
     type: ClusterIP
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v1.10.0
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 2.43.0
+version: 2.43.1
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-rollouts/assets/logo.png
 keywords:
@@ -18,5 +18,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump argo-rollouts to v1.10.0
+    - kind: security
+      description: Harden the dashboard container security context by default
--- a/charts/argo-rollouts/README.md
+++ b/charts/argo-rollouts/README.md
@@ -156,7 +156,7 @@ For full list of changes please check ArtifactHub [changelog].
 |-----|------|---------|-------------|
 | dashboard.affinity | object | `{}` | Assign custom [affinity] rules to the deployment |
 | dashboard.component | string | `"rollouts-dashboard"` | Value of label `app.kubernetes.io/component` |
-| dashboard.containerSecurityContext | object | `{}` | Security Context to set on container level |
+| dashboard.containerSecurityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"readOnlyRootFilesystem":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security Context to set on container level |
 | dashboard.createClusterRole | bool | `true` | flag to enable creation of dashbord cluster role (requires cluster RBAC) |
 | dashboard.deploymentAnnotations | object | `{}` | Annotations to be added to the dashboard deployment |
 | dashboard.deploymentLabels | object | `{}` | Labels to be added to the dashboard deployment |


jmeridth:
@WoahitsJoe Triaged this today; apologies for the eleven-day silence. The direction is right and verified: upstream's `dashboard-install.yaml` sets no securityContext at all, so aligning the optional dashboard with the chart's existing controller hardening is the correct precedent, and the runtime risk is negligible for a Go web server. `kind: security` is also the right changelog label. A bonus in your favor: the chart's CI includes `ci/enable-dashboard-values.yaml`, so chart-test actually installs the dashboard with the new context rather than just linting it.

Two asks before approval:

1. **Complete the parity claim.** The controller's default `containerSecurityContext` also sets `readOnlyRootFilesystem: true`, which this PR omits for the dashboard. Either add it (the dashboard install test in CI will prove the container tolerates it) or note in the PR why it's excluded. As written, "aligns with the controls already configured for the controller" slightly oversells the diff.

2. **Rebase.** Main has moved to chart `2.42.0`, so bump to `2.42.1` with your changelog entry. Heads up that #4041 (argo-rollouts v1.10.0) may take `2.43.0` shortly, in which case you'd land on `2.43.1` instead.

Ready to approve once those are pushed.

WoahitsJoe:
@jmeridth Thanks! Addressed both asks: added `readOnlyRootFilesystem: true` for dashboard parity, rebased onto current main, and bumped the chart to `2.43.1`. 
Chart lint and dashboard rendering pass locally.

jmeridth (approved):
Both asks delivered: `readOnlyRootFilesystem: true` brings the dashboard to full parity with the controller's default containerSecurityContext, and the rebase to 2.43.1 sits cleanly on the v1.10.0 release. The parity claim in the description is now accurate. Worth noting the CI run is real verification here: chart-test installs with `ci/enable-dashboard-values.yaml`, so the dashboard container booted under the hardened context including the read-only root filesystem. Checklist and changelog complete. LGTM, thanks @WoahitsJoe.

charts/argo-rollouts/values.yaml
```diff
@@ -394,7 +394,14 @@ dashboard:
   podSecurityContext:
     runAsNonRoot: true
   # -- Security Context to set on container level
-  containerSecurityContext: {}
+  containerSecurityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+        - ALL
+    readOnlyRootFilesystem: true
```
tico24, line 402: Has the dashboard actually been run with this, rather than just rendered? It comes out with `readOnlyRootFilesystem: true` and no volumes or volumeMounts at all, where the controller you're matching pairs the same block with `tmp` and `plugin-bin` emptyDirs.

mkilchhofer: I thought about the same potential issue.
Maybe drop the `readOnlyRootFilesystem: true` for now?

jmeridth: @tico24 @mkilchhofer fair challenge, and there is direct evidence to answer it: chart-test on this PR doesn't just render — `ct install` runs `helm install --wait` for every `ci/*-values.yaml`, which includes `enable-dashboard-values.yaml`, so the dashboard Deployment was actually created in kind under the full hardened context including `readOnlyRootFilesystem: true` and reached Ready before the green result: https://github.com/argoproj/argo-helm/actions/runs/33591747234/job/100133744371

That also matches expectations from the binary's behavior: the dashboard is a static-serving Go server that doesn't write to its filesystem at runtime, unlike the controller, whose `tmp`/`plugin-bin` emptyDirs exist because it downloads plugin binaries.

If you'd still rather land the uncontroversial 90% now and take RORF separately with longer soak time, dropping that one line is fine by me and my approval stands either way — but the "has it run" question specifically has a yes.

WoahitsJoe:
I ran an additional local runtime test against the exact PR commit (55e4b5a) in a temporary kind cluster.

Test setup:

  - Installed the chart using ci/enable-dashboard-values.yaml
  - Enabled the chart CRDs for functional Rollout testing
  - Confirmed the live dashboard container had:
      - allowPrivilegeEscalation: false
      - All capabilities dropped
      - **readOnlyRootFilesystem: true**
      - seccompProfile: RuntimeDefault
      - No writable volume mounts

Runtime/application checks:

  - Dashboard Deployment reached Ready
  - /rollouts/, its JavaScript, and CSS returned HTTP 200
  - Version, namespace, Rollout list/detail, and watch APIs worked
  - Created Healthy and Paused sample Rollouts
  - Confirmed both appeared correctly in the dashboard
  - Navigated around the UI and promoted the paused canary successfully
  - Ran 120 repeated page, asset, and API requests

Final state: both Rollouts were fully available, and the dashboard remained Ready with zero restarts and no application errors. This provides additional confirmation that the dashboard works normally with readOnlyRootFilesystem: true, beyond simply rendering or starting the container.

charts/argo-rollouts/values.yaml
tico24: @mkilchhofer WoahitsJoe ran it in kind after you posted, up the page rather than in here: dashboard Ready with RORF on, APIs served, a canary promoted, no restarts. Does that cover it for you?

yu-croco (dismissed):

charts/argo-rollouts/values.yaml
WoahitsJoe: @mkilchhofer @tico24  Just checking to make sure this doesn't get stale. Is this good to go?

mkilchhofer (approved):
LGTM
Thank you @WoahitsJoe for the contribution and for your patience 🙏 .

PR #4031, opened 2026-08-22
fix(argo-cd): grant argocd-server impersonate on serviceaccounts in default RBAC

### LINKED ISSUE #4028: Missing `impersonate` verb for `serviceaccounts` in default server clusterRoleRules (v3.5.x / Chart v10.4.0)
### Describe the bug

After upgrading Argo CD to version `v3.5.1` (using helm chart `v10.4.0`), users encounter a `forbidden` error when attempting to delete resources via the Argo CD UI or API. 

This happens because Argo CD v3.5 extended the application impersonation capabilities to server operations. The `argocd-server` now attempts to impersonate the application controller's ServiceAccount during deletion/management tasks. However, the default ClusterRole generated by the Helm Chart does not grant the `argocd-server` the permission to impersonate `serviceaccounts`.

### Related helm chart

argo-cd

### Helm chart version

v10.4.0

### To Reproduce

1. Deploy Argo CD using helm chart `v10.4.0` (Argo CD app version `v3.5.1`) with default RBAC settings (`server.clusterRoleRules.enabled` set to `false`).
2. Log into the Argo CD UI.
3. Attempt to delete a live resource managed by Argo CD, for example pod.
4. The deletion fails with the RBAC impersonation error.

### Expected behavior

The default `ClusterRole` (or namespace-scoped `Role`) created for `argocd-server` should automatically include the `impersonate` verb for `serviceaccounts` to support the architectural changes introduced in Argo CD v3.5.x.

### Screenshots

Image

### Additional context

I think that adding following rule to the default server role templates (both ClusterRole and Role) should fix the problem:

```yaml
  - apiGroups:
      - ""
    resources:
      - serviceaccounts
    verbs:
      - impersonate
```

lemon0333:
Fixes #4028

## What this PR does

Argo CD v3.5 extended application impersonation to server operations: `argocd-server` now impersonates the application controller's ServiceAccount during resource delete/manage. The chart's default server RBAC never granted the `impersonate` verb on `serviceaccounts`, so deleting a live resource from the Argo CD UI/API fails with a `forbidden` error (reported against app `v3.5.1` / chart `v10.4.0`).

This adds the following rule to the default server RBAC templates:

```yaml
- apiGroups:
    - ""
  resources:
    - serviceaccounts
  verbs:
    - impersonate
```

- Added to the default server **ClusterRole** (`templates/argocd-server/clusterrole.yaml`), guarded under the existing default-rules path (`server.clusterRoleRules.enabled=false`) so it does not affect users who supply their own `server.clusterRoleRules.rules`.
- Added to the namespace-scoped server **Role** (`templates/argocd-server/role.yaml`).

No values changed, so this is backwards compatible with no configuration required.

Checklist:

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.8.1
+version: 10.9.0
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump argocd-extension-installer to v1.1.0
+    - kind: added
+      description: Add opt-in server.impersonation.enabled/serviceAccounts to grant argocd-server a scoped (resourceNames) impersonate rule for the Argo CD v3.5+ app-sync impersonation feature; default off leaves default RBAC unchanged
--- a/charts/argo-cd/templates/argocd-server/clusterrole.yaml
+++ b/charts/argo-cd/templates/argocd-server/clusterrole.yaml
@@ -17,6 +17,18 @@ rules:
       - delete  # supports deletion a live object in UI
       - get     # supports viewing live object manifest in UI
       - patch   # supports `argocd app patch`
+  {{- if .Values.server.impersonation.enabled }}
+  - apiGroups:
+      - ""
+    resources:
+      - serviceaccounts
+    {{- with .Values.server.impersonation.serviceAccounts }}
+    resourceNames:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
+    verbs:
+      - impersonate  # opt-in: app-sync impersonation feature (Argo CD v3.5+); scoped to server.impersonation.serviceAccounts
+  {{- end }}
   - apiGroups:
       - ""
     resources:
--- a/charts/argo-cd/README.md
+++ b/charts/argo-cd/README.md
@@ -1394,6 +1394,8 @@ NAME: my-release
 | server.image.repository | string | `""` (defaults to global.image.repository) | Repository to use for the Argo CD server |
 | server.image.tag | string | `""` (defaults to global.image.tag) | Tag to use for the Argo CD server |
 | server.imagePullSecrets | list | `[]` (defaults to global.imagePullSecrets) | Secrets with credentials to pull images from a private registry |
+| server.impersonation.enabled | bool | `false` | Grant argocd-server the `impersonate` verb on serviceaccounts for the app-sync impersonation feature (Argo CD v3.5+). Only enable if you use application sync impersonation; leaving it off keeps the default RBAC unchanged. |
+| server.impersonation.serviceAccounts | list | `[]` | ServiceAccount names argocd-server may impersonate. Rendered as `resourceNames` to scope the grant; leaving it empty allows impersonating any ServiceAccount (NOT recommended). |
 | server.ingress.annotations | object | `{}` | Additional ingress annotations |
 | server.ingress.aws.backendProtocolVersion | string | `"GRPC"` | Backend protocol version for the AWS ALB gRPC service |
 | server.ingress.aws.serviceAnnotations | object | `{}` | Annotations for the AWS ALB gRPC service |
--- a/charts/argo-cd/templates/argocd-server/role.yaml
+++ b/charts/argo-cd/templates/argocd-server/role.yaml
@@ -33,6 +33,18 @@ rules:
   - update
   - delete
   - patch
+{{- if .Values.server.impersonation.enabled }}
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  {{- with .Values.server.impersonation.serviceAccounts }}
+  resourceNames:
+    {{- toYaml . | nindent 2 }}
+  {{- end }}
+  verbs:
+  - impersonate  # opt-in: app-sync impersonation feature (Argo CD v3.5+); scoped to server.impersonation.serviceAccounts
+{{- end }}
 - apiGroups:
   - ""
   resources:
--- a/charts/argo-cd/values.yaml
+++ b/charts/argo-cd/values.yaml
@@ -3035,6 +3035,20 @@ server:
     # -- List of custom rules for the server's ClusterRole resource
     rules: []
 
+  ## Opt-in impersonation RBAC for the Argo CD v3.5+ app-sync-using-impersonation feature
+  ## (`application.sync.impersonation.enabled` in argocd-cm). When enabled, argocd-server
+  ## impersonates a tenant ServiceAccount during resource delete/manage triggered from the UI/API.
+  ## Default OFF: default installs never impersonate, so default RBAC grants nothing.
+  ## Keep this least-privilege — always set `serviceAccounts` so the grant is scoped via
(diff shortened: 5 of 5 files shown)


jmeridth:
@lemon0333 Triaged this today, and this is the one PR of your recent batch where I have to push back on the approach rather than the mechanics. The bug in #4028 is real, but this fix can't ship as written.

**What I verified upstream:** argo-cd v3.5.1's `install.yaml` contains zero `impersonate` RBAC grants; the two matches are CRD description strings. No upstream component receives that verb by default. Impersonation is the opt-in app-sync-using-impersonation feature (`application.sync.impersonation.enabled` in `argocd-cm`), which v3.5 extended to UI-triggered operations, matching the reported symptom. Default installs never impersonate, so default RBAC needs nothing; the reporter almost certainly has the feature enabled (asking on #4028 to confirm).

**Why the blanket grant is a problem:** `impersonate` on all `serviceaccounts`, cluster-wide, unconditionally means argocd-server can act as any SA in the cluster, including ones bound to cluster-admin. Any server compromise becomes a cluster takeover path. It also inverts the feature's own security model: upstream's documented flow has the platform admin grant impersonation scoped per tenant SA (specific `resourceNames` and namespaces) during onboarding. Least privilege is the reason the feature exists; a default blanket grant quietly defeats it for every chart user, enabled or not. Both the chart's mirror-upstream policy and general security judgment point the same way.

**The shape that could merge:** users who enable impersonation can already add scoped rules via `server.clusterRoleRules.rules`. So either:

1. Rework this into an explicitly opt-in value that renders a *scoped* rule (user-supplied `resourceNames`, namespace-scoped where possible), default off; or
2. Convert it to a documentation change pointing impersonation users at `server.clusterRoleRules.rules` with a worked example.

Happy to review either direction. Your other PRs in this sweep have been consistently well-researched, so I suspect the upstream RBAC check was the one missing piece here.

lemon0333:
@jmeridth Thanks a lot for the careful RBAC catch — you're right, the blanket cluster-wide `impersonate` grant was the wrong default and would have quietly handed every chart user (impersonation or not) a cluster-takeover path. I've reworked the PR along your option 1: an explicitly opt-in value that renders a **scoped** rule, default off.

**What changed**

Removed the unconditional `impersonate` grant from both `templates/argocd-server/clusterrole.yaml` and `role.yaml`. Added a dedicated opt-in block:

```yaml
server:
  impersonation:
    # Grant argocd-server `impersonate` on serviceaccounts for the v3.5+ app-sync impersonation feature.
    enabled: false
    # ServiceAccount names argocd-server may impersonate; rendered as resourceNames to scope the grant.
    serviceAccounts: []
```

I went with a dedicated toggle rather than a docs-only pointer at `server.clusterRoleRules.rules` because that passthrough (a) *replaces* the entire default server ruleset when enabled, so users would have to re-declare all default RBAC just to add impersonation, and (b) only covers the ClusterRole — it can't touch the namespace-scoped Role. The dedicated value is additive and renders into both, scoped by `resourceNames`. Users who want something broader can of course still use `clusterRoleRules.rules`.

**Rendered before/after**

Default (`helm template charts/argo-cd`) — no `impersonate` verb anywhere in the server RBAC (the only `impersonate` strings left in the full render are the AppProject CRD description fields, matching what you verified upstream):

```
$ helm template argocd charts/argo-cd \
    --show-only templates/argocd-server/clusterrole.yaml \
    --show-only templates/argocd-server/role.yaml | grep -c impersonate
0
```

With `server.impersonation.enabled=true` and `serviceAccounts: [tenant-a-deployer, tenant-b-deployer]` — a scoped rule (namespace-scoped in the Role, `resourceNames`-scoped in both), never cluster-wide-all:

```yaml
# ClusterRole
- apiGroups: [""]
  resources:
    - serviceaccounts
  resourceNames:
    - tenant-a-deployer
    - tenant-b-deployer
  verbs:
    - impersonate
```

```yaml
# namespace Role
- apiGroups: [""]
  resources:
    - serviceaccounts
  resourceNames:
    - tenant-a-deployer
    - tenant-b-deployer
  verbs:
    - impersonate
```

Default off ⇒ zero change to default RBAC, so it's backwards compatible. `helm lint` is green, README regenerated, and the changelog is a single `kind: added` entry. Chart bumped to `10.6.5`. Thanks again for pushing back on this one.

lemon0333:
Re-rebased onto current `main` (now argo-cd chart 10.6.3 after a sibling PR merged) to clear the conflict; bumped chart `version` to **10.6.6**. The opt-in scoped impersonation rework is unchanged — default off renders no `impersonate` verb in the server ClusterRole/Role, and enabling `server.impersonation.enabled=true` with `server.impersonation.serviceAccounts` renders a rule scoped via `resourceNames`. `helm lint` passes.

jmeridth (dismissed):

charts/argo-cd/templates/argocd-server/clusterrole.yaml
```diff
@@ -17,6 +17,18 @@ rules:
       - delete  # supports deletion a live object in UI
       - get     # supports viewing live object manifest in UI
       - patch   # supports `argocd app patch`
+  {{- if .Values.server.impersonation.enabled }}
+  - apiGroups:
+      - ""
+    resources:
+      - serviceaccounts
+    {{- with .Values.server.impersonation.serviceAccounts }}
```
tico24, line 25: With `serviceAccounts` at its default `[]` this `with` drops `resourceNames` altogether, so setting `enabled=true` on its own lets argocd-server impersonate any ServiceAccount in the cluster, including ones bound to cluster-admin. The comment on line 30 still says it's scoped. Could it fail when the list is empty instead of rendering the wide rule?

charts/argo-cd/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: v3.5.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.6.3
+version: 10.6.6
```
tico24: 10.6.3 to 10.6.6 here, main's still on 10.6.3. Skipping two deliberate?

tico24 (changes requested):

lemon0333:
Re-rebased onto current `main` (now 10.6.4) to clear the Chart.yaml conflict — version stays **10.6.6**, opt-in scoped impersonation rework unchanged (default renders no `impersonate` verb; enabling renders a `resourceNames`-scoped rule).

charts/argo-cd/templates/argocd-server/clusterrole.yaml
tico24, line 25: Opt-in sorts the default out, but `--set server.impersonation.enabled=true` on its own still renders the rule with no `resourceNames`, so it's impersonate on every ServiceAccount in the cluster. The comment on the rendered rule says it's scoped to `server.impersonation.serviceAccounts` while it isn't. Can it fail when the list is empty rather than rendering the wide rule?

mkilchhofer: Same question from my side. 

If it doesn't break, I think it would be better that the user explicitly needs to declare something we do not recommend.

**Update:**
Oh, seems like an empty set will still result in the same behavior:
```bash
$ kubectl explain role.rules.resourceNames
...

DESCRIPTION:
    ResourceNames is an optional white list of names that the rule applies to.
    An empty set means that everything is allowed.
```

Only thing we could do it to preload the list with a dummy value I assume - maybe something like `["example-sa"]`. Ugly as well :/

lemon0333:
Reworked to opt-in scoped impersonation (default off, no blanket grant) and rebased to 10.9.0 — could you take another look? Thanks!

charts/argo-cd/templates/argocd-server/clusterrole.yaml
tico24: @lemon0333 this one's still open. `--set server.impersonation.enabled=true` on its own still renders `impersonate` on every ServiceAccount, and `role.yaml` has the same `with`. Can both of them `fail` when `serviceAccounts` is empty? You'll need a rebase too. Main's on 10.9.1 now, so this wants 10.10.0. ([versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning))

PR #4038, opened 2026-08-25
fix(argo-workflows): honor crds.annotations when rendering minified CRDs

### LINKED ISSUE #4037: argo-workflows: crds.annotations has no effect; minified workflowtasksets CRD exceeds client-side apply annotation limit
### Describe the bug

`crds.annotations` is documented in `values.yaml` as:

```yaml
# -- Annotations to be added to all CRDs (only applies when crds.full=false)
annotations: {}
```

but it has no effect: no template references it. The minified CRD path (`templates/crds.yaml`) only calls the `argo-workflows.insertResourcePolicyAnnotation` helper, which does a literal string replacement of the hard-coded `helm.sh/resource-policy: keep` block (keeping it when `crds.keep=true`, replacing it with `annotations: {}` when `false`). `.Values.crds.annotations` is never rendered.

This matters more than a cosmetic dead value since v4.0: the minimal `workflowtasksets` CRD is no longer minimized (argoproj/argo-workflows#15175 — "TaskSets are no longer minimized in the minimized set", minimization threshold is 1MB) and now weighs **~646 KiB**:

| chart | appVersion | `files/crds/minimal/argoproj.io_workflowtasksets.yaml` |
|---|---|---|
| 0.47.5 | v3.7.11 | 1,005 B |
| 1.0.18 | v4.0.6 | 643,223 B |
| 2.0.2 | v4.1.2 | 661,825 B |

This exceeds the 262,144-byte limit of the `kubectl.kubernetes.io/last-applied-configuration` annotation used by client-side apply. GitOps tools default to client-side apply (Argo CD requires an explicit `ServerSideApply=true` opt-in), so users who pick `crds.full=false` — the option that keeps CRDs in the GitOps desired state instead of the hook Job — fail to sync this CRD with:

```
CustomResourceDefinition.apiextensions.k8s.io "workflowtasksets.argoproj.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes
```

The argo-cd chart hit the same wall with the ApplicationSet CRD (#3704) and solved it by rendering `crds.annotations` with a default of `argocd.argoproj.io/sync-options: ServerSideApply=true`. The argo-workflows chart documents the same value but never implemented it, so the per-resource opt-in is not available here.

### Related helm chart

argo-workflows

### Helm chart version

2.0.2 (affects every chart >= 1.0.0 shipping v4.x minimal CRDs)

### To Reproduce

```console
$ helm template aw argo/argo-workflows --version 2.0.2 \
    --set crds.full=false \
    --set 'crds.annotations.argocd\.argoproj\.io/sync-options=ServerSideApply=true' \
  | yq -N 'select(.kind=="CustomResourceDefinition") | .metadata.annotations'
```

Every CRD renders only `helm.sh/resource-policy: keep` — the configured annotation is missing.

### Expected behavior

Annotations set in `crds.annotations` are rendered on the minified CRDs, so users can e.g. opt the oversized CRDs into Argo CD server-side apply per-resource, as the values documentation implies and as the argo-cd chart already does.

### Additional context

I have a small patch ready (rendering `crds.keep` + `crds.annotations` through the existing helper, output byte-identical to 2.0.2 when `crds.annotations` is empty) and will open a PR right after filing this issue.

chmikata:
Fixes #4037

### Motivation

`crds.annotations` is documented in `values.yaml` ("Annotations to be added to all CRDs (only applies when crds.full=false)") but no template references it. The minified CRD path only runs the `insertResourcePolicyAnnotation` helper, a literal string replacement that toggles the hard-coded `helm.sh/resource-policy: keep` block.

Since v4.0 the minimal `workflowtasksets` CRD is no longer minimized (argoproj/argo-workflows#15175) and weighs ~646 KiB — beyond the 262,144-byte `last-applied-configuration` annotation limit of client-side apply. Users on `crds.full=false` with GitOps tooling need a per-resource escape hatch such as `argocd.argoproj.io/sync-options: ServerSideApply=true`, which is exactly what the argo-cd chart already offers through its (implemented) `crds.annotations` after #3704.

### Modifications

- Replace the `insertResourcePolicyAnnotation` helper with `insertCRDAnnotations`: builds the annotations block from `crds.keep` plus `crds.annotations` and substitutes it into the CRD files. Falls back to `annotations: {}` when both are empty, matching current behavior.
- `templates/crds.yaml` passes `crds.annotations` to the helper.
- Extend `ci/aa-minimal-crds-values.yaml` so CI exercises the new path.
- `helm.sh/resource-policy` cannot be overridden via `crds.annotations` when `crds.keep=true` (merge precedence keeps the chart-managed value).

No values schema change: the key already exists with a sensible default (`{}`).

### Verification

`helm template` matrix against the released 2.0.2 package:

| case | result |
|---|---|
| `crds.annotations` empty, `keep=true` | output byte-identical to released 2.0.2 |
| `crds.annotations` empty, `keep=false` | output byte-identical to released 2.0.2 (`annotations: {}`) |
| SSA annotation + `keep=true` | both `helm.sh/resource-policy: keep` and the custom annotation rendered on all 8 CRDs |
| SSA annotation + `keep=false` | only the custom annotation rendered |
| multiple annotations, value containing `:` | correctly quoted via `toYaml` |
| `crds.full=true` | CRDs not rendered in templates (unchanged) |

Checklist:

### DIFF
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -211,18 +211,20 @@ Allows overriding it for multi-namespace deployments in combined charts.
 {{- end }}
 
 {{/*
-Insert annotations into the workflows CRDs: helm.sh/resource-policy (crds.keep)
-plus any custom annotations (crds.annotations, e.g. argocd.argoproj.io/sync-options)
+Set helm.sh/resource-policy (crds.keep) and crds.annotations on the workflows CRDs
 */}}
 {{- define "argo-workflows.insertCRDAnnotations" -}}
+{{- $crd := fromYaml .input -}}
+{{- if $crd.Error -}}
+  {{- fail (printf "argo-workflows: failed to parse CRD as YAML: %s" $crd.Error) -}}
+{{- end -}}
 {{- $annotations := dict -}}
 {{- if .keep -}}
   {{- $_ := set $annotations "helm.sh/resource-policy" "keep" -}}
 {{- end -}}
 {{- $annotations = merge $annotations (.annotations | default dict) -}}
-{{- $replacement := "\n  annotations: {}\n" -}}
-{{- if $annotations -}}
-  {{- $replacement = printf "\n  annotations:\n%s\n" (toYaml $annotations | indent 4) -}}
-{{- end -}}
-{{- regexReplaceAllLiteral "\n  annotations:\n    helm.sh/resource-policy: keep\n" .input $replacement -}}
+{{- $metadata := $crd.metadata | default dict -}}
+{{- $_ := set $metadata "annotations" $annotations -}}
+{{- $_ := set $crd "metadata" $metadata -}}
+{{- toYaml $crd -}}
 {{- end -}}
--- a/charts/argo-workflows/templates/_helpers.tpl
+++ b/charts/argo-workflows/templates/_helpers.tpl
@@ -211,8 +211,18 @@ Allows overriding it for multi-namespace deployments in combined charts.
 {{- end }}
 
 {{/*
-Insert helm/resource-policy annotations into the workflows CRDs
+Insert annotations into the workflows CRDs: helm.sh/resource-policy (crds.keep)
+plus any custom annotations (crds.annotations, e.g. argocd.argoproj.io/sync-options)
 */}}
-{{- define "argo-workflows.insertResourcePolicyAnnotation" -}}
-{{- regexReplaceAllLiteral "\n  annotations:\n    helm.sh/resource-policy: keep\n" .input (ternary "\n  annotations:\n    helm.sh/resource-policy: keep\n" "\n  annotations: {}\n" .value) -}}
+{{- define "argo-workflows.insertCRDAnnotations" -}}
+{{- $annotations := dict -}}
+{{- if .keep -}}
+  {{- $_ := set $annotations "helm.sh/resource-policy" "keep" -}}
+{{- end -}}
+{{- $annotations = merge $annotations (.annotations | default dict) -}}
+{{- $replacement := "\n  annotations: {}\n" -}}
+{{- if $annotations -}}
+  {{- $replacement = printf "\n  annotations:\n%s\n" (toYaml $annotations | indent 4) -}}
+{{- end -}}
+{{- regexReplaceAllLiteral "\n  annotations:\n    helm.sh/resource-policy: keep\n" .input $replacement -}}
 {{- end -}}
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v4.1.2
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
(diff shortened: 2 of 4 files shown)


charts/argo-workflows/templates/_helpers.tpl
```diff
@@ -211,8 +211,18 @@ Allows overriding it for multi-namespace deployments in combined charts.
...
+{{- $annotations := dict -}}
+{{- if .keep -}}
+  {{- $_ := set $annotations "helm.sh/resource-policy" "keep" -}}
+{{- end -}}
+{{- $annotations = merge $annotations (.annotations | default dict) -}}
+{{- $replacement := "\n  annotations: {}\n" -}}
+{{- if $annotations -}}
+  {{- $replacement = printf "\n  annotations:\n%s\n" (toYaml $annotations | indent 4) -}}
+{{- end -}}
+{{- regexReplaceAllLiteral "\n  annotations:\n    helm.sh/resource-policy: keep\n" .input $replacement -}}
```
vladlosev: This is fragile. Should any of the following changes happen in a CRD file, the replacement will cease working:
- The original annotation value is changed to something other than `keep`.
- The whitespace amount between the annotation name and value is chanegd.
- The original annotation is removed.
- Another annotation is inserted above the original one.
- The indentation of the `annotations` section or annotation keys within it is changed.

chmikata: Thanks for the review — you're right, and I'd rather not leave that in.

The substitution depended on the exact formatting of the annotations block in `files/crds/minimal`, which is generated by `scripts/update-argo-workflows-crds.sh`. Any of the changes you listed would have made the match fail, and because `regexReplaceAllLiteral` returns the input unchanged when nothing matches, it would have failed silently — the CRDs would still render, just without the annotations.

Pushed 1e3bfe9, which drops the string matching entirely: the CRD is parsed with `fromYaml`, `metadata.annotations` is set structurally, and the result is re-emitted with `toYaml`. It no longer depends on the formatting of the annotations block, so none of the five cases apply. A CRD that fails to parse now calls `fail` instead of passing through unchanged.

Verified against the released 2.0.2 chart:
- All 8 CRDs render semantically identical with `crds.annotations` unset, for both `crds.keep=true` and `false`
- Non-CRD resources unchanged
- `crds.annotations` is applied to all 8 CRDs; `helm.sh/resource-policy` still wins over a user-supplied value while `crds.keep=true`
- Re-ran each of your five scenarios against a mutated CRD file: annotations are now applied correctly in all of them
- Render time 0.09s → 0.14s (the 646 KiB `workflowtasksets` CRD is now parsed rather than scanned)

vladlosev (approved):
LGTM

chmikata:
Thanks for the review and approval, @vladlosev.

Good catch on the string substitution — `regexReplaceAllLiteral` returns the input unchanged when nothing matches, so any formatting drift in the generated CRDs would have silently dropped the annotations. It now parses the CRD with `fromYaml` and fails loudly instead.

yu-croco (approved):
Thank you for your contribution.

Issue #2887, opened 2024-08-19
ArgoCD stuck in `waiting for completion of hook batch/Job/argocd-redis-secret-init`


oponomarov-tu:
### Describe the bug

Ref: https://github.com/argoproj/argo-helm/pull/2861

I still face the same issues even on Helm charts >= v7.4.1. My ArgoCD deployed on Fargate is managing itself after bootstrap. Tested in clean EKS (v1.30). Attaching `values.yaml` for reference.

<details><summary>Details</summary>
<p>

```yaml
controller:
  replicas: 1
  env:
    - name: ARGOCD_K8S_CLIENT_QPS
      value: "300"
    - name: ARGOCD_SYNC_WAVE_DELAY
      value: "30"
  resources:
    limits:
      memory: 3Gi
    requests:
      cpu: 1
      memory: 3Gi
repoServer:
  autoscaling:
    enabled: true
    minReplicas: 1
  resources:
    requests:
      cpu: "100m"
      memory: "1Gi"
    limits:
      memory: "1Gi"
applicationSet:
  replicaCount: 1
dex:
  enabled: false
server:
  autoscaling:
    enabled: true
    minReplicas: 1
  resources:
    requests:
      cpu: "100m"
      memory: "256Mi"
    limits:
      memory: "512Mi"
redis:
  enabled: false
redis-ha:
  enabled: true
  haproxy:
    enabled: true
    init:
      resources:
        requests:
          memory: 100Mi
          cpu: 10m
        limits:
          memory: 100Mi
    resources:
      requests:
        memory: 200Mi
        cpu: 15m
      limits:
        memory: 200Mi
  redis:
    resources:
      requests:
        memory: 500Mi
        cpu: 20m
      limits:
        memory: 1Gi
  sentinel:
    resources:
      requests:
        memory: 20Mi
        cpu: 10m
      limits:
        memory: 20Mi
  exporter:
    resources:
      requests:
        memory: 25Mi
        cpu: 5m
      limits:
        memory: 25Mi
  init:
    resources:
      requests:
        memory: 25Mi
        cpu: 25m
      limits:
        memory: 25Mi
configs:
  secret:
    githubSecret: "$github-argocd-webhook:webhook.github.secret"
  repositories:
    aws-public-ecr:
      name: aws-public-ecr
      type: helm
      url: public.ecr.aws
      enableOCI: "true"
```

</p>
</details> 

Perhaps the `ttlSecondsAfterFinished` should be configurable? 

[...]

PR #2861, opened 2024-08-03
fix(argo-cd): Set TTL on redis-secret-init Job

jkleinlercher:
set `ttlSecondsAfterFinished` to prevent stuck during sync when managed by argocd

should fix https://github.com/argoproj/argo-cd/issues/6880#issuecomment-2263802072

---
Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [ ] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #4050, opened 2026-09-01
fix(argo-cd): make ttlSecondsAfterFinished configurable for redisSecretInit

duncan485:
## Description

This PR makes `ttlSecondsAfterFinished` configurable via `redisSecretInit.ttlSecondsAfterFinished` (default bumped from `60` to `600` seconds), giving ArgoCD's controller much more time to observe hook completion before the Job is cleaned up.

Note on DCO:

-->

Checklist:


jmeridth:
Bumping from 60 to 600 is a serious change IMO. 1 minute to 10 minutes. Please change back to 60 as the default. People can bump it higher if needed after your change. Unless you can defend the 9 minute bump? Thank you for the contribution.

jmeridth:
:robot: Following up on my earlier comment after digging through the history, because the 60/600 question turns out to be more interesting than a simple default change.

**What I found:**

- The current `60` was itself a bug-fix attempt: #2861 (Aug 2024) *added* the TTL to resolve the "stuck waiting for hook" symptom from argoproj/argo-cd#6880. Then #2887 was opened weeks later reporting the same symptom persists, and it's been open for two years.
- Upstream confirmed the exact race you're describing. From the maintainer discussion in argoproj/argo-cd#6880: the hang occurs when a hook object is deleted externally (like the Job controller acting on TTL) before the gitops engine observes the terminal state, and it then waits forever. So the mechanism behind your change is real, not speculation, and the self-managed case is worst because the application controller can restart mid-sync during its own upgrade.
- The proper fix merged upstream: argoproj/gitops-engine#646 attaches a finalizer to hook resources so they can't be deleted before Argo CD records the outcome. That shipped in Argo CD 3.0+, yet intermittent reports persist (including on v3.2.6 as recently as Jan 2026).

**So the key question: which Argo CD version are you running when you hit this?** This chart's current appVersion (3.5.x) should contain the finalizer fix. If you're reproducing the race on 3.5.x, that's an upstream regression worth reporting on argoproj/argo-cd#6880 with your details, and a wider TTL default here is a reasonable stopgap. If you're on an older Argo CD, the default should stay `60` and the configurability this PR adds is the fix for your case.

**Either way, two asks:**

1. If we do change the default, the values comment and changelog need to say *why* (the TTL-vs-hook-observation race, with links to #2887 and argoproj/argo-cd#6880), because `600` looks arbitrary without that context and the next person will try to "clean it up" back to 60.
2. Version collision heads-up: this targets chart `10.6.1`, which the just-approved #3888 also targets; whichever merges second needs a re-bump.

The configurability itself is welcome regardless of where the default lands. Thanks for tracing this into #2887; that issue has needed movement for a long time.

charts/argo-cd/values.yaml
```diff
@@ -2003,6 +2003,9 @@ redisSecretInit:
   # -- Additional command line arguments for the Redis secret-init Job
   extraArgs: []
 
+  # -- Seconds after which the completed Redis secret-init Job is automatically deleted
+  ttlSecondsAfterFinished: 600
```
tico24: This bumps the default from 60 to 600 as well as making it configurable, so the render changes for everyone rather than only for people who set it. Shouldn't that be a minor rather than a patch? ([versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning))

duncan485:
Hey @jmeridth  I am actually on argo 3.3.9 (helm version 9.5.11). So I don't think that upsteam worked for me. I will edit my MR so it just makes it configurable, but still defaults to 60. Than at least we can try if this fixes the actual issue. If it does, we can change the default in a later MR?

jmeridth:
@duncan485 thanks for confirming the version — argo-cd 3.3.9 matters more than you might expect: the upstream finalizer fix for exactly this race (argoproj/gitops-engine#646) shipped in the 3.0 line, so reproducing the TTL race on 3.3.9 is evidence of a gap in that fix. When you have a moment, please post your reproduction details on argoproj/argo-cd#6880 (version, self-managed setup, the stuck-hook state); the maintainers there have been asking whether post-3.0 reports still exist.

Your revised plan is exactly right for this PR: configurable `ttlSecondsAfterFinished` with the default staying `60`. Two mechanical notes for the update: per @tico24's point and our versioning conventions, a new value means a chart **minor** bump (and with the recent merge activity, rebase onto current main — argo-cd chart is at 10.7.0, so target 10.8.0); and please include the rationale in the value's helm-docs comment (why one might raise it, with a pointer to #2887) so the number has context. Happy to review once pushed.

duncan485:
Hey @jmeridth Thanks, I updated the MR. As the default is now kept at 60, I think say a patch version should be the right bump. 

I will definitely post a to the original story on how to reproduce. But I am also verry curious to see if this solves it in my case.

charts/argo-cd/values.yaml
```diff
@@ -2026,6 +2026,11 @@ redisSecretInit:
   # -- Additional command line arguments for the Redis secret-init Job
   extraArgs: []
 
+  # -- Seconds after which the completed Redis secret-init Job is automatically deleted.
+  # Increase this value if you are experiencing the hook-completion race described in
+  # https://github.com/argoproj/argo-helm/issues/2887 and the Argo CD issue linked there.
+  ttlSecondsAfterFinished: 60
```
tico24, line 2054: The Job template still has `ttlSecondsAfterFinished: 60` hardcoded, so the new value doesn't do anything. `--set redisSecretInit.ttlSecondsAfterFinished=600` renders 60. `templates/redis-secret-init/job.yaml` needs to read it.

charts/argo-cd/Chart.yaml
```diff
@@ -26,5 +26,7 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
+    - kind: fixed
```
tico24, line 29: The changelog should only show the change from this PR. Delete the old one please. Running `lint-changelog.sh` would have caught this. ([changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog))

charts/argo-cd/values.yaml
duncan485: whew, somehow this got lost while rebasing I think, my bad. Adjusted

charts/argo-cd/Chart.yaml
```diff
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: security
-      description: Bump the redis-ha Redis image to 8.6.4-alpine so it no longer ships an OpenSSL version affected by the RSASVE uninitialized-memory issue (reported by m98anu)
+    - kind: fixed
+      description: Allow redis-secret-init Job ttlSecondsAfterFinished to be configured while preserving the default 60s value for compatibility with older Argo CD hook-completion behavior; increase it if you are experiencing the race described in argo-helm#2887.
```
tico24: Ran `ct lint` on this locally. Once the version's sorted it still fails: `charts/argo-cd/Chart.yaml 30:264 no new line character at the end of file`. Worth doing both in the same push.

charts/argo-cd/templates/redis-secret-init/job.yaml
```diff
@@ -13,7 +13,9 @@ metadata:
   labels:
     {{- include "argo-cd.labels" (dict "context" . "component" .Values.redisSecretInit.name "name" .Values.redisSecretInit.name) | nindent 4 }}
 spec:
-  ttlSecondsAfterFinished: 60
+  {{- with .Values.redisSecretInit.ttlSecondsAfterFinished }}
```
tico24: `--set redisSecretInit.ttlSecondsAfterFinished=0` drops the key entirely, since `with` reads 0 as empty. A Job with no `ttlSecondsAfterFinished` never gets cleaned up, so the value that means delete it immediately gets you the opposite. `if ... ne nil` instead?

charts/argo-cd/values.yaml
tico24: Patch is right, ignore my earlier point — that was about the default moving to 600 and you've put it back to 60, so it doesn't apply. You'll want 10.8.5, main's moved to 10.8.4 since.

Issue #840, opened 2021-07-19
The argo-events chart doesn't include an EventBus


justinmchase:
**Is your feature request related to a problem? Please describe.**
I installed argo-events through helm but it just doesn't work. I setup my kafka event source and it never connects giving an error about a missing EventBus.

When I manually run `kubectl apply -n argo-events -f https://raw.githubusercontent.com/argoproj/argo-events/stable/examples/eventbus/native.yaml` then the bus appears and I see it starts connecting.

It really seems like the helm chart should create the EventBus when you install it and just have configuration options to configure it. But the event system seems to not function without it so it probably should be in the chart.

**Describe the solution you'd like**
I'd like to just be able to `helm install argo-events` and boom it has everything to function, at least simply, by default.

**Describe alternatives you've considered**
I can just run more stuff and manually create resources but it is confusing and takes a lot of effort just to even try argo out at all.

**Additional context**
Additionally, I hate that argo goes into one namespace and argo-events goes into another... Please just put everything under the 1 argo namespace.

osherlevi:
Same for me, saw this error:
[...]

PR #4045, opened 2026-08-30
fix(argo-events): update JetStream NATS version

ishanupadhyay:
## Description

Updates the JetStream NATS versions in the `argo-events` Helm chart to match the upstream Argo Events `release-1.9` configuration.

The following versions are updated:
- `latest`: NATS `2.10.10` → `2.10.29`
- Explicit JetStream version: `2.10.10` → `2.10.29`

The existing older JetStream versions are left unchanged.

This also bumps the `argo-events` chart version from `2.4.24` to `2.4.25` and adds the corresponding Artifact Hub changelog entry.

## Testing

- `./scripts/lint.sh argo-events` ✅
- `helm template argo-events charts/argo-events --values charts/argo-events/ci/test-values.yaml` ✅
- Verified the rendered JetStream configuration contains NATS `2.10.29` ✅

closes [ISSUE #4025]

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #4052, opened 2026-09-02
docs(argo-events): document EventBus creation via extraObjects

jmeridth:
Relates to #840

## What/Why

Documents that the argo-events chart intentionally creates no EventBus (matching upstream install manifests) and shows how to create one from the same values file via `extraObjects`, with JetStream and native NATS examples. Users have been hitting "missing EventBus" errors after install since 2021 with the answer buried in the issue #840 thread; the issue was closed today with this docs PR as the follow-up.

## Proof it works

`./scripts/lint.sh` passes and `./scripts/helm-docs.sh` regenerated the README cleanly from the updated gotmpl. Docs-only change; no template or values changes.

## Risk + AI role

Low: README/gotmpl plus the required chart version bump and changelog entry. AI-assisted drafting, human-reviewed.

## Review focus

Whether the JetStream example version (`2.10.29`) is the right one to pin in docs, and the framing of the "chart deliberately doesn't create a bus" rationale.

---

Checklist:

* [x] I have updated the chart changelog according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog)
* [x] Any new values are backwards compatible and/or have sensible default. (n/a: docs only)

### DIFF
--- a/charts/argo-events/README.md.gotmpl
+++ b/charts/argo-events/README.md.gotmpl
@@ -30,6 +30,38 @@ NAME: my-release
 ...
 ```
 
+## EventBus
+
+This chart does **not** create an [EventBus](https://argoproj.github.io/argo-events/eventbus/eventbus/), matching the upstream argo-events install manifests. Event sources and sensors will not connect until an EventBus exists in the namespace. The bus type (native NATS, JetStream, or Kafka) is an environment decision, so the chart leaves it to you.
+
+You can create one from the same values file using `extraObjects`:
+
+```yaml
+extraObjects:
+  - apiVersion: argoproj.io/v1alpha1
+    kind: EventBus
+    metadata:
+      name: default
+    spec:
+      jetstream:
+        version: 2.10.29
+```
+
+or, for the native NATS bus:
+
+```yaml
+extraObjects:
+  - apiVersion: argoproj.io/v1alpha1
+    kind: EventBus
+    metadata:
+      name: default
+    spec:
+      nats:
+        native: {}
+```
+
+See the [EventBus documentation](https://argoproj.github.io/argo-events/eventbus/eventbus/) for the full spec and the supported JetStream versions in this chart's `values.yaml` (`configs.jetstream.versions`).
+
 ## Upgrading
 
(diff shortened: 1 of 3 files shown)


charts/argo-events/README.md.gotmpl
```diff
@@ -30,6 +30,38 @@ NAME: my-release
...
+
+```yaml
+extraObjects:
+  - apiVersion: argoproj.io/v1alpha1
+    kind: EventBus
+    metadata:
+      name: default
+    spec:
+      jetstream:
+        version: 2.10.29
```
tico24: Won't this fail? 2.10.29 isn't in `configs.jetstream.versions`, which tops out at 2.10.10, and the controller matches the version string exactly, so the bus never comes up. `latest` or `2.10.10` would work.

jmeridth: @tico24 Good catch, and you are right about the mechanism. The controller matches the version string exactly against `configs.jetstream.versions`, and 2.10.29 is not in the shipped list (it arrives with the pending sync in #4045), so the example could not work as written. Addressed in https://github.com/argoproj/argo-helm/commit/3b0c039736a117fd8f0690eebd119704d92ed339 using `latest` rather than `2.10.10`, since #4045 replaces the 2.10.10 entry when it merges and would break a pinned example again. The example now also warns that out-of-list versions prevent the bus from starting.

tico24: Ignore the 2.10.10 half of that. #4045 looks like it's replacing 2.10.10 with 2.10.29 rather than adding it, so once that lands your example is right as written. `latest` works either way.

jmeridth:
Rebased onto current main after #4045 merged at `2.4.25`; this PR now targets `2.4.26`. Docs content unchanged from the last review round (`latest` example per @tico24's catch), only Chart.yaml moved.

tico24 (approved):

jmeridth:
@yu-croco 🙏

yu-croco (approved):

PR #4062, opened 2026-09-04
docs(github): add AGENTS.md and CLAUDE.md for AI coding agents

tico24:
Agents that open PRs here keep deleting this template, writing their own description, and ticking boxes nobody checked. AGENTS.md tells them to leave the template alone and tick only what they verified.

The rest of it is the chart stuff they get wrong: bump the version, add the `artifacthub.io/changes` entry, never hand-edit a generated README, one chart per PR. Everything else points at CONTRIBUTING.md.

If an agent wrote the PR it adds a disclosure line to the checklist, unticked, for the human to tick. There's one in the list below. [genai.md](https://github.com/argoproj/argoproj/blob/main/community/genai.md) already expects contributors to be clear about LLM use but doesn't say how, so this is one way of doing it.

CLAUDE.md is one line importing AGENTS.md. No second copy to drift.

No chart touched, so the chart boxes below don't apply.

## Rollback

Revert this PR. Nothing reads these files at build or release time.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] I wrote this PR with the help of a LLM but I fully vetted the work before raising this PR for review
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #4063, opened 2026-09-04
docs(github): tell agents to lint locally before opening a PR

tico24:
Follow-up to #4062, for the addition @jmeridth asked for in his review: tell agents to run the lint scripts locally before opening the PR.

Two things I changed from the suggested wording:

- `./scripts/lint.sh` takes no chart argument. It runs `ct lint`, which diffs against `origin/main` and works out the changed charts itself, and an extra argument is silently ignored. The line says so, otherwise an agent passes a chart name and believes it linted only that chart.
- `lint.sh` does not look at the changelog. That is `scripts/lint-changelog.sh` locally and `ah lint` in CI, so the line names the changelog script too. It needs only bash, no Docker, which makes it the cheap half of the pair.

Ran both on this branch. `lint-changelog.sh` passes, and `lint.sh` prints "No chart changes detected", which is right for a docs-only change.

No chart touched, so the chart boxes below don't apply.

## Rollback

Revert this PR. Nothing reads this file at build or release time.

Checklist:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,6 +30,8 @@ Every change to a chart needs all of the following, including a change that only
 
 Run `helm template` on the chart before and after your change and read the diff. If it shows anything you did not intend, fix that before opening the pull request.
 
+Run `./scripts/lint.sh` before you open the pull request. It takes no arguments and runs chart-testing in Docker over the charts you changed, the same lint CI runs, so it is what catches a missing version bump. Run `./scripts/lint-changelog.sh` too. That one needs only bash and checks that `artifacthub.io/changes` holds exactly one entry. Both are quicker than waiting for CI to fail.
+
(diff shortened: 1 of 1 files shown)


jmeridth (approved):

jmeridth (approved):
Verified against the scripts and approving. `lint.sh` indeed takes no arguments (I checked the script: nothing reads $1, so my suggested wording was wrong and my own earlier `./scripts/lint.sh <chart>` invocations were silently ignoring the argument), ct lints the changed charts exactly as `lint-and-test.yml` does, and the version-bump catch plus the one-entry `lint-changelog.sh` description both match observed CI failures from this month's PRs. Thanks for correcting my wording rather than copying it.

mkilchhofer (approved):

PR #4068, opened 2026-09-06
fix(argo-cd): set EXTENSIONS_DIR so extensions install to the served path

### LINKED ISSUE #3320: `server.extensions` installing in the wrong path
### Describe the bug

When using `server.extensions` in the helm chart, extensions are being installed in the wrong path. They are expected in `/tmp/extensions` but they are instead inside `/tmp/extensions/resources`.

The CI test defined [here](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/ci/extension-values.yaml) is configured properly but it is not doing a full validation checking that the extensions are being loaded as expected in the UI, so it is not failing...

The issue seems to come from the `argocd-extension-installer` [repo](https://github.com/argoproj-labs/argocd-extension-installer/issues/6). I opened a [PR](https://github.com/argoproj-labs/argocd-extension-installer/pull/14) to fix it, but seems there is no answer yet.

Either we configure this repo to load the proper mount path of the extension installer (add `/resources` [here](https://github.com/argoproj/argo-helm/blob/d1a211b8062530afacc90e63c17648fd4863239b/charts/argo-cd/templates/argocd-server/deployment.yaml#L470)), or we merge in my PR to install the extensions in the expected path already.

Also, seems that the CI, would be nice to check that the extension is really working.

### Related helm chart

argo-cd

### Helm chart version

8.0.9

### To Reproduce

1. Install with the following values
```
server:
  extensions:
    enabled: true
    extensionList:
      - name: rollout-extension
        env:
          - name: EXTENSION_URL
            value: https://github.com/argoproj-labs/rollout-extension/releases/download/v0.3.7/extension.tar
      - name: extension-metrics
        env:
          - name: EXTENSION_URL
            value: https://github.com/argoproj-labs/argocd-extension-metrics/releases/download/v1.0.3/extension.tar.gz
          - name: EXTENSION_CHECKSUM_URL
            value: https://github.com/argoproj-labs/argocd-extension-metrics/releases/download/v1.0.3/extension_checksums.txt
```

2. open a shell to the pod, see that extensions are being loaded into `/tmp/extensions/resources` and UI extensions are not visible in the dashboard.

### Expected behavior

UI Extensions should be visible in the dashboard and they should be loaded from `/tmp/extensions/resources` or extensions-installer image to be updated with the right path.

### Screenshots

_No response_

### Additional context

_No response_

jmeridth:
Closes #3320

## What/Why

`server.extensions` mounted the shared volume at `/tmp/extensions` (where argocd-server discovers extensions) while the installer image copies into `/tmp/extensions/resources` by default, so installed extensions never loaded in the UI. Upstream made the target configurable via `EXTENSIONS_DIR` (argoproj-labs/argocd-extension-installer#14, thanks @Sturgelose, shipped in the v1.1.0 image the chart pins since #4064), and its documentation prescribes `/tmp/extensions` for this chart's wiring. This sets that env on every extension init container. Entries from each extension's `.env` render after it and take precedence, and the `.env` block is now `with`-guarded so env-less extensions render validly.

## Proof it works

`helm template` with an `extensionList` entry renders `EXTENSIONS_DIR` before the extension's own env, nothing leaks into output from the template comment, and `./scripts/lint.sh` passes. Older pinned installer images ignore the unknown variable, so pinning `v1.0.1` keeps exactly today's behavior.

## Risk + AI role

Low: only the opt-in extensions init container's rendered env changes. AI-assisted, maintainer-directed.

## Review focus

@Sturgelose confirmation welcome that `/tmp/extensions` is the served path per your upstream docs table, since you authored both the report and the fix.

---

Checklist:

* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation) (n/a: no values changes)
* [x] I have updated the chart changelog according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog)
* [x] Any new values are backwards compatible and/or have sensible default. (n/a: no new values)

### DIFF
--- a/charts/argo-cd/Chart.yaml
+++ b/charts/argo-cd/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v3.5.2
 kubeVersion: ">=1.25.0-0"
 description: A Helm chart for Argo CD, a declarative, GitOps continuous delivery tool for Kubernetes.
 name: argo-cd
-version: 10.8.2
+version: 10.8.3
 home: https://github.com/argoproj/argo-helm
 icon: https://argo-cd.readthedocs.io/en/stable/assets/logo.png
 sources:
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump redis_exporter to v1.91.0
+    - kind: fixed
+      description: Set EXTENSIONS_DIR on the extension installer so server.extensions install to the path argocd-server serves them from
--- a/charts/argo-cd/templates/argocd-server/deployment.yaml
+++ b/charts/argo-cd/templates/argocd-server/deployment.yaml
@@ -489,10 +489,6 @@ spec:
           - name: tmp
             mountPath: /tmp
         env:
-          {{- /* The shared extensions volume is mounted at /tmp/extensions and argocd-server
-          discovers extensions directly under it, while the installer image's built-in default
-          target is /tmp/extensions/resources. Entries in .env render after this and take
-          precedence if an extension sets its own EXTENSIONS_DIR. */}}
           - name: EXTENSIONS_DIR
             value: /tmp/extensions
           {{- with .env }}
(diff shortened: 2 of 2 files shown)


yu-croco (dismissed):

jmeridth:
Rebased onto main after #4070 took `10.8.2`; this PR now targets `10.8.3`. Template change and changelog text unchanged, version line only.

mkilchhofer (approved):

mbevc1 (approved):

PR #4071, opened 2026-09-07
fix(argo-workflows): derive ServiceMonitor scheme from metricsConfig.secure

KR-Ravindra:
With `controller.metricsConfig.secure: true` the workflow-controller serves `/metrics` over HTTPS with a self-signed certificate, but the ServiceMonitor still renders `scheme: http` (the values default), so Prometheus fails the TLS handshake and never scrapes the controller.

Fix: `controller.metricsConfig.scheme` and `controller.telemetryConfig.scheme` now default to `""` and the template derives `https`/`http` from the matching `secure` flag; an explicit `scheme` value still wins. When the derived metrics scheme is `https` and no `metricsConfig.tlsConfig` is set, the endpoint gets `tlsConfig.insecureSkipVerify: true` for the self-signed certificate. Default values render exactly as before (`helm template` output is byte-identical to `main`).

Checklist:

* [ ] I wrote this PR with the help of a LLM but I fully vetted the work before raising this PR for review

### DIFF
--- a/charts/argo-workflows/Chart.yaml
+++ b/charts/argo-workflows/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: v4.1.2
 name: argo-workflows
 description: A Helm chart for Argo Workflows
 type: application
-version: 2.0.4
+version: 2.0.5
 icon: https://argo-workflows.readthedocs.io/en/stable/assets/logo.png
 home: https://github.com/argoproj/argo-helm
 sources:
@@ -17,4 +17,4 @@ annotations:
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
     - kind: fixed
-      description: Install notes no longer instruct submitting workflows without executor RBAC; they now explain the workflow ServiceAccount requirement and the values that create it
+      description: ServiceMonitor scheme is derived from metricsConfig.secure / telemetryConfig.secure when not set explicitly, adding tlsConfig.insecureSkipVerify for HTTPS metrics
--- a/charts/argo-workflows/README.md
+++ b/charts/argo-workflows/README.md
@@ -347,7 +347,7 @@ Fields to note:
 | controller.metricsConfig.port | int | `9090` | Port is the port where metrics are emitted |
 | controller.metricsConfig.portName | string | `"metrics"` | Container metrics port name |
 | controller.metricsConfig.relabelings | list | `[]` | ServiceMonitor relabel configs to apply to samples before scraping |
-| controller.metricsConfig.scheme | string | `"http"` | serviceMonitor scheme |
+| controller.metricsConfig.scheme | string | `""` | ServiceMonitor scheme. Leave empty to derive from `secure` flag. |
 | controller.metricsConfig.secure | bool | `false` | Flag that use a self-signed cert for TLS |
 | controller.metricsConfig.servicePort | int | `8080` | Service metrics port |
 | controller.metricsConfig.servicePortName | string | `"metrics"` | Service metrics port name |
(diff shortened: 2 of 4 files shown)


KR-Ravindra:
Self-review before asking maintainers to look (this draft was prepared by an AI agent I operate; I review every PR it opens). Keeping it as a draft: **changes requested**.

1. **Empty PR description.** It must explain the problem: with `controller.metricsConfig.secure: true` the workflow-controller serves metrics over HTTPS, but the ServiceMonitor renders `scheme: http` (the values default), so Prometheus gets a TLS handshake error and the controller is unscraped. That is the actual bug; this PR does not state it.
2. **Making `scheme` conditional does not fix anything.** Prometheus does not "auto-detect HTTPS": an omitted `scheme` means `http`. The fix should derive the scheme from the secure flag when the user has not set one explicitly, e.g. `scheme: {{ .Values.controller.metricsConfig.scheme | default (ternary "https" "http" .Values.controller.metricsConfig.secure) }}`, and when HTTPS is used add `tlsConfig: { insecureSkipVerify: true }` (the controller uses a self-signed cert), or a values knob for it.
3. **Regression on the telemetry endpoint.** The second endpoint now uses `metricsConfig.scheme` instead of `telemetryConfig.scheme`; telemetry has its own `secure`/`scheme` and must keep using them.
4. `honorLabels` being conditional only drops `honorLabels: false` from the output, which is a no-op; it can stay unconditional. The changelog entry should describe the real change once the above is done.

I will revise this with the derived-scheme approach and a `helm template` diff showing `secure: true` producing `scheme: https` on both endpoints.

KR-Ravindra:
Round 2 self-review (AI agent operated by me; I review every revision). Much better direction; **two blockers before ready**:

1. **`default` never fires.** `values.yaml` sets `controller.metricsConfig.scheme: http`, so `.Values.controller.metricsConfig.scheme | default (...)` always evaluates to `http` and `secure: true` still renders `scheme: http`. Change the values default to empty (`scheme: ""` with a comment "leave empty to derive from `secure`") for both `metricsConfig` and `telemetryConfig`, and update the values comment/README table. Render both `secure: true` and `secure: false` with `helm template` and paste the two endpoint blocks in the PR body.
2. **Invalid YAML on the telemetry endpoint.** The new `scheme:` line under the telemetry endpoint is indented 8 spaces; the sibling keys are at 6. `helm template` must succeed with `telemetryConfig.enabled: true` before pushing.

Nits: the `tlsConfig` guard should key off the *derived* scheme being `https` (a user who sets `scheme: https` explicitly also needs it), e.g. compute the scheme into a variable first; the `artifacthub.io/changes` list in this chart is per-release, so replace the previous entry instead of appending to it. Squash to one signed-off commit.

KR-Ravindra:
Round 3 self-review (AI agent operated by me; I review every revision). The template change is now correct and renders as intended. **Only cleanup left before I mark this ready:**

1. Revert `charts/argo-workflows/values.yaml` `serviceMonitor.enabled: false -> true`. That is a chart-wide default change left over from local testing and must not ship.
2. Revert the `.github/pull_request_template.md` change; it is unrelated to this chart fix.
3. Chart version: `2.0.4 -> 2.0.5` (one patch bump), not `2.0.6`.
4. DCO is failing: squash the 7 commits (including the merge commit) into one commit with `Signed-off-by`, and force-push this branch.

After that: metrics endpoint derives `scheme` from `secure` (with `tlsConfig.insecureSkipVerify` when https), telemetry endpoint derives from its own flag, values/README defaults emptied, changelog replaced. Looks good.

KR-Ravindra:
Squashed to one signed-off commit on current main, reverted the unrelated PR-template and serviceMonitor.enabled changes, version 2.0.5. Verified with helm template: secure=true renders scheme https + tlsConfig.insecureSkipVerify on the metrics endpoint, telemetry stays on its own scheme, an explicit scheme still wins; helm lint clean. Marking ready for review.

tico24:
> This change was prepared with an AI agent operated by KR-Ravindra and reviewed by hand before being marked ready.

Please get your agent to follow AGENTS.md

PR #4061, opened 2026-09-04
fix(argo-workflows): restart controller on config change

amarkdotdev:
## Summary

Adds an opt-in `controller.configMap.restartOnChange` value (default `false`) for the workflow controller deployment.

When enabled, the controller pod template gets a `checksum/config` annotation that hashes only the rendered controller ConfigMap `data`/`stringData` (via `argo-workflows.controller.config.checksum`). That follows the Helm automatic roll pattern without changing default behavior.

This is an alternative to the built-in config watcher. It does not cover semaphore ConfigMaps.

Fixes argoproj/argo-helm#2622

## Test plan

- [x] `helm template` with defaults: no `checksum/config` on the controller pod
- [x] `helm template` with `controller.configMap.restartOnChange: true`: annotation present and changes when config data changes

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #4060, opened 2026-09-03
feat(argo-workflows): render controller.metricsConfig.temporality in ConfigMap

dichico:
## Summary

Adds support for `controller.metricsConfig.temporality` so it is rendered into `templates/controller/workflow-controller-config-map.yaml`. Previously this value was silently ignored, preventing users from configuring OTLP metric temporality (for example, `Delta` vs `Cumulative`) via chart values.

## Changes
- `values.yaml`: added `controller.metricsConfig.temporality` with an unset default
- `workflow-controller-config-map.yaml`: render `temporality` under `metricsConfig` only when set
- `README.md`: added the corresponding documentation row
- `Chart.yaml`: bumped chart version from 2.0.3 to 2.1.0 and added the Artifact Hub changelog entry

## Validation
- `helm lint charts/argo-workflows`
- `helm template` verified `temporality` is omitted by default and rendered correctly when set

## Checklist

- [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
- [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
- [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
- [x] Any new values are backwards compatible and/or have sensible default.
- [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
- [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
- [ ] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

Fixes #4059

PR #3982, opened 2026-07-27
feat(argo-workflows): add server.sso.rootCA for custom OIDC CA

lemon0333:
Adds a `server.sso.rootCA` value to the `argo-workflows` chart so a custom PEM-encoded CA certificate can be supplied to validate the OIDC provider's TLS certificate (useful for self-signed / private CAs). argo-workflows already supports this via the `rootCA` SSO config field (`config/sso.go`); the chart now renders it into the workflow-controller config map alongside the existing `insecureSkipVerify` option.

Verified with `helm lint` and `helm template`: a multiline `server.sso.rootCA` renders as a correctly-indented `rootCA: |-` block scalar under the `sso:` config, and the field is omitted when left at its `""` default.

Closes #3963

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #3967, opened 2026-07-13
feat(argo-workflows): add optional podLabels to crd installer

ekarademir:
Closes #3954 

* Add optional `crds.upgradeJob.podLabels` for the CRD installer.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
* [x] My build is green ([troubleshooting builds](https://argo-cd.readthedocs.io/en/stable/developer-guide/ci/)).

PR #4073, opened 2026-09-08
feat(argo-workflows): support unhealthyPodEvictionPolicy on PDBs

Ranexia04:
Adds `controller.pdb.unhealthyPodEvictionPolicy` and `server.pdb.unhealthyPodEvictionPolicy` to the argo-workflows chart, passed through to the rendered PodDisruptionBudgets. Unset by default, so existing installations render exactly as before. Setting it to `AlwaysAllow` lets a not-Ready replica be evicted during a node drain instead of blocking it, per the [Kubernetes drain recommendation](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#pdb-unhealthy-pod-eviction-policy).

Checklist:


charts/argo-workflows/values.yaml
```diff
@@ -415,6 +415,7 @@ controller:
     enabled: false
     # minAvailable: 1
     # maxUnavailable: 1
+    # unhealthyPodEvictionPolicy: AlwaysAllow
```
jmeridth, line 418: :robot: (medium) The argo-cd twin of this feature (#4074, merged two days ago) exposed this as a real documented value with a helm-docs row rather than a commented example, so the two charts currently end up with different public surfaces for the same feature. The commented style does match `minAvailable`/`maxUnavailable` in this block, but cross-chart consistency with the fresh precedent is worth more. The empty-string default keeps rendering identical (the `with` guard skips it):
```suggestion
    # -- Policy for evicting unhealthy (not ready) pods, either `IfHealthyBudget` or `AlwaysAllow`. Defaults to `IfHealthyBudget` if not set
    unhealthyPodEvictionPolicy: ""
```
(Then re-run `./scripts/helm-docs.sh` to pick up the README rows.)

charts/argo-workflows/values.yaml
```diff
@@ -692,6 +693,7 @@ server:
...
```
jmeridth, line 696: :robot: (medium) The argo-cd twin of this feature (#4074, merged two days ago) exposed this as a real documented value with a helm-docs row rather than a commented example, so the two charts currently end up with different public surfaces for the same feature. The commented style does match `minAvailable`/`maxUnavailable` in this block (server), but cross-chart consistency with the fresh precedent is worth more. The empty-string default keeps rendering identical (the `with` guard skips it):
```suggestion
    # -- Policy for evicting unhealthy (not ready) pods, either `IfHealthyBudget` or `AlwaysAllow`. Defaults to `IfHealthyBudget` if not set
    unhealthyPodEvictionPolicy: ""
```
(Then re-run `./scripts/helm-docs.sh` to pick up the README rows.)

jmeridth (changes requested):
:robot: (summary) Verified twin of #4074 for the workflows chart: template pattern identical to the merged precedent, `unhealthyPodEvictionPolicy` valid on policy/v1 (opt-in, so no version gating needed, matching #4074), and empirically rendered: unset produces byte-identical output to main, set places the field correctly in both PDBs. One alignment item inline (documented value vs commented example, so both charts expose the same API), plus a rebase: main moved to 2.0.6, and #4060 holds 2.1.0, so target `2.2.0` if that merges first. With those, this is an approve. Thanks @Ranexia04.

jmeridth:
@Ranexia04 both suggestions are applied exactly as proposed (documented values matching the #4074 surface, README rows generated) and CI is green. Two mechanical items to close it out: rebase onto current main (#4089 took the chart to `2.0.7`), and target **`2.2.0`** rather than `2.1.0`, which #4061 claimed first. Approval follows the rebase; the substance is done.

jmeridth:
Correction to my comment above: **take a patch, not a minor** — rebase onto `2.0.7` and use `2.0.9` (#4061 takes `2.0.8`). @tico24 pointed out on #4060 that new values have consistently shipped as patches (#3943, #3967, #3982, #3659) and I confirmed that as the convention. My earlier minor guidance was wrong; apologies for the churn.

PR #4044, opened 2026-08-30
fix(argo-cd): hash only config data in checksum/* annotations to avoid restarting pods on chart bump

lemon0333:
Fixes #3958

### What this does

The `checksum/*` pod annotations (`checksum/cm`, `checksum/cmd-params`, `checksum/cmp-cm`, `checksum/repo-server-tls`, `checksum/dex-server-tls`) previously hashed the **entire rendered** ConfigMap/Secret manifest. Each manifest's `metadata.labels` is built from `argo-cd.labels`, which includes `helm.sh/chart: argo-cd-<version>`. That value changes on **every** chart version bump, so the checksums changed on every upgrade and **all** Argo CD pods were rolled even when no configuration actually changed.

This PR adds a small shared helper `argo-cd.config.checksum` that hashes only the `data`/`stringData` section of a rendered config resource, and switches all `checksum/*` annotations to use it:

```yaml
{{- define "argo-cd.config.checksum" -}}
{{- $rendered := include (print .context.Template.BasePath .path) .context | fromYaml -}}
{{- $data := merge (dict) (dig "data" dict $rendered) (dig "stringData" dict $rendered) -}}
{{- $data | toYaml | sha256sum -}}
{{- end -}}
```

Result: a pure chart version bump no longer changes any checksum (pods are not restarted), while a real change to the config `data`/`stringData` still updates the checksum and rolls the affected pods.

> **Note:** adopting this release rolls all Argo CD pods once (every `checksum/*` value changes as it switches to the data-only hash); subsequent bumps do not.

### Verification

Rendered `helm template` at `10.4.2` vs `10.4.3` (version bump only) — all checksums identical:

| annotation | 10.4.2 | 10.4.3 |
|---|---|---|
| checksum/cm | `154d089e…` | `154d089e…` |
| checksum/cmd-params | `82a74192…` | `82a74192…` |
| checksum/cmp-cm | `de56ae00…` | `de56ae00…` |
| checksum/repo-server-tls | `868858687…` | `868858687…` |
| checksum/dex-server-tls | `868858687…` | `868858687…` |

Changing a real `.data` value still moves the checksum, e.g. adding a key to `configs.cm` → `checksum/cm` `154d089e…` → `57ec2838…`, and `configs.params` → `checksum/cmd-params` `82a74192…` → `33cba314…`.

`helm lint charts/argo-cd` passes.

### Context

[...]

PR #4079, opened 2026-09-11
fix(argo-rollouts): hash only config data in checksum/cm annotation

DrFaust92:
Ports the argo-cd fix from #4044 to argo-rollouts.

`checksum/cm` hashed the entire rendered ConfigMap, whose `metadata.labels` include `helm.sh/chart`. That value changes on every chart version bump, so the controller was rolled on every upgrade even when no configuration had changed. The `argo-rollouts.config.checksum` helper hashes only the `data`/`stringData` sections instead.

Rendered `2.43.1` against a bumped version with the same values: the checksum is unchanged after this PR and differs before it. Changing `controller.trafficRouterPlugins` still changes it. Upgrading to this version rolls the controller once as the annotation value changes.

`argo-workflows`, `argo-events` and `argocd-image-updater` have the same pattern — happy to follow up with separate PRs.

Checklist:

### DIFF
--- a/charts/argo-rollouts/Chart.yaml
+++ b/charts/argo-rollouts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: v1.10.0
 description: A Helm chart for Argo Rollouts
 name: argo-rollouts
-version: 2.43.1
+version: 2.43.2
 home: https://github.com/argoproj/argo-helm
 icon: https://argoproj.github.io/argo-rollouts/assets/logo.png
 keywords:
@@ -18,5 +18,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: security
-      description: Harden the dashboard container security context by default
+    - kind: fixed
+      description: Hash only the ConfigMap data in the checksum/cm annotation so a chart version bump does not restart the controller
--- a/charts/argo-rollouts/templates/_helpers.tpl
+++ b/charts/argo-rollouts/templates/_helpers.tpl
@@ -67,6 +67,24 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{/*
+Return a sha256sum of only the `data` and `stringData` sections of a rendered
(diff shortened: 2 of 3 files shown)


charts/argo-rollouts/Chart.yaml
```diff
@@ -18,5 +18,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: security
-      description: Harden the dashboard container security context by default
+    - kind: fixed
+      description: Hash only the ConfigMap data in the checksum/cm annotation so a chart version bump does not restart the controller
```
jmeridth: :robot: (medium) This ports #4044 faithfully in code but diverges from it in the changelog. That PR shipped as `kind: changed` with an explicit one-time-roll disclosure, because adopting the release recomputes the annotation once and rolls the controller a single time on that specific upgrade; without the note, the first user report after release tends to be "the fix didn't work". Same fix class, same standard:
```suggestion
    - kind: changed
      description: Hash only the ConfigMap data in the checksum/cm annotation so a chart version bump no longer restarts the controller. Adopting this release recomputes the annotation once, so the controller rolls a single time on this specific upgrade.
```

jmeridth (changes requested):
:robot: (summary) Verified port of the #4044 data-only checksum pattern, and verified empirically: with this PR applied, a chart version bump leaves `checksum/cm` unchanged while a real config change (e.g. adding a `trafficRouterPlugins` entry) moves it, and `createConfigmap=false` renders safely. Helper matches the argo-cd original, the controller deployment is the chart's only checksum site, and 2.43.2 is the right patch. One changelog item to align with the #4044 precedent (inline), then this is an approve. Thanks @DrFaust92 for spreading the fix.

charts/argo-rollouts/Chart.yaml
DrFaust92: Applied — kind: changed with the one-time roll spelled out.

jmeridth (approved):
:robot: (summary) Suggestion applied verbatim: changelog now carries kind: changed with the one-time-roll disclosure, matching the #4044 precedent. Code was already verified empirically in the prior review (hash stable across a version bump, moves on config change, empty-CM safe). CI green. LGTM, thanks @DrFaust92.

tico24 (approved):

PR #4088, opened 2026-09-17
fix(argo-cd): select server probe scheme from TLS mode

truongnht:
## What / Why

Select the Argo CD server liveness, readiness, and startup probe scheme from the existing `configs.params."server.insecure"` setting.

The chart currently always renders HTTP probes. Although secure Argo CD servers accept those through their redirect listener, the probe protocol should match the configured server protocol. This also prevents TLS-only server configurations from receiving plaintext probes.

The behavior is automatic:

- `server.insecure: true` → `HTTP`
- `server.insecure: false` or unset → `HTTPS`

## Proof it works

- `./scripts/lint.sh`
- `helm lint charts/argo-cd`
- Rendered default values: all enabled server probes use `HTTPS`
- Rendered `server.insecure=true`: all enabled server probes use `HTTP`
- Rendered `server.insecure=false`: all enabled server probes use `HTTPS`

## Risk + AI role

Low: the probe endpoints and timings are unchanged. Kubernetes HTTPS probes do not verify the certificate. AI-assisted implementation and validation.

## Review focus

Confirm deriving the probe scheme from the same `server.insecure` value already used for server Service and route protocol selection is preferred over adding separate probe values.

Checklist:


tico24 (commented):
Won't this break anyone who turns insecure on outside `configs.params`? `--set 'server.extraArgs[0]=--insecure'` still renders `scheme: HTTPS`, but an insecure server only listens for plain HTTP and gRPC ([server.go](https://github.com/argoproj/argo-cd/blob/v3.5.3/server/server.go#L632-L635)), so I think the liveness probe fails and argocd-server ends up restarting over and over. Same for `configs.params.create: false` with the setting in their own configmap.

What is it fixing on v3.5.3? With TLS on, the server still accepts plain HTTP on 8080 and 307s it to HTTPS ([server.go](https://github.com/argoproj/argo-cd/blob/v3.5.3/server/server.go#L609-L611)), which the kubelet counts as a pass. Upstream's own argocd-server-deployment.yaml probes over plain HTTP too. I couldn't find a TLS-only mode, am I missing one?

It also changes the default render (`scheme: HTTPS` on liveness and readiness), so it wants more than a patch. ([versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning))

truongnht:
I agreed, the PR was a bit too optimistic approach. Closed it down now, thank @tico24 for the review.

PR #4090, opened 2026-09-18
fix(github): use a regex for the redis allowedVersions rule

JanWelker:
The `ecr-public.aws.com/docker/library/redis` rule has not produced an update since it was introduced in #4058. `8.6.5-alpine` (2026-07-24) and `8.6.6-alpine` (2026-09-18) were both skipped, while `redis_exporter` bumps from the neighbouring rule flow through normally.

### Cause

`docker` versioning cannot parse `<7.4.0 || >=8.0.0`, so Renovate falls back to npm semver syntax. The tags carry an `-alpine` suffix, which node-semver parses as a **prerelease** and excludes from any range that does not explicitly name one:

```js
semver.validRange('<7.4.0 || >=8.0.0')      // => '<7.4.0||>=8.0.0'  (accepted, so nothing is logged)

semver.satisfies('8.6.4-alpine', range)     // => false   <-- the currently pinned tag
semver.satisfies('8.6.5-alpine', range)     // => false
semver.satisfies('8.6.6-alpine', range)     // => false
semver.satisfies('8.6.6',        range)     // => true    <-- only a suffix-less tag passes
```

The pinned version does not satisfy its own rule. Because the npm-fallback path accepts the range as valid, no `Invalid allowedVersions` error is raised — the filter simply matches nothing, silently.

### Fix

A negated regex is evaluated against the raw tag and bypasses semver entirely, so the licensing intent is preserved exactly:

```js
const re = /^7\.[4-9]\./;
'7.2.8-alpine'  // allowed
'7.4.0-alpine'  // excluded
'7.9.9-alpine'  // excluded
'8.0.0-alpine'  // allowed
'8.6.6-alpine'  // allowed
```

#4058's "Review focus" asked whether the `||` form was honored. This is the answer: it is parsed, but not in the way intended.

### Note for reviewers

Once this merges, Renovate will likely open several redis PRs at once, since the rule has been inert for some weeks. That is the intended effect, but worth expecting rather than discovering.

A companion PR bumps the pin to `8.6.6-alpine` directly, so the current CVEs are not left waiting on the next bot run.

Checklist:


JanWelker:
Consolidated into #4099, which carries this commit unchanged.

Moving it there because it is not really a separate concern: bumping the redis pin without this rule ships a version that silently re-freezes. The `allowedVersions` range excluded every `-alpine` tag as a semver prerelease, which is why the pin sat on a June build for three months with no Renovate PR and no error logged — so the bump and the matcher fix only work together.

It is a distinct commit in that PR (`a9b81d4`), and `renovate.json` is repo-root rather than chart scope, so if a maintainer would rather review it on its own it can be dropped from #4099 and I will reopen this.

Closing in favour of #4099. Nothing is dropped.

PR #4091, opened 2026-09-18
fix(argo-cd): bump Redis image to 8.6.6-alpine for patched OpenSSL

JanWelker:
Bumps the Redis image from `8.6.4-alpine` to `8.6.6-alpine`, clearing two critical CVEs:

| Package | 8.6.4-alpine | 8.6.6-alpine | CVE |
|---|---|---|---|
| libcrypto3 / libssl3 | 3.5.7-r0 | 3.5.8-r0 | CVE-2026-63073, CVE-2026-75803 |

`8.6.6-alpine` was rebuilt 2026-09-18 on Alpine 3.23; the package versions are from the Alpine v3.23 APKINDEX. Same 8.6.x line, patch-level change, and still on an open-source Redis version as required by the note in `values.yaml`.

### Both occurrences

This changes **both** `tag: 8.6.4-alpine` lines — `redis.image.tag` (line 1625) and `redis-ha.image.tag` (line 1926). #4057 moved only the redis-ha one because the standalone was already at 8.6.4; leaving them divergent now would be noise. Renovate itself bumped both together in #3108.

### Why this did not arrive as a bot PR

The `allowedVersions` rule for this package silently filters out every `-alpine` tag: `docker` versioning cannot parse the range, Renovate falls back to npm semver, and the `-alpine` suffix is treated as a prerelease and excluded. `8.6.4-alpine` does not satisfy its own rule. #4090 fixes the rule; this PR fixes the pin so the current CVEs are not left waiting on the next bot run.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning)
* [x] I have updated the documentation according to [documentation](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#documentation)
* [x] I have updated the chart changelog with all the changes that come with this pull request according to [changelog](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#changelog).
* [x] I wrote this PR with the help of a LLM but I fully vetted the work before raising this PR for review
* [x] Any new values are backwards compatible and/or have sensible default.
* [x] I have signed off all my commits as required by [DCO](https://github.com/argoproj/argoproj/blob/master/community/CONTRIBUTING.md).
* [x] I have created a separate pull request for each chart according to [pull requests](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#pull-requests)
[...]

PR #4099, opened 2026-09-18
fix(argo-cd): pin the vendored redis-ha images to rebuilt Alpine bases

### LINKED ISSUE #4098: argo-cd ships stale redis and HAProxy pins; exact-patch image tags stop receiving base rebuilds
> **Edited 2026-09-19.** This body has been rewritten to fold in two follow-up comments, which I have since removed so the issue reads as one argument rather than three overlapping ones. If you received those comments by email, this supersedes them — in particular my original request to wait for a fixed `redis-ha` release, which is not needed and which the text below replaces. Nothing has been withdrawn on the substance; the diagnosis changed and the asks are now stated once, here.

### Summary

`argo-cd` vendors `redis-ha` from `DandyDeveloper/charts` into its published release tarball, and the two container images that subchart contributes — HAProxy and redis — are both currently pinned to stale, CVE-carrying builds. They are stale for two *different* reasons, and neither is really about how fast anyone is reviewing:

- **HAProxy:** `charts/argo-cd/values.yaml` overrides `repository` but not `tag`, so the tag falls through to the vendored subchart default of `3.3.10-alpine`.
- **redis:** the tag *is* overridden here — but the override has frozen in place and now sits *behind* the vendored subchart's own pin.

Underneath both is one mechanism: **an exact patch tag stops receiving base-image rebuilds the moment the next patch ships.** That makes every hand-maintained `X.Y.Z-alpine` pin decay on a timescale of weeks, no matter how attentive the maintainer is. I think that, rather than upstream cadence, is the thing worth fixing.

### Instance 1 — HAProxy: `tag` is not overridden

`argo-cd-10.9.2.tgz` vendors redis-ha 4.38.0:

```
$ tar xzf argo-cd-10.9.2.tgz -O argo-cd/charts/redis-ha/Chart.yaml | grep ^version
version: 4.38.0
```

which pins `haproxy.image.tag: 3.3.10-alpine`, carrying:

| Package | In 3.3.10-alpine | Fixed in | CVE |
|---|---|---|---|
| libcrypto3 / libssl3 | 3.5.7-r0 | 3.5.8-r0 | CVE-2026-63073, CVE-2026-75803 |
| socat | 1.8.1.1-r0 | 1.8.1.2-r0 | CVE-2026-56123 |

The HAProxy block in this repo stops one line short of fixing it (`charts/argo-cd/values.yaml:1967-1969` on `main`):

```yaml
    image:
      # -- HAProxy Image Repository
      repository: ecr-public.aws.com/docker/library/haproxy
```

`repository` is overridden; `tag` is not. **This does not need upstream.** Adding one line here, in the same shape and for the same reason as the redis override 40 lines above, ships the fix in the next `argo-cd` release:

```yaml
    image:
      # -- HAProxy Image Repository
      repository: ecr-public.aws.com/docker/library/haproxy
      # -- HAProxy Image Tag
      ## Runs ahead of the upstream HA manifests' pin, as with the redis image above:
      ## 3.3.10-alpine carries CVE-2026-63073, CVE-2026-75803 (libcrypto3/libssl3) and
      ## CVE-2026-56123 (socat). 3.3.14-alpine is the same 3.3 line rebuilt on Alpine 3.24.
      tag: 3.3.14-alpine
```

Verified against `argo-cd-10.9.2.tgz` with `helm template`; the HAProxy tag is the entire functional delta between the vendored 4.38.0 and the fixed chart, so this does not mask other drift.

### Instance 2 — redis: the override has inverted its own purpose

`argo-cd` overrides the redis image in two places in its own `values.yaml` (line numbers from the 10.9.2 tarball): line 1625 in the non-HA `redis:` block, and line 1926 in `redis-ha:`, where the comment states the intent:

```yaml
    ## Do not use 7.4.0 <= v < 8.0.0, otherwise you are no longer using an open source version of Redis
    ## Runs ahead of the upstream HA manifests' pin: the redis 8.2.x line is only built on Alpine 3.22,
    ## whose OpenSSL carries known vulnerabilities (GHSA-5p3w-hgjv-f6q3 report); 8.6.x is the patched base.
    tag: 8.6.4-alpine
```

That override was added to run *ahead* of the vendored pin. It now runs *behind* it — vendored redis-ha 4.38.0 pins `8.8.0-alpine`, while this override holds users at `8.6.4-alpine`. And `8.6.4-alpine` carries the same OpenSSL pair as the HAProxy image above:

| Package | In 8.6.4-alpine | Fixed in | CVE |
|---|---|---|---|
| libcrypto3 / libssl3 | 3.5.7-r0 | 3.5.8-r0 | CVE-2026-63073, CVE-2026-75803 |

So the image the chart deliberately took control of is now stale in the same way as the one it left alone. That is what convinced me the cadence framing was the wrong diagnosis.

### The mechanism

Docker Official Images rebuild only the **newest** patch tag in a line. Pin an exact patch and base-image security rebuilds stop arriving as soon as the next patch ships. Build dates from Docker Hub, gathered 2026-09-19:

| Image | Tag | Last built | Status |
|---|---|---|---|
| redis | `8.6.4-alpine` | 2026-06-23 | **frozen** — this chart's override |
| redis | `8.6.6-alpine` | 2026-09-18 | current head of 8.6 |
| redis | `8.8.0-alpine` | 2026-06-23 | **frozen** — vendored redis-ha's pin |
| redis | `8.8.2-alpine` | 2026-09-18 | current head of 8.8 |
| redis | `8.10.1-alpine` | 2026-09-18 | current head of 8.10 |
| haproxy | `3.3.10-alpine` | 2026-06-19 | **frozen** — vendored redis-ha's pin |
| haproxy | `3.3.14-alpine` | 2026-09-18 | current head of 3.3 |
| haproxy | `3.3-alpine` | 2026-09-18 | minor line, rebuilt continuously |

Alpine shipped `openssl 3.5.8-r0` on 2026-08-25. All three frozen pins predate it by two months and can never receive it.

The telling part is that those three froze within four days of each other — 2026-06-19 and 2026-06-23 — across two different images, one overridden here and two inherited. Nobody made three bad decisions in June. They stopped being rebuilt because a newer patch shipped in each line, which is what an exact-patch pin does by construction. Meanwhile the minor-line tags for the same lines were rebuilt on 2026-09-18.

### On severity, stated plainly

The scanner ratings on the OpenSSL pair are Critical, but OpenSSL — who are the CNA for both — rate them **Low**. The Critical figure comes from an auto-generated CISA-ADP vector. CVE-2026-63073 is a format string in the CMP client, reachable only when a caller configures an expected sender; CVE-2026-75803 skips AEAD tag verification in a one-shot `EVP_Cipher` call that OpenSSL's own TLS stack never takes. Neither is reachable from redis or HAProxy as deployed here.

I am raising this as hygiene on frozen bases, not as an exposure, and I would rather say so than have it triaged on a scanner number that overstates it.

### What I am asking for

1. **Override the HAProxy tag** (`3.3.14-alpine`), one line, no upstream dependency. This supersedes my original ask to wait for a fixed `redis-ha` release — that framing made the fix hostage to [DandyDeveloper/charts#424](https://github.com/DandyDeveloper/charts/pull/424), which is unreviewed against a repo whose last `master` commit was 2026-06-15. It is not needed.
2. **Bump the two redis pins** off the frozen `8.6.4-alpine`.
3. **Track these pins automatically**, so neither the overridden nor the inherited tag has to be caught by hand after a scanner finds it. argo-helm already runs Renovate, and the `helm-values` manager reads these blocks natively once `repository` and `tag` sit together — already true for redis, and true for HAProxy with the change in (1).

On (2) and (3) there are two directions and I genuinely don't know which you'd prefer:

- **Keep exact tags and let Renovate move them.** Preserves reproducibility, which is presumably why they are exact today.
- **Pin the minor line** (`8.6-alpine`, `3.3-alpine`), so base rebuilds arrive with no PR at all — at the cost of reproducibility and of digest pinning.

Happy to raise either as a draft PR, including the `Chart.yaml` bump and `artifacthub.io/changes` entry. Just say which direction.

### Workarounds for anyone hitting this now

```yaml
redis-ha:
  image:
    tag: 8.6.6-alpine
  haproxy:
    image:
      tag: 3.3.14-alpine
```

`8.6.6-alpine` stays inside the 8.6 line, keeping the original comment's intent — clear of `7.4.0 <= v < 8.0.0` for the licence change and clear of 8.2.x with its Alpine 3.22 base. It was rebuilt 2026-09-18 on Alpine 3.23. Both tags verified present on the `ecr-public` mirror the chart pulls from, for amd64 and arm64.

### Supporting context: why the inherited pin drifted unnoticed

This part is no longer the core of the ask, but it explains why (3) matters. Evidence from `DandyDeveloper/charts`, gathered 2026-09-18:

- **No dependency automation.** No `renovate.json*`, `.renovaterc*`, `.github/renovate.*` or `.github/dependabot.*`, and no bot-authored PR in the repo's history. Every image tag in `values.yaml` — haproxy, redis, busybox, shellcheck, redis_exporter — is hand-maintained.
- **Bursty cadence.** Last `master` commit 2026-06-15. Releases 4.36.1 through 4.39.0 were all cut inside a single 24-hour window on 2026-06-14/15; before that 2026-02-14, before that 2026-01-27.
- **Review latency.** 14 open PRs, oldest from 2025-12-09; 30 open issues.

To be fair to that repository: it is not archived, and the maintainer merged six PRs in the June burst including a prior CVE bump for this same image. The problem is that the cadence is long relative to CVE timelines with no automation filling the gaps, while argo-helm's release process bakes whatever state it is in into a tarball. `argo-cd` currently vendors 4.38.0 against 4.39.0 released, so that tail is not hypothetical.

This overlaps with #3960 (replace the redis image with valkey) and #4087 (no way to use hardened images with an HA Redis setup) — all symptoms of the HA Redis path being sourced from a chart this repo does not control and cannot patch between upstream bursts.

JanWelker:
Closes #4098. Supersedes #4091 and #4090, both consolidated here — see *Why one PR* below.

### What this PR does / why we need it

`argo-cd` vendors `redis-ha` into its published release tarball, so that subchart's image pins travel with this chart and consumers cannot redirect them:

```
$ tar xzf argo-cd-10.9.2.tgz -O argo-cd/charts/redis-ha/Chart.yaml | grep ^version
version: 4.38.0
```

Both images that subchart contributes are currently on bases that can no longer receive security updates:

| Image | Pinned | Last rebuilt | Affected packages |
|---|---|---|---|
| HAProxy | `3.3.10-alpine` (vendored default) | 2026-06-19 | libcrypto3/libssl3 3.5.7-r0, socat 1.8.1.1-r0 |
| redis | `8.6.4-alpine` (this chart's override) | 2026-06-23 | libcrypto3/libssl3 3.5.7-r0 |

CVEs: CVE-2026-63073 and CVE-2026-75803 (libcrypto3/libssl3, fixed in 3.5.8-r0) on both, plus CVE-2026-56123 (socat, fixed in 1.8.1.2-r0) on HAProxy.

This PR moves HAProxy to `3.3.14-alpine`, redis to `8.6.6-alpine`, and fixes the Renovate rule that let the redis pin freeze in the first place.

**On severity, so this is triaged on the right basis:** scanners report the OpenSSL pair as Critical, but OpenSSL — the CNA for both — rates them **Low**, and the Critical figure comes from an auto-generated CISA-ADP vector. CVE-2026-63073 is a format string in the CMP client, reachable only when a caller configures an expected sender; CVE-2026-75803 skips AEAD tag verification in a one-shot `EVP_Cipher` call that OpenSSL's own TLS stack never takes. Neither is reachable from redis or HAProxy as deployed here. I have not independently checked the socat rating and am not claiming one.

So this is hygiene on frozen bases, not a live exposure. I would rather say that than have it reviewed against a number that overstates it.

### Why one PR

These were three PRs. They are one concern and they conflicted by construction — #4099 and #4091 both bumped `Chart.yaml` to 10.9.3 and both replaced the single `artifacthub.io/changes` entry, so landing either forced a rebase of the other, for one security fix on one chart with one cause.

More importantly, the redis half is not fixed by the version bump alone. #4090 is what makes it stay fixed, so splitting them ships a bump that silently re-freezes.

One note for reviewers: `renovate.json` is repo-root, not chart scope — it needs no version bump and does not trigger chart-testing. It is a separate commit, so if you would rather take it separately, dropping `a9b81d4` is enough.

### Why these pins went stale, and what stops it recurring

Docker Official Images rebuild only the **newest** patch tag in a line. An exact-patch pin therefore stops receiving base-image rebuilds the moment the next patch ships — which is why three pins across these two images all froze within four days of each other:

| Image | Tag | Last built | |
|---|---|---|---|
| haproxy | `3.3.10-alpine` | 2026-06-19 | frozen — vendored default, replaced here |
| redis | `8.6.4-alpine` | 2026-06-23 | frozen — this chart's override, replaced here |
| redis | `8.8.0-alpine` | 2026-06-23 | frozen — vendored default, unused once overridden |
| haproxy | `3.3.14-alpine` | 2026-09-18 | current head of 3.3 |
| redis | `8.6.6-alpine` | 2026-09-18 | current head of 8.6 |

Nobody made three bad decisions in June; that is what an exact-patch pin does. The two images were then unreachable by the bot for two *different* reasons, and both needed fixing:

- **HAProxy:** the tag was not expressed in `values.yaml` at all — `repository` was overridden but `tag` fell through to the subchart — so there was nothing for Renovate to see. There is no HAProxy rule in `renovate.json`, so putting `tag` beside `repository` is sufficient; the `helm-values` manager reads the block natively.
- **redis:** the tag *was* expressed and Renovate still never bumped it, because `allowedVersions: "<7.4.0 || >=8.0.0"` silently matched nothing. The tags carry an `-alpine` suffix, which npm semver treats as a prerelease and excludes from any range that does not name one, so every candidate was dropped with no error logged. A matcher that excludes everything is indistinguishable from "no updates available", which is why this went unnoticed for three months. Replaced with `!/^7\.[4-9]\./`, which expresses the same licence exclusion without the prerelease trap.

If you would rather pin the minor lines (`3.3-alpine`, `8.6-alpine`, both rebuilt 2026-09-18) and let base rebuilds arrive with no PR at all, I am happy to switch — that trades reproducibility and digest pinning for not needing the bot. #4098 lays out both directions.

### Also included

`docs(argo-cd)`: the HAProxy comment claimed it "runs ahead of the upstream HA manifests' pin, as with the redis image above", but the redis override sat *behind* the vendored `8.8.0-alpine`, so the cross-reference pointed at a counter-example. Replaced with the rule to preserve on the next bump.

`fix(argo-cd)`: `README.md.gotmpl` appended five literal rows after the `range` loop that already emits every `redis-ha` value — `exporter.image`, `exporter.tag`, `haproxy.image.repository`, `haproxy.image.tag`, `image.repository` — each claiming `nil` (follows subchart default). All five are set in `values.yaml`, so the table carried every one of those keys **twice with contradictory defaults**. Four of those duplicates predate this branch; expressing `haproxy.image.tag` creates the fifth, which is what surfaced it. Removed from the template and the generated README identically — helm-docs copies these literal rows through verbatim, so deleting them from the template deletes exactly those lines and nothing else.

### Scope

`3.3.14-alpine` is the newest HAProxy 3.3 patch, rebuilt 2026-09-18 on Alpine 3.24 — verified by digest: `3.3.14-alpine` and `3.3.14-alpine3.24` are both `sha256:30a5bbb4a54d92b42fff3d99ffe679c56b3a9d4a43cb3c4649963d84a0a08605`. `8.6.6-alpine` is the newest 8.6 patch, rebuilt 2026-09-18 on Alpine 3.23, and stays inside the line the existing comment deliberately chose — clear of `7.4.0 <= v < 8.0.0` for the licence change and of 8.2.x with its Alpine 3.22 base.

Patch-level changes to two images; no template or behavioural change. The one new value has a sensible default and is backwards compatible — anyone already setting `redis-ha.haproxy.image.tag` keeps their override.

### Testing

```
$ helm dependency build && helm template argocd . --set redis-ha.enabled=true --set redis.enabled=false \
    | grep -E '^\s+image:' | sort -u
        image: ecr-public.aws.com/docker/library/haproxy:3.3.14-alpine
        image: ecr-public.aws.com/docker/library/redis:8.6.6-alpine
    image: ecr-public.aws.com/docker/library/redis:8.6.6-alpine
        image: quay.io/argoproj/argocd:v3.5.3
        image: ghcr.io/dexidp/dex:v2.45.1

$ helm lint . --set redis-ha.enabled=true --set redis.enabled=false
1 chart(s) linted, 0 chart(s) failed

$ bash scripts/lint-changelog.sh
All charts have a single changelog entry ✔
```

I could not run `scripts/lint.sh` or `scripts/helm-docs.sh`: both require Docker, which was unavailable on the machine I prepared this on. The README change is a pure deletion of verbatim template rows, so it does not depend on helm-docs output; but please treat the generated README as unverified by the tool and let CI confirm it.

Checklist:

* [x] I have bumped the chart version according to [versioning](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#versioning) (10.9.2 -> 10.9.3, patch: backwards compatible security fix)


charts/argo-cd/values.yaml
```diff
@@ -1967,6 +1969,13 @@ redis-ha:
     image:
       # -- HAProxy Image Repository
       repository: ecr-public.aws.com/docker/library/haproxy
+      # -- HAProxy Image Tag
+      ## Pinned here because the vendored redis-ha default, 3.3.10-alpine, was last rebuilt
+      ## 2026-06-19 and carries CVE-2026-63073, CVE-2026-75803 (libcrypto3/libssl3) and
+      ## CVE-2026-56123 (socat). 3.3.14-alpine is the newest 3.3 patch, rebuilt on Alpine 3.24.
+      ## Keep this on the newest patch in the line: Docker rebuilds only the newest patch tag, so
+      ## an older exact patch stops receiving base-image security updates. Renovate tracks it here.
```
tico24, line 1977: > Renovate tracks it here.

It'll find it, but won't its PRs skip the chart bump? The redis and exporter rules run `renovate-bump-version.sh`, and nothing in `renovate.json` matches `ecr-public.aws.com/docker/library/haproxy`, so Renovate's HAProxy PRs would fail `ct lint` on the version check. Adding it to the `matchPackagePatterns` list next to `redis_exporter` should cover it.

charts/argo-cd/Chart.yaml
```diff
@@ -26,5 +26,5 @@ annotations:
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
   artifacthub.io/changes: |
-    - kind: fixed
-      description: Honor per-component gRPC TXT service config settings
+    - kind: security
+      description: Pin the vendored redis-ha images to rebuilt Alpine bases - redis 8.6.6-alpine and HAProxy 3.3.14-alpine - clearing CVE-2026-63073, CVE-2026-75803 and CVE-2026-56123
```
tico24, line 30: The redis bump hits the default single-node Redis as well, not just redis-ha, so everyone on the chart gets it. Could the changelog say so instead of "the vendored redis-ha images"?

